PR #22, opened 2017-03-24
Checksumed addresses

debris:


src/util/native.js
```diff
@@ -1,8 +1,16 @@
 import { EthkeyBridge } from 'NativeModules'
 
 export const brainWalletAddress = (seed) => EthkeyBridge.brainWalletAddress(seed)
+  .then(address => keccak(address).then(hash => ({address, hash: hash.slice(-40)})))
+  .then(acc => {
+    let result = ''
```
tomusdrw: Shouldn't it contain `0x` prefix?

debris: it's added elsewhere, only for display

PR #23, opened 2017-03-24
Account details blockies

### LINKED ISSUE #16: display identicon in account details

debris:
based on #22 

closes #16 

screen shot 2017-03-24 at 17 35 32
screen shot 2017-03-24 at 17 39 27


src/components/AccountDetails.js
```diff
@@ -13,9 +26,19 @@ export default class AccountDetails extends Component {
     onDeleteAccountPressed: PropTypes.func.isRequired,
   }
 
+  constructor(props) {
+    super(props)
+    this.state = {}
```
tomusdrw: This should be part of the class:
```
class AccountDetails ... {
 state = {};
}
```

and `displayIcon` should be called in `componentDidMount`
```
componentDidMount () {
  displayIcon(this, `0x` + this.props.account.address)
}
```
TBH `displayIcon` could be just part of the class and if we should also call this method on `componentWillReceiveProps`, but currently I guess it's not possible for a parent to modify `this.props.account` so it should work fine.

src/components/AccountDetails.js
```diff
@@ -1,8 +1,21 @@
 'use strict'
 
 import React, { Component, PropTypes } from 'react'
-import { StyleSheet, View, ScrollView, Text, Button } from 'react-native'
+import { StyleSheet, View, ScrollView, Text, Button, Image } from 'react-native'
 import AppStyles from '../styles'
+import { blockiesIcon } from '../util/native'
+
+async function displayIcon(self, seed) {
```
tomusdrw: Seems like a copy-paste from `AccountListRow`?

src/components/AccountListRow.js
```diff
@@ -26,11 +26,10 @@ export default class AccountListRow extends Component {
   constructor(props) {
     super(props)
     this.state = {}
+    displayIcon(this, this.props.lowerText)
```
tomusdrw: Same stuff, should be part of `componentDidMount` or `componentWillMount` and should also have `componentWillReceiveProps` (react may re-use the component if say order of the list is changed)

```
componentWillReceiveProps (newProps) {
   if (newProps.lowerText !== this.props.lowerText) {
     displayIcon(this, newProps.lowerText);
   }
}

File: src/components/AccountDetails.js (hunk 2 above)
Comment (debris) [inline src/components/AccountDetails.js]: yes... I will just create a new component called `Blockies`

**Comment by debris** [comment]
also added launch screen for iOS

File: .travis.yml [hunk 4]
```diff
@@ -10,5 +10,6 @@ cache:
 install: true
 
 script:
+  - npm i
```
tomusdrw: Should go to `install`

src/components/AccountIcon.js
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  state = {}
+
+  componentDidMount () {
+    displayIcon(this, this.props.seed)
+  }
+
+  componentWillReceiveProps (newProps) {
+    displayIcon(this, newProps.seed)
```
tomusdrw: This can be fired even if `seed` was not changed, so it's better to early exit if there are no changes.

src/components/TxDetails.js
```diff
@@ -20,6 +22,9 @@ export default class Send extends Component {
   render () {
     return (
       <ScrollView style={AppStyles.view}>
+        <AccountIcon style={AppStyles.icon} seed={'0x' + this.props.txSenderAddress} />
+        <Text style={AppStyles.hintText}>sender address</Text>
+        <Text style={AppStyles.valueText}>{this.props.txSenderAddress}</Text>
```
tomusdrw: Should probably be selectable (to check on block explorer for instance)

src/components/AccountIcon.js
debris: not really, this is the only property of this component :p

src/components/TxDetails.js
debris: filled new issue https://github.com/paritytech/native-signer/issues/29

.travis.yml
debris: I added it, cause `install: true` did not work... weird

tomusdrw: Yeah, in my PR I did this to make it work:
```
install:
 - npm i
```

src/components/AccountIcon.js
tomusdrw: I think it can be triggered spouriously anyway, but it's not critical :)

PR #28, opened 2017-03-25
Fixing PIN screens

### LINKED ISSUE #5: PIN screens are missing a button
There is no button on those screens and when you close the keyboard there is no way to move to next step.
screenshot_20170322-115915

Affected screens: Set PIN, Confirm PIN, * PIN (when confirming a transaction)

### LINKED ISSUE #13: pin is shown when confirming a transaction

### LINKED ISSUE #15: pin should have numeric keyboard

tomusdrw:
- Input for brain wallet is auto-adjustable
screen
- Selectable values on Acount Details
- PINs are now masked and numeric (Closes #15 and closes #13 )
- Buttons on PIN screens (Closes #5) although I think it would be better to put in in the topbar, but it's a bit more complicated.
screen

### DIFF
--- a/src/components/AccountPin.js
+++ b/src/components/AccountPin.js
@@ -1,7 +1,7 @@
 'use strict'
 
 import React, { Component, PropTypes } from 'react'
-import { TextInput, View } from 'react-native'
+import { TextInput, View, Button, StyleSheet } from 'react-native'
 import AppStyles from '../styles'
 
 export default class AccountPin extends Component {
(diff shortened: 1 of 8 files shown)


src/components/AccountPin.js
```diff
@@ -16,25 +16,57 @@ export default class AccountPin extends Component {
...
             value={this.state.text}
-            onEndEditing={() => { this.props.onNextPressed(this.state.text, this.props.account, this.props.extra) }}
-        />
+          />
+          <View style={[AppStyles.buttonContainer, styles.button]}>
+            <Button
+              onPress={this.onNext}
+              color='green'
+              title='Next'
+              accessibilityLabel='Confrim PIN'
```
debris: pressing this button triggers `this.onNext` twice. First time from `onPress` and second time from `onEndEditing`

debris: To reproduce:
- create new account and when confirming pin, enter incorrect one. It will show `Invalid PIN` popup twice

debris: actually, I will fix it by myself and merge the pr after

Issue #29, opened 2017-03-25
AccountAddress component


debris:
create selectable `Text` component (eg. to check the block explorer)

tomusdrw:
Just read about this: https://facebook.github.io/react-native/docs/touchablehighlight.html

PR #30, opened 2017-03-26
AccountAddress component and new qr code transaction format

### LINKED ISSUE #20: new qr code transaction format
```
{
	action: "signTransaction",
	data: {
		account: "",
		rlp: "",
	}
}
```

```
{
	action: "signTransactionHash",
	data: {
		account: "",
		hash: "",
		details: {
			to: "",
			value: 0,
			...
		}
	}

}
```

debris:
fixes #20 and #29


src/containers/AccountPin.js
```diff
@@ -6,17 +6,17 @@ import { Actions } from 'react-native-router-flux'
...
-  account: state.accounts.selected,
-  extra: state.transactions.pendingTransaction.rlp
+  account: state.accounts.selected
 })
 
-async function signTransaction (dispatch, account, rlp) {
+async function signTransaction (dispatch, account) {
   try {
-    let hash = await keccak(rlp)
+    let hash = store.getState().transactions.pendingTransaction.rlpHash
```
tomusdrw: Accessing `store` in actions is a little bit smelly to me. I think it's better handled in a middleware. We can also use `redux-thunk` to get this abstraction.

src/components/AccountAddress.js
```diff
@@ -0,0 +1,29 @@
...
+export default class AccountAddress extends Component {
+  static propTypes = {
+    address: PropTypes.string.isRequired
+  }
+
+  state = {
+    address: this.props.address
+  }
+
+  componentWillReceiveProps (newProps) {
```
tomusdrw: Why do you copy `props` to `state`? Doesn't seem to process those in any way, so I think it's superfluous.

src/containers/QrScanner.js
```diff
@@ -7,38 +7,65 @@ import QrScanner from '../components/QrScanner'
...
 
+function displayAlert (text) {
+  Alert.alert(text, undefined, [
+    { text: 'OK', onPress: () => { scanning = false } }
+  ])
+}
+
+function findAccountWithAddress (address) {
+  return store.getState().accounts.all.find(account => {
+    return account.address.toLowerCase() === address
```
tomusdrw: `=== address.toLowerCase()`?

src/containers/QrScanner.js
```diff
@@ -7,38 +7,65 @@ import QrScanner from '../components/QrScanner'
 import { selectAccount } from '../actions/accounts'
 import { scannedTx } from '../actions/transactions'
 import transaction from '../util/transaction'
+import keccak from '../util/native'
 import store from '../util/store'
 
 var scanning = false
```
tomusdrw: That's a bit ugly, should probably be part of the store.

src/containers/QrScanner.js
```diff
@@ -7,38 +7,65 @@ import QrScanner from '../components/QrScanner'
...
-    let account = store.getState().accounts.all.find(account => {
-      return account.address.toLowerCase() === from
-    })
-    if (!account) {
-      Alert.alert('Invalid sender address ' + txRequest.from, undefined, [
-        { text: 'OK', onPress: () => { scanning = false } }
-      ])
-      return
+    switch (txRequest.action) {
+      case 'signTransaction': {
```
tomusdrw: Syntax is a bit weird, `{}` block is not necessary :)

I prefer early-exiting if statements instead of `switch`, less probability to forget break/return, simpler flow

```js
if (txRequest.action === 'signTransaction') {
   ...
   return
}

if (txRequest.action === 'signTransactionHash') {
  ...
  return
}

displayAlert('')
```

src/components/AccountAddress.js
debris: you are right

src/containers/QrScanner.js
debris: hah, yes :)

src/containers/AccountPin.js
debris: I'm not familiar with `redux-thunk`. How would it improve this function?

tomusdrw: `redux-thunk` allows you to produce actions that are actually functions that look like this: `action = (dispatch, getState) => {}`. You can then create a chain of interconnected actions handling async calls correctly.

Under the hood it's just a simple middleware that literally just does this:
```js
state => next => action => {
  if (typeof action === 'function') {
    return action(dispatch, state.getState)
  }
  
  return next(action);
}
```
https://github.com/gaearon/redux-thunk/blob/master/src/index.js#L2

So you wouldn't need to import the whole `store` singleton and I think it would be a bit cleaner thanks to that (action could be re-used for other stores and wouldn't have direct dependency on it).

PR #37, opened 2017-03-29
Use wordlist package.

tomusdrw:


package.json
```diff
@@ -14,6 +14,7 @@
     "commit": "commit-wizard"
   },
   "dependencies": {
+    "@parity/wordlist": "^1.0.1",
```
debris: I've never seen it. how is it resolved?

tomusdrw: It's just a scoped package name. So you don't need to worry about conflicts.

PR #54, opened 2017-04-10
fixed #53, checksum mismatch

### LINKED ISSUE #53: checksum address mismatch
native-signer incorrectly caluclates checksum

debris:
checksum's keccak hashes addresses as strings, not as bytes


tomusdrw:
Can you add couple of tests to prove the correctness? (Preferably the same vectors as in parity)

debris:
I would like to, but native modules are not available in test environment :/ I will create an issue to set up this

tomusdrw:
You can mock native modules or separate wallet derivation from checksum encoding.

debris:
I've added 2 tests. They are not exactly the same as parity's (parity has only 1 test for it), but I've verified those results agains parity

PR #55, opened 2017-04-10
fixed #36, proper random source

### LINKED ISSUE #36: Make sure that proper random source is used
Related https://github.com/paritytech/wordlist/issues/2

There is no `Buffer` nor `Uint8Array` available, so we need something that works in pure JS and has good enough entropy.

debris:


src/components/IconChooser.js
```diff
@@ -6,8 +6,12 @@ import AppStyles from '../styles'
 
 import AccountIcon from './AccountIcon'
 
-import { words } from '../util/random'
-import { brainWalletAddress } from '../util/native'
+import { brainWalletAddress, words } from '../util/native'
+
+async function refreshIcons (self) {
```
tomusdrw: Why do we move it outside of the `Component`? You also need to try-catch asynchrounous `setState` invocations.
I think:
```js
refreshIcons = async () => {
  const icons = await ...;
 
  // try is required, since the component might not be mounted any more.
  try {
   this.setState({ icons });
  } catch (e) {
     console.error(e);
  }
}
```
Should work as well.

src/components/IconChooser.js
```diff
@@ -42,7 +41,7 @@ export default class IconChooser extends Component {
           }
         </View>
         <Button
-          onPress={this.refreshIcons}
+          onPress={() => refreshIcons(this)}
```
tomusdrw: This has much worse performance since with every `render` the callback function is different, keeping the function in the component is recommended. (or writing another wrapper if it's not possible).

src/components/IconChooser.js
debris: I didn't know about `async ()` syntax and `async function` was not working :p

Issue #48, opened 2017-04-01
android app comes up with a message "Invalid transaction StntaxError: JSON Parse error: Unable to p...." every 15 seconds or so


debris:

debris:
@tomusdrw can you check that? :/

tomusdrw:
I'm getting this error when scanning an address QR code instead of transaction QR.
I think the solution will be:
1. If address is detected display a message that you should scan transaction instead of address (but only for the first time).
2. Later ignore all invalid QR codes found (or just display a toast/non-blocking notification instead of blocking message).

PR #57, opened 2017-04-10
refactored qr scanner

### LINKED ISSUE #47: contents of data also empty when it's a big data tx - should be <Data unknown!>

debris:
fixed #47 #48 #49 

- signTransaction

screen shot 2017-04-10 at 19 11 38
screen shot 2017-04-10 at 19 11 57

- signTransactionHash

screen shot 2017-04-10 at 19 11 16

screen shot 2017-04-10 at 19 11 20


tomusdrw:
I think that for unsafe transactions it would reasonable to display the hash of the transaction so that you can compare it and see that you're indeed signing the correct message.

src/containers/TxDetails.js
```diff
@@ -29,7 +29,14 @@ const mapStateToProps = (state, ownProps) => ({
   txNonce: state.transactions.pendingTransaction.transaction.nonce,
   txGas: state.transactions.pendingTransaction.transaction.gas,
   txGasPrice: state.transactions.pendingTransaction.transaction.gasPrice,
-  txData: state.transactions.pendingTransaction.transaction.data
+  txData: state.transactions.pendingTransaction.transaction.data,
+  isSafe: state.transactions.pendingTransaction.transaction.isSafe,
+  fetchAccountName: (address) => {
```
tomusdrw: Why do we need this? We already know `txSenderAddress` and `txRecipientAddress` here, why not fetch the names here as well?

src/reducers/scanner.js
```diff
@@ -0,0 +1,55 @@
...
+      }
+
+    case DISABLE_SCANNER_WARNINGS:
+      return {
+        ...state,
+        shouldDisplayWarning: false
+      }
+
+    case RESET_SCANNER:
+      return {
```
tomusdrw: IMHO for future-proofness it would be good to consistently have:
```js
return {
  ...state,
  scannerEnabled: false,
  shouldDisplayWarning: false,
}
```
as well.

src/components/TxDetails.js
```diff
@@ -19,41 +19,56 @@
 import React, { Component, PropTypes } from 'react'
 import { ScrollView, View, Text, Button } from 'react-native'
 import AppStyles from '../styles'
-import AccountIcon from './AccountIcon'
-import AccountAddress from './AccountAddress'
+import AccountPrettyAddress from './AccountPrettyAddress'
+
+const orUnknown = (value = 'Unknown') => value
```
tomusdrw: neat! :)

src/components/TxDetails.js
```diff
@@ -19,41 +19,56 @@
...
+        <Text style={AppStyles.valueText}>{orUnknown(this.props.txGasPrice)}</Text>
         <Text style={AppStyles.hintText}>data</Text>
-        <Text style={AppStyles.valueText}>{this.props.txData}</Text>
+        <Text style={AppStyles.valueText}>{orUnknown(this.props.txData)}</Text>
+        {
+          !this.props.isSafe
+            ? <Text style={AppStyles.hintText}>
+              Signing this transaction is unsafe. Proceed only if this transaction comes from trusted source.
+            </Text>
+            : <View />
```
tomusdrw: You can just return `null`

src/containers/TxDetails.js
debris: you are right :D

src/components/TxDetails.js
debris: didn't know that :)

src/reducers/scanner.js
debris: fixed

PR #59, opened 2017-04-11
seeds are always encrypted using pin

debris:
still in progress, for now it works only on iOS


tomusdrw:
IMHO would be better to encrypt (using some heavy algorithm to discourage bruteforcing pins) only for storage and store unencrypted keys in the memory for simplicity/performance reasons (we can drop unencrypted data from memory after some inactivity).

debris:
@tomusdrw could you review again?

src/actions/accounts.js
```diff
@@ -16,14 +16,35 @@
...
+      let account = getState().accounts.selected
+      if (!account) {
+        return
+      }
+
+      let seed = await encryptData(account.seed, pin)
+      dispatch({
+        type: ADD_ACCOUNT,
+        account: {
+          encryptedSeed: seed,
```
tomusdrw: ADD_ACCOUNT is not overwriting `selected` account so it won't contain `encryptedSeed` property required for instance to call `setOldPin` action.

PR #67, opened 2017-05-17
signData

debris:


src/actions/scanner.js
```diff
@@ -90,18 +90,26 @@ export function scannerDispatch (data) {
...
       } else if (txRequest.action === 'signTransactionHash') {
         var details = txRequest.data.details
         details.isSafe = false
         let hash = txRequest.data.hash
         dispatch(selectAccount(account))
         dispatch(scannedTx(hash, details))
+        Actions.txDetails()
+        dispatch(resetScanner())
+      } else if (txRequest.action === 'signData') {
+        let hash = txRequest.data.hash
```
debris: @tomusdrw can we make it a `keccak256("\x19Ethereum Signed Message:\n" + len(message) + message))` ?

src/actions/scanner.js
```diff
@@ -90,18 +90,26 @@ export function scannerDispatch (data) {
...
+        dispatch(selectAccount(account))
+        dispatch(scannedHash(hash))
+        Actions.accountEnterPin()
+        dispatch(resetScanner())
       } else {
         dispatch(displayScannerWarning('Invalid request'))
```
tomusdrw: Doesn't the scanner need to be reset after the warning is displayed?

src/actions/scanner.js
tomusdrw: But the whole point is that you should be able to see the data before you sign it.
I can prepend the text in the UI (before constructing the QR code), but that could be done here as well (you can even re-use the rust function maybe if it's extracted to `ethkey/ethstore`)

src/containers/TxDetails.js
```diff
@@ -30,17 +30,17 @@ const fetchAccountName = (state, address = '') => {
 const mapStateToProps = (state, ownProps) => ({
```
tomusdrw: We should also display the message to be signed, and it requires a separate screen for that.

debris: DataDetails view displays it now

src/actions/scanner.js
```diff
@@ -90,18 +90,28 @@ export function scannerDispatch (data) {
...
         var details = txRequest.data.details
         details.isSafe = false
         let hash = txRequest.data.hash
         dispatch(selectAccount(account))
         dispatch(scannedTx(hash, details))
+        Actions.txDetails()
+        dispatch(resetScanner())
+      } else if (txRequest.action === 'signData') {
+        let data = txRequest.data.data
+        let hash = await keccak(txRequest.data.data)
```
debris: Here we should 'recognize' what kind of data is being signed. If it's ascii or not ;)

src/components/DataDetails.js
```diff
@@ -0,0 +1,43 @@
...
+
+  render () {
+    return (
+      <View style={AppStyles.view}>
+        <Text style={AppStyles.hintText}>Data to sign</Text>
+        <Text style={AppStyles.valueText}>{this.props.text}</Text>
+        <Button
+          onPress={this.props.onNextButtonPressed}
+          title='Next'
+          color='gree'
```
tomusdrw: typo

src/components/DataDetails.js
```diff
@@ -0,0 +1,43 @@
...
+          accessibilityLabel='Enter Pin'
```
tomusdrw: Not in line with `title`

src/containers/DataDetails.js
```diff
@@ -0,0 +1,31 @@
...
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+'use strict'
+
+import { connect } from 'react-redux'
+import { Actions } from 'react-native-router-flux'
+import DataDetails from '../components/DataDetails'
+
+const DataDetailsContainer = connect(state => ({
+  text: 'dupa'
```
tomusdrw: Should most probably take something out of state :)

src/actions/scanner.js
tomusdrw: That's how it's done in Parity:
https://github.com/paritytech/parity/blob/2fd44e6ce46a1ac546cd9b52d3dc6130b6a368b1/js/src/views/Signer/components/SignRequest/signRequest.js#L144

I currently see that you're not prepending the text before hashing, should it be done in the UI?

src/components/DataDetails.js
debris: ?

tomusdrw: I thought that `accessibilityLabel` is completely different than `title` and it's not relevant on that screen, but on the second though it might actually be correct.

PR #90, opened 2018-04-09
updated app icon to Parity logo

bnvk:
Updated to use [app icon](https://github.com/paritytech/design-assets/blob/master/icons/app-icon.png) based on Parity's branding!


maciejhirsz:
@debris LGTM but you might want to have a look

PR #134, opened 2018-08-16
Various CI fixes for parity-signer

ddorgan:
- Move SDK install to container
- Enable key creation and apk build
- Collect artfifacts in gitlab

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -28,7 +28,7 @@ cache:
 #    - rust 
 android-build:
   stage: build
-  image: parity/parity-android
+  image: parity/android-sdk
   only:
     - beta
     - tags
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -37,21 +37,15 @@ android-build:
...
-    - echo "oracle-java8-installer shared/accepted-oracle-license-v1-1 select true" | sudo debconf-set-selections
-    - apt-get install -y oracle-java8-installer
-    - yes | /opt/android-sdk/tools/bin/sdkmanager --licenses || true
+    - apt-get install -y nodejs
     - echo "ndk.dir=/usr/local/android-ndk-r16b" > android/local.properties
     - echo "sdk.dir=/opt/android-sdk" >> android/local.properties
     - npm install -g yarn
     - bash ./setup_linux.sh
-    - npm run android
+    - bash ./gen_key.sh
```
fckt: is there any good reason why we generating new keys each time?

ddorgan: Hi @fckt just for the sake of not actually hard coding a key. Although we are using the same password etc... each time. 

I could also switch it to a static key, but have a habit of generating them at runtime if possible.

fckt: I mean, in the final version of CI we'll have to sign with the same private key, so we need to think on how to organize it

Tbaut:
What's the status of this PR?

fckt:
@Tbaut it's morphing 🙂

ddorgan:
Which encryption type and key size is required?

fckt:
@ddorgan take a look https://developer.android.com/studio/publish/app-signing

Tbaut:
Should we go on with this PR?

Tbaut:
@ddorgan this PR has been open for several months, can you update me on its state or close it if you think it can't be merged any time soon?

ddorgan:
@Tbaut can we merge this please. Once that's done and all is working I'll open a new PR to introduce:

1)  Remove self signing and use the correct apk key
2) Enable uploads to s3

But these will take a few changes so it may be best to merge this as is. Agreed?

PR #170, opened 2018-10-12
Double 0x fix in address

fckt:
~~NOTE:~~
~~TODO: recompile the iOS binary in order to not trim the first byte of the address~~

### DIFF
--- a/rust/signer/Makefile
+++ b/rust/signer/Makefile
@@ -1,5 +1,5 @@
 NDK_STANDALONE = $(shell pwd)/../../NDK
-ARCHS_IOS = i386-apple-ios x86_64-apple-ios armv7-apple-ios armv7s-apple-ios aarch64-apple-ios
+ARCHS_IOS = armv7-apple-ios armv7s-apple-ios aarch64-apple-ios
 ARCHS_ANDROID = aarch64-linux-android armv7-linux-androideabi i686-linux-android
 LIB=libsigner.a
 
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -99,9 +99,9 @@ android {
     defaultConfig {
         applicationId "io.parity.signer"
         minSdkVersion 16
-        targetSdkVersion 22
-        versionCode 1
-        versionName "1.0"
+        targetSdkVersion 26
+        versionCode 2
+        versionName "2.0"
         ndk {
             abiFilters "armeabi-v7a", "x86"
         }
@@ -118,7 +118,6 @@ android {
     }
     buildTypes {
         release {
-            signingConfig signingConfigs.release
         }
     }
     splits {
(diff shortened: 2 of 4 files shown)


rust/signer/Makefile
```diff
@@ -1,5 +1,5 @@
 NDK_STANDALONE = $(shell pwd)/../../NDK
-ARCHS_IOS = i386-apple-ios x86_64-apple-ios armv7-apple-ios armv7s-apple-ios aarch64-apple-ios
```
debris: this change disables iPhone simulator and iPhone 6

debris: this change disables iPhone simulator and iPhone 6

and it's unrelated to the scope of this pull request

fckt: Good points, you're right.

> and it's unrelated to the scope of this pull request

to be able to merge this, I had to rebuild `libsigner.a`, as [this particular change](https://github.com/paritytech/parity-signer/pull/170/files#diff-bcf08b8f1f5d5966fe583d09d70a07ecR28) isn't compatible with the old `libsigner.a`. 

> this change disables iPhone simulator and iPhone 6

Right, I had to come up with a decision which was't easy to made there. It'll be not possible to make an `libsigner.a` for `i386-apple-ios` and `x86_64-apple-ios` currently due to https://github.com/briansmith/ring/issues/184 , as I mentioned in the commit message. I've created an issue for that https://github.com/paritytech/parity-signer/issues/176

fckt: > part of this pull request should be reverted

In the current of uncertainty (I, personally don't know how much time it'll take to fix that issue) I'd made a tradeoff decision to let it be able to build, even we don't have a simulator support ATM.

debris: @fckt I don't understand. I used to compile rust for iPhone simulator and with ring. If we updated ring and it broke building for those devices, we should revert this update.

afaik several people in parity and web3 foundation use signer on iphone 4. We cannot make those devices useless

fckt: > afaik several people in parity and web3 foundation use signer on iphone 4

If the issue (https://github.com/paritytech/parity-signer/issues/176) affects the real devices, we should revert this update, absolutely no question there. But what I can't understand is how this could affect ARM-devices (as we're disabling `i386-apple-ios` and `x86_64-apple-ios` targets only)?

fckt: > afaik several people in parity and web3 foundation use signer on iphone 4. We cannot make those devices useless

I think, I've got your message. What you mean here (I guess), that in case we don't have the actual device we wouldn't be able to test the app without simulator. Right?

fckt: The problem is that, we can't support both binaries at the same time, due to https://github.com/paritytech/parity-signer/pull/164 issue.. So I'd better find a way, how to solve the problem without downgrade and maintain outdated libs for iOS specifically.

PR #185, opened 2018-11-09
a hardware wallet -> a cold storage

fckt:
https://en.bitcoin.it/wiki/Hardware_wallet
https://en.bitcoin.it/wiki/Cold_storage

We shouldn't claim Parity Signer as a something which "Turns your smartphone into a hardware wallet" . Simply because it confuses user: in the first place, Parity Signer doesn't turn user's phone into a *wallet*, and it doesn't turn it into a *hardware wallet* especially.

We need to fix it everywhere we declare it so (blogposts, wikis etc.).

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 [google-play-badge]: ./res/google-play-badge.png
 [google-play-url]: https://play.google.com/store/apps/details?id=io.parity.signer
 
-# Parity Signer - Turn your smartphone into a hardware wallet
+# Parity Signer - Turn your smartphone into a cold storage
 
 Parity Signer is a mobile application that turns a smartphone into a secure air-gapped wallet also called cold storage. It means that, once the app has been installed, users can create Ethereum accounts, sign transactions and transfer funds from and to these accounts without any sort of connectivity enabled on the device. Wifi, Mobile Network, Bluetooth can be shut down and even removed physically to ensure that the mobile phone containing these accounts will not be exposed to any online threat. Have a look at the tutorial on our wiki to learn how to [use MyCrypto together with Parity Signer](https://wiki.parity.io/Parity-Signer-Mobile-App-MyCrypto-tutorial).
 


tomusdrw:
I disagree. Parity Signer is air-gapped hardware wallet (as the wiki pages states, this could be a mean of cold storage: "Use a offline Bitcoin Hardware wallet") - we satisfy all 4 points in the linked wiki page.

Why do you claim it's not a wallet? The only difference being that there is no way to create a transaction from within the app? If that's the case it's trivial to add, but if we do that only to be compliant with some vague definition of a "wallet" IMHO there its no point.

ligi:
No this PR is correct. The difference is simple: Physical access protection. If you loose your hardware wallet - you are not in the same trouble as you are when you loose your parity signer phone.

fckt:
> Why do you claim it's not a wallet?

One important property of the wallet is that typically wallets contains coins.. Isn't MyCrypto a wallet? Isn't it a confusion when user (for whatever reason) have to use 2 wallets (with a current wording) to send a transaction?

Tbaut:
I disagree. 
> Simply because it confuses user

I've not seen any such confusions in my talks, on Twitter or Reddit. I think on the opposite that the term hardware helps us a lot, because users who know what a hardware wallet is, will associate it with what they know, and Ledger or Trezor do need an additional software to send their transactions.

fckt:
> and Ledger or Trezor do need an additional software to send their transactions.

that's a valid argument.

>  I think on the opposite that the term hardware helps us a lot, because users who know what a hardware wallet is, will associate it with what they know

We (at Parity) are here not to bullshit people in favor to have more users.

So, we have to be as strict as possible in our wordings, **in respect to this specific tool**. Consider Signer as a tool, not as a complete product. It's not our goal to have a great marketing or a lot of users with Signer, just because the "market" is not really defined for. What we really want - it is an **adoption**. With **this specific tool**, the adoption can be archived **only and only if** we remain to be as correct as possible in **the tool design** and **wording** as well.

Maybe the "cold storage" isn't the best, but it is definitely more correct and less dangerous wording (in respect to Signer) than "hardware wallet" in respect to Parity Signer IMHO.

tomusdrw:
>  Physical access protection. If you loose your hardware wallet - you are not in the same trouble as you are when you loose your parity signer phone.

@ligi That's a fair point, although on the page linked by Alexey, you can find at least "Pi wallet - cold storage", which suffers from exactly the same issue, yet is listed as "Hardware Wallet".

So since the definition of Hardware Wallet is not super strict about what security model it assumes, I'd be  in favour of keeping the name, but adding a comprehensive section of our threat model and security assumptions (should contain comparison with other devices as well).

I'm 100% sure that we should never mention Parity Signer as a safe replacement for hardware wallets like Ledger or Trezor, but calling it merely a "Cold Storage" is equally misleading - in my view cold storage is something that should be used infrequently, potentially storing large amounts of coins (sic!) and using such storage is generally hard (involves some additional steps).

fckt:
> I'm 100% sure that we should never mention Parity Signer as a safe replacement for hardware wallets like Ledger or Trezor, but calling it merely a "Cold Storage" is equally misleading - in my view cold storage is something that should be used infrequently, potentially storing large amounts of coins (sic!) and using such storage is generally hard (involves some additional steps).

good points!

ligi:
Why don't you just call it "offline signer"? I think this is what it actually is.

Yes wording is hard - b it a hardware wallet as then people would assume it has the same security guarantees as a TREZOR - wut IMHO there should be a differentiation in naming between a device like a TREZOR and this model (which has it's use cases) - I also support this in [WallETH](https://walleth.org) - because I think it is useful - but I would never call it a hardware wallet as then people would assume it has the same security guarantees as a TREZOR - which it does not have. But I would also not call this raspberry PI solution a hardware wallet. Yes the lines are blurry - but IMHO still important to draw them - so thanks for the initiative @fckt 

While we are at it. I would also change:

> Parity Signer is a mobile application that turns a smartphone into a **secure** air-gapped wallet also called cold storage. 

into 

> Parity Signer is a mobile application that turns a smartphone into a **reasonably secure** air-gapped wallet also called cold storage.

take them as an example: https://www.qubes-os.org

> A **reasonably secure** operating system

fckt:
>reasonably secure

I like the term!

AtLeastSignificant:
> Isn't MyCrypto a wallet?

No, it's not.  And despite the name, neither is MyEtherWallet.  They don't store any value (sensitive information), which is the primary requirement for using the analogy at all IMO.

AtLeastSignificant:
You could view this as a debate about definitions, and in that case it really doesn't matter how confusing or wrong the terminology is - whatever people use will become the de-facto "correct" term.

If you look at it in a more useful manner though, as an analogy, then MEW/MC aren't analogous to wallets at all. No amount of incorrect usage of the word changes that.  

I'm of the opinion that terminology should actually be useful and intuitive to users, not a maze of illogical rhetoric.  The best time to fix bad terminology is yesterday, and the second best time is today.

See https://ethereum-magicians.org/t/what-defines-a-hardware-wallet/1854/4, I've already written up a small glossary of which terms I use and why - as well as terms I find confusing/conflicting.

General-Beck:
I'll add my 5 satoshi to this discussion. A cold wallet is a pair from an address-private key that has never been used anywhere and is not tied directly to any device or program.

AtLeastSignificant:
> A cold wallet is a pair from an address-private key that has never been used anywhere and is not tied directly to any device or program. and no matter what it is generated and how it is stored.

What do you mean by "never sued anywhere"?  If I use my everyday desktop to generate a PK on MEW, then save that PK to my desktop but never use it to sign a transaction, is the corresponding address a "cold wallet" until I send something out of it?  If that's the case, this definition is not only misleading, it's completely contrary to what we generally ascribe to cold wallets.

Security has very little to do with how much something is used, it's about *how* it's used.

General-Beck:
@AtLeastSignificant A simple example, I will write this in a will to my grandchildren. Then they will decide themselves, they will use these assets and it does not matter to me what wallet will be used, but I will not need to store a smartphone for them, which will have the battery die or it will be a coinbase account "inherited" (lol)

fckt:
@Tbaut could I ask you to comment the resolution on the issue? Just to make it clear for people involved in the discussion

Tbaut:
Sure, I want to cut the discussion short. We don't say parity signer *is* a hardware wallet, we say "you can *turn your phone into* a hardware wallet" which is fine.

AtLeastSignificant:
> Sure, I want to cut the discussion short. We don't say parity signer _is_ a hardware wallet, we say "you can _turn your phone into_ a hardware wallet" which is fine.

It's really not fine... but glad we have *decided* to use misleading incorrect verbiage for the sake of marketing and not just leaving it up in the air.  I'll add it to my list of "except for..." when talking about commonly accepted terms / meanings.

PR #187, opened 2018-11-16
Proper lock on went into background

fckt:


Tbaut:
It did not do the trick for me on android, I could take a screenshot and the screen of the app was visible on the "application overview".  (Signal just above is blank for instance).

image

ligi:
to make it behave like signal there you need to do:

window.setFlags(FLAG_SECURE, FLAG_SECURE)

Tbaut:
I couldn't manage to implement this fix in a timely manner. I messed up with the main Activity but couldn't make it compile. Any chance you can help with that @ligi :) ?

tomusdrw:
@fckt what about ligi's suggestion?

fckt:
> @fckt what about ligi's suggestion?

oh, I haven't tested on Android (due to lack of device). Sure, I'm going to do it now.

fckt:
> @fckt what about ligi's suggestion?

Addressed

PR #192, opened 2018-11-27
Attempt to have a better autofocus

### LINKED ISSUE #189: scanner focus
when using the app in an older phone, there are difficulties scanning qr codes. The camera inside the parity app is a little blurry. I've tried another camera app, and it regulates the focus automatically, which results in a clearer view. 
adding the automatically focus, would help older phones focus better when scanning qr codes

Tbaut:
- Attempt to close #189 from [comment](https://github.com/react-native-community/react-native-camera/issues/675#issuecomment-334109803)


fckt:
I'll check on my old device

Tbaut:
Did you have the chance to test this on a old device? otherwise I can propose the user that raised this issue to test with an apk.

ltfschoen:
@Tbaut Since it's camera and QR related we could also try updating to the latest:
* https://www.npmjs.com/package/react-native-camera
* https://www.npmjs.com/package/react-native-qrcode

PR #198, opened 2018-12-12
Replace failing "yarn run cli.." by "react-native.."

Tbaut:
It's failing and "has never worked"

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,9 +5,7 @@
   "license": "GPL-3.0",
   "scripts": {
     "build-rust-ios": "cd rust/signer && make ios",
-    "ios": "yarn run build-rust-ios",
     "build-rust-android": "cd rust/signer && make android",
-    "android": "yarn run build-rust-android",
     "start": "yarn run ios",
     "test": "jest",
     "lint": "",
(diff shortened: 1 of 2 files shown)


maciejhirsz:
this basically makes `ios` and `build-rust-ios` equivalent (ditto for android), we should remove one of of the commands entirely IMO.

Tbaut:
right, I kept the most descriptive ones.

maciejhirsz:
We could keep the short ones if we change them to:

```json
"ios": "npm run build-rust-ios && react-native run-ios",
"android": "npm run build-rust-android && react-native run-android",
```

Tbaut:
it makes sense, thanks for the guidance :)

package.json
```diff
@@ -4,10 +4,8 @@
   "private": true,
   "license": "GPL-3.0",
   "scripts": {
-    "build-rust-ios": "cd rust/signer && make ios",
-    "ios": "yarn run build-rust-ios && yarn run cli run-ios",
-    "build-rust-android": "cd rust/signer && make android",
-    "android": "yarn run build-rust-android && yarn run cli run-android",
+    "ios": "npm run build-rust-ios && react-native run-ios",
```
maciejhirsz: This is refering to `build-rust-ios`, but `build-rust-ios` is removed :see_no_evil:.

PR #200, opened 2018-12-20
Sign apk and upload result to s3

ddorgan:
- Now signs the apk in CI 
- Uploads result to s3 bucket
- Needs a few debug lines removed once the issue below is resolved.

So using my local machine my more recent android_sdk, android_ndk, yarn and node (and in the android/android-sdk docker image) the resulting apk files won't load in android.

Can you let us know what exact versions are being used with the release apk's please.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -55,7 +55,6 @@ android-build:
     - cat android/gradle.properties
     - bash ./setup_linux.sh
     - bash ./gen_key.sh
-    - npm run android
     - cd android && ./gradlew assemble && cd ..
     - find . | grep apk
   tags:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -16,6 +16,7 @@ variables:
     - beta
     - tags
     - master
+    - dd-sign-apk
 
 #android-test:
 #  stage: test
@@ -44,6 +45,7 @@ android-build:
     - stable
     - triggers
     - master
+    - dd-sign-apk
   script:
     - curl -sL https://deb.nodesource.com/setup_8.x | sudo -E bash -
     - apt-get install -y nodejs
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,14 +1,22 @@
 stages:
   - test
   - build
+  - release
+
 variables:
   GIT_DEPTH: "3"
   SIMPLECOV: "true"
   RUST_BACKTRACE: "1"
   RUSTFLAGS: ""
   CARGOFLAGS: ""
-cache:
-  untracked: true
+
+.releaseable_branches:             # list of git refs for building GitLab artifacts (think "pre-release binaries")
+  only:                            &releaseable_branches
+    - stable
+    - beta
+    - tags
+    - master
+
 #android-test:
 #  stage: test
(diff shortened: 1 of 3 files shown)


fckt:
> Can you let us know what exact versions are being used with the release apk's please.

Use the latest stable versions of deps, as long as it builds ok.

> So using my local machine my more recent android_sdk, android_ndk, yarn and node (and in the android/android-sdk docker image) the resulting apk files won't load in android.

Which message it outputs?

Tbaut:
As far as I'm concerned there is no other message than the generic force close message "the app has stopped working". The installation process ran smoothly though..

ligi:
> * Now signs the apk in CI

so you are exposing your signing key to the CI? Which CI is this?

better think about something like this: https://github.com/airgap-it/airgap-raspberry-apk-signer

be really careful here as currently there is no way to practically revoke a signing key

fckt:
> Will try with the latest yarn / node / sdk / ndk locally. One question, which node version are you running?

I'm sure we all run a different ones. Mine:

```
node -v
v10.7.0
```

ddorgan:
It's currently on one trusted runner with a modified docker image. Let me speak to Kirill about offline other signing methods.

Tbaut:
Any news on that front @ddorgan?

ddorgan:
Guys I keep having problems with this.

Please send me the *exact* version of the software that works. When I build this using the most recent android studio I cannot use the apk file. The dockerfile is doing the exact same and has the same problem.

fckt:
We are on it don't worry, @Tbaut

.gitlab-ci.yml
```diff
@@ -35,20 +45,33 @@ android-build:
...
     - echo "ndk.dir=/usr/local/android-ndk-r16b" > android/local.properties
     - echo "sdk.dir=/opt/android-sdk" >> android/local.properties
     - npm install -g yarn
+    - cat android/gradle.properties
     - bash ./setup_linux.sh
     - bash ./gen_key.sh
-    - cd android && ./gradlew assembleRelease
-    - pwd
-    - ls
+    - cd android && ./gradlew assemble && cd ..
```
fckt: I use `assembleRelease`

fckt:
> Please send me the exact version of the software that works. When I build this using the most recent android studio I cannot use the apk file. The dockerfile is doing the exact same and has the same problem.

Take a look https://github.com/paritytech/parity-signer#requirements

maciejhirsz:
I managed to get an unsigned release APK build with gradle, but it fails to install (asks for permissions, shows progress bar for a bit and just says "App not installed."). I'll spend some more time on that to see if we can provide unsigned builds from the CI.

Tbaut:
Talking with @ddorgan the othre day, I managed to build a working APK (unsigned) from master using `build-tools;27.0.3` and `node v11.0.0`.

maciejhirsz:
Oh cool! I'll check that.

Tbaut:
FWIW I ran into the problems you had as well. It seems that a fix has been integrated in  later versions of  react-native https://github.com/facebook/react-native/issues/18357

Tbaut:
Closing this one as the signing process should most probably remain manual

PR #204, opened 2019-01-16
Additional account validation added

### LINKED ISSUE #127: No recovery words on Backup Account screen
screen shot 2018-08-06 at 13 23 46

fckt:
fixes #127

### DIFF
--- a/src/screens/AccountNew.js
+++ b/src/screens/AccountNew.js
@@ -43,6 +43,7 @@ import {
   DEFAULT_NETWORK_COLOR
 } from '../constants';
 import colors from '../colors';
+import { validateSeed } from '../util/account';
 
 export default class AccountNew extends React.Component {
   static navigationOptions = {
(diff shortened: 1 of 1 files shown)


ltfschoen:
I think it would help to disable the Next Step button by default and enable it in a different colour when all the input fields are valid.

I also think we should highlight what form fields still need to be completed before the Next Step button will be enabled.

Tbaut:
We will not do a release for this fix. The release will be done for the next feature that we expect to have which is Bip39 support. This fix will not be included then.

fckt:
> I think it would help to disable the Next Step button by default and enable it in a different colour when all the input fields are valid.
> 
> I also think we should highlight what form fields still need to be completed before the Next Step button will be enabled.

It's just a quick fix for the exact annoying problem. If it doesn't makes the app better - we can just leave it as it is and close this PR.

ltfschoen:
I just tried these changes and they work great!

PR #217, opened 2019-01-30
[WIP][iOS] Secure Enclave Storage impl

fckt:


maciejhirsz:
I thought we can't really use Secure Enclave? The key is created on the device (no way to backup, which might be ok for some usecases) and I'm not sure it supports secp256k1 (to say nothing of the new `sr25519` for Substrate). 

[The list of all supported algos.](https://developer.apple.com/documentation/security/seckeyalgorithm)

fckt:
Oh, sry, hadn't had a chance to answer on that sooner.

So, we made some research on the topic. Basically the idea is to give users an option to use a `Secure Enclave` for additional security. It's based on ability of using an additional entropy from the `Secure Enclave` (for iOS only ATM).

maciejhirsz:
Do you have any links to resources about using RNG from secure enclave? I'm pretty sure you can't get the randomized keys out of it (it's kind of the whole point).

maciejhirsz:
Just talked to Kirill about it, gonna open an issue.

PR #194, opened 2018-11-29
Rust bindings refactoring and bip39_mnemonic impl

fckt:
~~there is a problem with a Java build~~

PR #221, opened 2019-02-12
[WIP] BIP39-rs25519

fckt:
Merge #194 first, please


maciejhirsz:
I'll grab stuff from #194 (with attribution), but the branches are likely to diverge. `bip39-rs` 0.6 has a different API, and Substrate is replacing Ed25519 with Ristretto.

maciejhirsz:
Will re-open the PR once there is actually some code here, not just dependency hell resolution.

PR #223, opened 2019-02-20
BIP32/39/44 and Substrate (sr25519) Support

maciejhirsz:
### Rust

+ [x] Remove dependency on `parity-ethereum`, use pure Rust crates when possible.
+ [x] Handle BIP39 phrases.
+ [x] Add BIP32 derivation with BIP44 paths.
+ [x] Generate BIP39 phrases by default.
+ [x] Attempt to derive with BIP32/44 first, fall back to parity derivation.
+ [x] Separate functions for deriving BIP/Parity.
+ [ ] ...

### React

+ [ ] ...

### Issues

+ [x] Creating new accounts is borked (cannot continue after entering name and selecting icon).
+ [x] Fields for entering and displaying phrases are too small (BIP39 requires 24 words for good entropy).
+ [x] Account recovery produces wrong keys? Are we doing previews in JS?
+ [ ] Not handling different networks in BIP44 yet!


maciejhirsz:
Moving to a cleaner PR.

Issue #224, opened 2019-02-23
Remove Tab navigation


Tbaut:
Add a "scan" button to the account list.

Issue #162, opened 2018-10-10
[UX] The scanner screen is shown 1st when launching the app


Tbaut:
It feels unnatural to have the scanner view as first screen when launching the app for the 2nd time.
This should be the account view.

Tbaut:
Will be solved by https://github.com/paritytech/parity-signer/issues/161

5chdn:
totally!

Issue #142, opened 2018-09-17
The back button acts as a rewind and brings to unexpected screens


Tbaut:
On Android.

If you create an account (select network, name, enter a pin code twice and copy the seed phrase), then go to account to see the QR code for instance. 
Now tap the "back" button a couple times and you'll get back to the PIN configuration of the account etc.. 

The back button should instead get back to the account overview, and if tapped again ask to quick the app (or quit it without asking).

fckt:
have you tested the build for iOS? @Tbaut

Issue #225, opened 2019-02-23
Enhance menus


Tbaut:
- Move "New accounts" to a small button on the `accounts list` screen to reach
  - New Account
  - Recover Account

- Move "Edit" to a multi-select menu at the top right and allow to reach:
  - Edit
  - Change Pin
  - Backup
  - Delete

![image](https://user-images.githubusercontent.com/33178835/53350893-6ce20800-3920-11e9-8cbc-6c6222e4b82e.png)
![image](https://user-images.githubusercontent.com/33178835/53351017-b894b180-3920-11e9-88f7-d8f33f46ded4.png)

fckt:
Proposed UI doesn't follow IOS nor Android mobile guidelines.

Tbaut:
I think that's not a big issue, even [gmail](https://1ad20d09099cef6447ca02f7da2adf1f007b9244-m.proxy.startpage.com/npd/mbsuofdejpseob/xxx/ST/mDFUUqLFUmFSkCKuB80Kj980Q//////////SURFLYROOT//////////sites/androidcentral.com/files/styles/xlarge/public/article_images/2016/06/recover-gmail-contacts-android-screens-05.jpeg?SURFLY=R?itok=sczYqX45) does it that way. What I'm interested about is to not surprise users for such behaviors.

fckt:
on iOS it's different, for sure

miha999rus777:
Xz

PR #226, opened 2019-02-23
Tbaut remove tab navigation

Tbaut:
This PR mainly removes the tabs (Account and Scanner) and sets the AccountList screen as the main screen.

It also fixes a couple things related to the navigation such as #142 :
- closes https://github.com/paritytech/parity-signer/issues/224
- closes https://github.com/paritytech/parity-signer/issues/162
- closes https://github.com/paritytech/parity-signer/issues/142

Tested on Android with Tx and message signing.

There are now 2 buttons (that's at least 1 too many) on the account list, I'll tackle this (https://github.com/paritytech/parity-signer/issues/225) in a separate PR

Question: how should I proceed to avoid committing `package-lock.json`, I read it shouldn't be in `gitignore`.. what about this annoying wizard in `package.json` ? :)


src/screens/TxDetails.js
```diff
@@ -96,9 +90,6 @@ export class TxDetailsView extends React.PureComponent {
           title={this.props.sender.name}
           address={this.props.sender.address}
           chainId={this.props.sender.chainId}
-          onPress={() => {
-            this.props.onPressAccount(this.props.sender);
```
maciejhirsz: Has there been any issues with navigation here?

(I don't think I ever wanted to navigate from tx screen to account by tapping the icon, or expected that to work that way, so I think it's fine, just curious about the reasons)

Tbaut: There has been no issue opened for this indeed. I stumbled upon this and would not expect this either, hence the removal. I have the bad habit to tackle several issues at once :). I did my best not to shoot in too many directions for this one.

fckt: > I have the bad habit to tackle several issues at once :)

how sweet is it

fckt:
have you tested the build for iOS? @Tbaut

Tbaut:
Not yet, I'll do it after https://github.com/paritytech/parity-signer/issues/225

fckt:
not the best practice

PR #232, opened 2019-03-28
Fixed Rust dependencies + BIP39 support

maciejhirsz:
This is squashed #223 to date with fixed Rust dependencies that should hopefully forever solve all of our build issues (no linked C code).

It includes the BIP39 for Ethereum (new phrases are BIP39, recovery attempts BIP39 recovery and falls back to Parity phrases) along with minor UI changes (phrase validation and making the boxes bigger as 24 word BIP39 phrases are just longer).

#223 will be rebased after this is merged.

### DIFF
--- a/rust/signer/Cargo.toml
+++ b/rust/signer/Cargo.toml
@@ -8,7 +8,7 @@ edition = "2018"
 [dependencies]
 base64 = "0.10.1"
 blockies = "0.2"
-ethsign = { git = "https://github.com/tomusdrw/ethsign.git", default-features = false, features = ["pure-rust"] }
+ethsign = { version = "0.6.1", default-features = false, features = ["pure-rust"] }
 jni = { version = "0.10.2", default-features = false, optional = true }
 libc = "0.2"
 rlp = { version = "0.3.0", features = ["ethereum"] }
--- a/rust/signer/Cargo.toml
+++ b/rust/signer/Cargo.toml
@@ -3,17 +3,23 @@ name = "signer"
 version = "0.1.0"
 authors = ["debris <marek.kotewicz@gmail.com>"]
 build = "build.rs"
+edition = "2018"
 
 [dependencies]
-libc = "0.2"
-rustc-serialize = "0.3"
-ethkey = { git = "https://github.com/paritytech/parity-ethereum" }
-ethstore = { git = "https://github.com/paritytech/parity-ethereum" }
-rlp = { version = "0.3.0", features = ["ethereum"] }
-tiny-keccak = "1.1"
+base64 = "0.10.1"
 blockies = "0.2"
-parity-wordlist = { git = "https://github.com/paritytech/wordlist" }
-jni = { git = "https://github.com/prevoty/jni-rs", default-features = false, optional = true }
+ethsign = { git = "https://github.com/tomusdrw/ethsign.git", default-features = false, features = ["pure-rust"] }
+jni = { version = "0.10.2", default-features = false, optional = true }
+libc = "0.2"
+rlp = { version = "0.3.0", features = ["ethereum"] }
+rustc-hex = "2.0.1"
+schnorrkel = "0.0.0"
+serde = "1.0"
+serde_json = "1.0"
+sha2 = "0.8"
+tiny-bip39 = { version = "0.6.1", default-features = false }
+tiny-hderive = "0.1"
+tiny-keccak = "1.4"
 
 [lib]
 name = "signer"
@@ -21,7 +27,3 @@ crate-type = ["staticlib", "cdylib"]
 
 [features]
 default = ["jni"]
-
-# https://github.com/DaGenix/rust-crypto/issues/383
-[replace]
-"rust-crypto:0.2.36" = { git = "https://github.com/debris/rust-crypto" }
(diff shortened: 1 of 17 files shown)


rust/signer/Cargo.toml
```diff
@@ -3,25 +3,27 @@ name = "signer"
 version = "0.1.0"
 authors = ["debris <marek.kotewicz@gmail.com>"]
 build = "build.rs"
+edition = "2018"
 
 [dependencies]
+base64 = "0.10.1"
+blockies = "0.2"
+ethsign = { git = "https://github.com/tomusdrw/ethsign.git", default-features = false, features = ["pure-rust"] }
```
tomusdrw: Let's release it and use the crates.io version.

rust/signer/src/lib.rs
```diff
@@ -159,32 +141,35 @@ pub unsafe extern fn blockies_icon(blockies_seed: *mut StringPtr) -> *mut String
...
 
 #[no_mangle]
 pub unsafe extern fn encrypt_data(data: *mut StringPtr, password: *mut StringPtr) -> *mut String {
   let data = (*data).as_str();
-  let password = Password::from((*password).as_str());
-  let crypto = Crypto::with_plain(data.as_bytes(), &password, 10240).unwrap();
-  Box::into_raw(Box::new(crypto.into()))
+  let password = Protected::new((*password).as_str().as_bytes().to_vec());
+  let crypto = Crypto::encrypt(data.as_bytes(), &password, CRYPTO_ITERATIONS).unwrap();
+  Box::into_raw(Box::new(serde_json::to_string(&crypto).unwrap()))
```
tomusdrw: Might be good to have `expect`s instead of unwraps

rust/signer/src/lib.rs
```diff
@@ -199,13 +184,12 @@ pub unsafe extern fn decrypt_data(encrypted_data: *mut StringPtr, password: *mut
 #[cfg(feature = "jni")]
 #[allow(non_snake_case)]
 pub mod android {
-  extern crate jni;
+  // extern crate jni;
```
tomusdrw: remove?

rust/signer/src/lib.rs
```diff
@@ -318,117 +293,117 @@ pub mod android {
...
-          jvm.env(),
-          any_class,
-          jni_crypto_str,
-          env.new_string("wrong password").unwrap()
-        );
-        assert_eq!(env.exception_check().unwrap(), true);
-      }
-    }
-  }
+  // #[cfg(test)]
```
tomusdrw: ?

rust/signer/src/lib.rs
```diff
@@ -318,117 +293,117 @@ pub mod android {
         env.new_string(String::from_utf8_unchecked(decrypted)).expect("Could not create java string").into_inner()
       },
       Err(_) => {
-        let result = env.new_string("").expect("second result to be created").into_inner();
-        env.throw(new_exception(&env)).expect("second throw failed");
+        let result = env.new_string("").expect("Could not create java string").into_inner();
+        env.throw(new_exception(&env)).expect("second trhow failed");
```
tomusdrw: typo

rust/signer/src/util.rs
```diff
@@ -14,8 +14,23 @@
...
 
+use tiny_keccak::Keccak;
 use libc::size_t;
 
+pub trait Keccak256<T> {
+	fn keccak256(&self) -> T where T: Sized;
+}
+
+impl Keccak256<[u8; 32]> for [u8] {
+	fn keccak256(&self) -> [u8; 32] {
```
tomusdrw: I think such function is already part of `tiny_keccak`: https://docs.rs/tiny-keccak/1.4.2/tiny_keccak/fn.keccak256.html

src/util/account.js
```diff
@@ -42,32 +42,11 @@ export function validateSeed(seed) {
...
-      return {
-        valid: false,
-        reason: `The word "${word}" is not from the word list`
-      }
-    }
-    set.add(word);
-  }
-  if (set.size < 11) {
+
+  if (words.length < 11) {
```
tomusdrw: What happened with the previous validation? It seemed quite nice

rust/signer/src/lib.rs
maciejhirsz: Unfortunately they aren't very useful here since you never see the message when the mobile app crashes.

I want to abstract out the bridge glue code so type conversions are automatic and have some way of propagating errors up to JS without panicing, then get rid of all `unwrap`s. Another thing is the duplicated logic between iOS and Android that I want to get rid of as well. That said, that will be a separate PR (likely part of #223).

src/util/account.js
maciejhirsz: It broke for BIP39 since it uses a different dictionary. This is pretty MVP, we want to add the double-dictionary hinting mechanism that figures out which phrase type you are entering after 1-n words, right now the checks are done in the "backend". I did reintroduce checking for extra whitespace just now though, as that is useful for both.

maciejhirsz:
Thanks for the review @tomusdrw, I've addressed all points I think. Will wait for @Tbaut to look over the UI changes too.

fckt:
https://github.com/paritytech/parity-signer/pull/194#issuecomment-463702831
😔

PR #195, opened 2018-11-29
Add error messages when wrong pin is entered

Tbaut:
- fix https://github.com/paritytech/parity-signer/issues/156
- added error messages at pin creation when pin is too short or don't match
- added an error message at account unlock

Screenshots:
image
image
image

PR #234, opened 2019-04-02
Refactoring bindings

### LINKED ISSUE #219: Blank QR code on accounts page
On Nexus 5, Android 6.0.1, just downloaded the latest version from the play store. Same for all accounts

image

maciejhirsz:
This PR abstracts out bindings in a way that makes iOS and Android extern functions DRY. Type conversion and error handling (throwing exceptions in Java) is done [automagically via the `export!` macro](https://github.com/paritytech/parity-signer/blob/c21f1378e0bdb7fa9b867ec46fcc2d7071eda798/rust/signer/src/lib.rs#L80-L171). This also removes a whole bunch of `unwrap` and `expects` smell.

It also adds native function for generating QR codes in Rust, fixes #219.

+ [x] Introduce error handling for iOS.
+ [x] Update C header file.
+ [x] Rework string deallocating in Swift bridge.
+ [x] Test that iOS compiles and works (it won't until all of the above are done).
+ [ ] ~Make Android use `KeyPair` pointers similar to how iOS was done before.~


yjkimjunior:
@maciejhirsz is this ready for testing on iOS?

maciejhirsz:
@yjkimjunior yes please! :)

yjkimjunior:
Build fails for me because using an older version of Swift, are there any plans to update it to latest? otherwise I'll go downgrade my Xcode to test 📦

yjkimjunior:
@maciejhirsz the build is failing for me because signer uses an older version of Swift than Xcode10.2 supports. Should this be updated, or should I just go downgrade my Xcode?

maciejhirsz:
If we can upgrade it I think we should. I think most (all?) of our swift code is internally using just objective C, so there shouldn't be any breakage, but in case there is feel free to ping me on Riot.

yjkimjunior:
Works well for me!

yjkimjunior:
compiles and no more memory allocation errors, able to create and restore accounts, however, I am able to delete it after i type in just 4 out of 6 of my pin digits.  http://g.recordit.co/C7maeWWAWi.gif

Tbaut:
Also I see on your recording that the qr code doesn't load and remains white, right?

yjkimjunior:
Ah yes, that too.

maciejhirsz:
So everything compiles at last, and the QR code is showing. Need to figure out what's up with account deletion though.

yjkimjunior:
leaving this here reference http://g.recordit.co/co1sYl9LrF.gif

Tbaut:
I've touched this logic last here: https://github.com/paritytech/parity-signer/pull/195/files#diff-ed7c745a0e11d1facbcbeca7e4a54c0aL46 and it might not have been tested enough on iphone

maciejhirsz:
JS looks fine, just a bit convoluted.

It might be possible that errors aren't propagated properly for iOS on this branch, which is what we use to check if the pin is correct. I'll dig deeper tomorrow to see if I can either figure this out or at least mock something to confirm or reject the hypothesis I have now.

yjkimjunior:
@maciejhirsz works well for me!

Issue #233, opened 2019-04-02
Creating a new account sometimes creates an empty phrase account


maciejhirsz:
Title has it all. When creating a new account, if the empty phrase account has not yet been created, it occasionally will be. This seems to happen to me when I create a second account on the test device.

Tbaut:
I could reproduce on `master`, taking a look.

PR #235, opened 2019-04-08
Fix race condition at account creation

Tbaut:
- closes https://github.com/paritytech/parity-signer/issues/233
- Remove `onSubmitEditing={this.submit}` from `AccountPin.js` (android users) to prevent users from both tapping the button and hitting `enter` key, that both submit the new pin.


src/screens/AccountNew.js
```diff
@@ -45,7 +45,13 @@ export default class AccountNew extends React.Component {
 }
 
 class AccountNewView extends React.Component {
-  render() {
+
+  shouldComponentUpdate (nextProps) {
+    const hasEmptySeed = nextProps.accounts.state.newAccount.seed === "";
```
maciejhirsz: This looks like a really big code stink to me, even if it solves the problem. `shouldComponentUpdate` should be used to prevent re-renders when relevant props haven't changed, not to be a band-aid over bad state. It's also unclear to me why this is receiving the entire `accounts` store instead of just getting the `newAccount` itself and maybe a callback to inform upstream that account has been selected/created/whatever, separation of concerns is missing here. Anyway, fixing all of that is beyond the scope of this PR I think.

I'd suggest trying to find where the `newAccount.seed` is set to empty string in code, and instead set the entire `newAccount` to `null`, then prevent this component from rendering in the `render` method (it seems to already do that check `if (!selected) {` at line 58).

I think a proper way would be to set the whole `newAccount`

src/screens/AccountBackup.js
```diff
@@ -39,13 +39,18 @@ export default class AccountBackup extends React.PureComponent {
...
-class AccountBackupView extends React.PureComponent {
+class AccountBackupView extends React.Component {
   constructor(...args) {
     super(...args);
     this.handleAppStateChange = this.handleAppStateChange.bind(this);
   }
 
-  componentDidMount() {
+  shouldComponentUpdate (nextProps) {
+    const hasEmptySeed = nextProps.accounts.state.newAccount.seed === "";
```
maciejhirsz: Ditto here.

src/screens/AccountNew.js
Tbaut: The `newAccount` is cleared cleanly [here](https://github.com/paritytech/parity-signer/blob/master/src/stores/AccountsStore.js#L94), right after having been submitted, which seems to be the right place. 

I originally thought there was a race between locking the account [here at componentWillUnmount](https://github.com/paritytech/parity-signer/blob/master/src/screens/AccountBackup.js#L64). which triggers a seed clearing [here in `lockAccount`](https://github.com/paritytech/parity-signer/blob/master/src/stores/AccountsStore.js#L166).

Now another possibility is that we clicked both on the `Done` button and the `enter` key on an Android keyboard at the pin creation step. Both action submit the form [here](https://github.com/paritytech/parity-signer/blob/master/src/screens/AccountPin.js#L124). This allows me to reproduce the problem 100% of the time.

I can't manage to reproduce the problem on master by carefully click only on `Done`, or the `enter` key.

Tbaut: The reason why `accounts` is passed and not only `newAccount` is because the components for the `Backup` and `pin` are re-used for "View your recovery phrase" and "Change your pin".

src/screens/AccountBackup.js
```diff
@@ -55,16 +60,18 @@ class AccountBackupView extends React.PureComponent {
...
 
-  componentWillUnmount() {
+  async componentWillUnmount () {
     const { accounts } = this.props;
     const selected =
       accounts.getNew().address && accounts.getNew().address.length
         ? accounts.getNew()
         : accounts.getSelected();
-    accounts.lockAccount(selected);
+    await accounts.lockAccount(selected);
```
fckt: `lockAccount` is not an `async` function. That means `await` has no sense here.

src/screens/AccountBackup.js
```diff
@@ -55,16 +60,18 @@ class AccountBackupView extends React.PureComponent {
     }
   }
 
-  componentWillUnmount() {
+  async componentWillUnmount () {
```
fckt: No need to make this function `async`, see next comment.

src/screens/AccountBackup.js
Tbaut: Thanks, I'll remove that.

src/screens/AccountBackup.js
Tbaut: :+1: reverted that

Issue #237, opened 2019-04-13
Handle BIP39 on top of Parity wordlist suggestions for account recovery


Tbaut:
As soon as a at least 1 word is entered that belongs to a word list, and not the other, only suggest words from the adequate word list.

We need to make sure a fat finger typing on a word suggestion doesn't trigger the selection of a single word list.

PR #238, opened 2019-04-14
Add bip39 suggestions

Tbaut:
- closes https://github.com/paritytech/parity-signer/issues/237
- removed an animation as it made the screen "heavier" for little to no UX improvement
- fixed an array overflow bug gone under the radar when searching for `zoo` or anything toward the end of the word list
- commented as I could to make things a little more obvious
- removed a great amount of code that made things complex, e.g it allowed a user to change a word in the middle of the input, while this is a rare case IMHO, what matters most is the current word.

Testing can be done using a couple words such as `zone` (in both word list), `bike`,`zoo` (only in BIP39),  `abacus` (only in Parity)


Tbaut:
I've tried to handle cases where the input had an upper case, but I had weird behaviours that I can't explain:

Change [this line](https://github.com/paritytech/parity-signer/pull/238/files#diff-c53bd872d4e0c1e77990d48a42e83ed3R89) with:
`  const inputWordArray = value.length ? value.toLowerCase().split(' ') : [];`

- Typing `Bike`
- the suggestion shows `bike`
- tapping the suggestion replaces the `Bike` by `bike`, all good.
- Now press the space bar and the input field will show `bikeBike`

src/components/AccountSeed.js
```diff
@@ -17,154 +17,93 @@
...
+  }
+
+  renderSuggestions () {
+    const { value } = this.props;
+    // array of the words in the input field 
+    const inputWordArray = value.length ? value.split(' ') : [];
+    const suggestions = this.generateSuggestions(inputWordArray);
+
     return (
-      <Animated.View
```
fckt: no animation? :(

src/components/AccountSeed.js
```diff
@@ -17,154 +17,93 @@
...
-      selection: { start, end }
-    }
-  }) {
-    if (start !== end) {
-      return;
-    }
-    this.setState({ cursorPosition: start });
-  }
-
-  getSearchInput() {
```
fckt: why remove this logic?

src/components/AccountSeed.js
```diff
@@ -17,154 +17,93 @@
...
-    const { cursorPosition } = this.state;
-    let startFrom = cursorPosition;
-    // find the first space or start of string
-    while (startFrom > 0 && [' '].indexOf(value.charAt(startFrom - 1)) === -1) {
-      --startFrom;
-    }
-    return value.substring(startFrom, cursorPosition);
-  }
-
-  getWordPosition() {
```
fckt: What if user wants to make a correction in the middle? Why remove?

Tbaut: I really thought twice before removing it. It's a matter of making things simple. There was a ton of complex logic just for this use case, and I didn't think it was worth it.

src/components/AccountSeed.js
Tbaut: see description

src/components/AccountSeed.js
Tbaut: The animation was nice but barely visible because it's such a small component, on a screen halfway covered by the keyboard. We target older phones and I thought we should avoid keeping the CPU busy with such animations.

src/components/AccountSeed.js
```diff
@@ -17,154 +17,93 @@
...
-    }
+
     const SUGGESTIONS_COUNT = 5;
     const result = [];
     let yielded = 0;
-    result.push(WORDS[fromIndex]);
-    while (yielded < SUGGESTIONS_COUNT && WORDS[fromIndex] !== undefined) {
-      ++fromIndex;
-      if (!WORDS[fromIndex].startsWith(input)) {
+    while (yielded < SUGGESTIONS_COUNT && fromIndex < word_list.length) {
```
maciejhirsz: I'd suggest doing:

```js
const previousWords = inputWordArray.slice(0, -1); // input without last word
const suggestions = wordList.slice(fromIndex + 1, fromIndex + SUGGESTIONS_COUNT));

const result = [wordList[fromIndex]];

for (const suggestion of suggestions) {
  if (!suggestion.startsWith(input)) {
    break;
  }

  // do not suggest words that where already used
  if (previousWords.indexOf(suggestion) === -1) {
    result.push(suggestion);
  }
}

return result;
```

No magic integers tracking things.

src/components/AccountSeed.js
```diff
@@ -17,154 +17,93 @@
...
+  narrowWordList = (inputWordArray) => {
+    // if the last word is only present in BIP39 word list
+    if (BIP39_WORDS.indexOf(inputWordArray[inputWordArray.length - 2]) !== -1 && PARITY_WORDS.indexOf(inputWordArray[inputWordArray.length - 2]) === -1) {
+      this.setState({ useParityWordList: false })
+      // if the second last word is only present in Parity word list
+    } else if (BIP39_WORDS.indexOf(inputWordArray[inputWordArray.length - 2]) === -1 && PARITY_WORDS.indexOf(inputWordArray[inputWordArray.length - 2]) !== -1) {
+      this.setState({ useBIP39WordList: false })
     }
-    const suggestions = this.getSuggestions(searchInput, words);
+  }
```
maciejhirsz: Would suggest:

```js
narrowWordList = (inputWordArray) => {
  if (inputWordArray.length < 2) {
    return;
  }

  const previousWord = inputWordArray[inputWordArray.length - 2];

  // Both of these should use binary search!
  const isBIP39 = BIP39_WORDS.indexOf(previousWord) !== -1;
  const isParity = PARITY_WORDS.indexOf(previousWord) !== -1;

  if (isBIP39 && !isParity) {
    this.setState({ useParityWordList: false });
  } else if (!isBIP39 && isParity) {
    this.setState({ useBIP39WordList: false });
  }
}
```

Is there a way to revert this if I enter a word and then delete it after?

src/components/AccountSeed.js
```diff
@@ -17,154 +17,93 @@
...
+    // try to narrow down the word list using the last word typed, 
+    // as soon as a second word is being typed
+    inputWordArray.length > 1 && useBIP39WordList && useParityWordList && this.narrowWordList(inputWordArray);
+
+    let suggestions = []
+    useParityWordList && suggestions.push(...this.getSuggestions(inputWordArray, PARITY_WORDS));
+    useBIP39WordList && suggestions.push(...this.getSuggestions(inputWordArray, BIP39_WORDS));
+
+    //return a deduplicated sorted array if both word lists are still used
+    return (useBIP39WordList && useParityWordList) ? [...new Set(suggestions.sort())] : suggestions;
```
maciejhirsz:
```js
// de-duplicate results if both lists were used
if (useBIP39WordList && useParityWordList) {
  suggestions = Array.from(new Set(suggestions)).sort();
}

return suggestions;
```

More verbose, but more readable. Also, no need to sort the results before creating the set, I don't think the Set necessarily has any guarantees about insertion order (although it likely does have that behavior in V8).

src/components/AccountSeed.js
```diff
@@ -17,154 +17,93 @@
...
+  generateSuggestions = (inputWordArray) => {
+    const { useBIP39WordList, useParityWordList } = this.state;
+
+    // try to narrow down the word list using the last word typed, 
+    // as soon as a second word is being typed
+    inputWordArray.length > 1 && useBIP39WordList && useParityWordList && this.narrowWordList(inputWordArray);
+
+    let suggestions = []
+    useParityWordList && suggestions.push(...this.getSuggestions(inputWordArray, PARITY_WORDS));
+    useBIP39WordList && suggestions.push(...this.getSuggestions(inputWordArray, BIP39_WORDS));
```
maciejhirsz: `push` and spread is a total overkill here, plus I hate using logical and as shortand `if`.

```js
let suggestions = [];

if (useParityWordList) {
  suggestions = suggestions.concat(this.getSuggestions(inputWordArray, PARITY_WORDS));
}
if (useBIP39WordList) {
  suggestions = suggestions.concat(this.getSuggestions(inputWordArray, BIP39_WORDS));
}
```

If we really want a one-liner, I'd prefer `if (foo) bar;` to `foo && bar;` as it makes intent clearer, but I tend to err on having mandatory braces anyway.

src/components/AccountSeed.js
```diff
@@ -17,154 +17,93 @@
...
-        wasLetter = false;
-      }
-      ++cursor;
-    }
-    return wordPosition;
-  }
-
-  getSuggestions(input, words) {
-    let fromIndex = WORDS.findIndex(w => w.startsWith(input));
+  getSuggestions = (inputWordArray, word_list) => {
```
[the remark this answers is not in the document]
Tbaut: I'm used to arrow functions and it avoids the need to think about binding which I find nice. I can change that if needed.

src/components/AccountSeed.js
Tbaut: So much better thanks.
Just a nit, didn't test, I guess we should actually have something like:
`wordList.slice(fromIndex, fromIndex + SUGGESTIONS_COUNT));`
to start at `fromIndex` and have the right amount of `SUGGESTIONS_COUNT`.

Also because of the last `if` statement, we might end up not suggesting the amount of words we originally wanted but that probably acceptable.

src/components/AccountSeed.js
Tbaut: Thanks, again much easier to read. No way I thought about of to reset the assumption. We could add some state reset if the `inputWords <= 2 &&  (this.state.useBIP39WordList || this.state.useBIP39WordList)`

src/components/AccountSeed.js
Tbaut: Just for my information, is what you suggest performing better, or is that just preferences? I prefer `&&` compared to `if` but I can change my preferences :)

src/components/AccountSeed.js
maciejhirsz: It should go from `fromIndex + 1` because the first one is already in the result, but now that I'm reading it, that won't check if it hasn't been used in the phrase before, will adjust...

src/components/AccountSeed.js
maciejhirsz: I don't think it performs better, it's just less magical and more obvious to whoever is going to read it later what it is you wanted to do. Replacing `if` with `&&` is the sort of a thing `UglifyJS` does to reduce file size :P.

src/components/AccountSeed.js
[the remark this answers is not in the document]
maciejhirsz: Well, this thing doesn't use any state at all, it's a pure function, could be a static method or extracted out of the component altogether.

maciejhirsz:
I've reworked a bunch of stuff here and added binary search. Would be cool if you can look through the changes and see if there is anything weird or not understandable @Tbaut.

One thing that should be a pretty low hanging fruit now is making sure the background turns green for BIP39 phrases only if they are actually valid phrases (with checksum and all that, given that we can know the selected dictionary now).

src/components/AccountSeed.js
maciejhirsz: I've reintroduced this feature in last commit here, logic is streamlined so it should make everyone happy :).

PR #239, opened 2019-05-30
Show seed type and green field if valid bip39

Tbaut:
- add validBip39Seed to new account's state in `AccountStore`
- only mark the recovery field as valid if the bip39 checksum passes
- show in the account card the seed type
- warn users if they recover a brain wallet (e.g no bip39 type)

Here is how it looks: (valid bip39, invalid bip39, warning)
image
image
image

### DIFF
--- a/src/components/AccountCard.js
+++ b/src/components/AccountCard.js
@@ -28,7 +28,7 @@ export default class AccountCard extends React.PureComponent<{
   address: string,
   chainId: string,
   onPress: () => any,
-  title: ?string,
+  title?: string,
   seedType?: string,
 }> {
   static propTypes = {
@@ -50,7 +50,7 @@ export default class AccountCard extends React.PureComponent<{
     title = title.length ? title : AccountCard.defaultProps.title;
     let seedTypeDisplay = ''
 
-    if (seedType && seedType!== ''){
+    if (seedType && seedType !== ''){
       seedTypeDisplay = 'seed: '+ seedType;
     }
 
--- a/src/components/AccountCard.js
+++ b/src/components/AccountCard.js
@@ -48,11 +48,7 @@ export default class AccountCard extends React.PureComponent<{
     const { address, chainId, onPress, seedType, style } = this.props;
     let { title } = this.props;
     title = title.length ? title : AccountCard.defaultProps.title;
-    let seedTypeDisplay = ''
-
-    if (seedType && seedType !== ''){
-      seedTypeDisplay = 'seed: '+ seedType;
-    }
+    const seedTypeDisplay = seedType || '';
 
     return (
       <TouchableItem
--- a/src/components/AccountCard.js
+++ b/src/components/AccountCard.js
@@ -25,17 +25,19 @@ import AccountIcon from './AccountIcon';
 import TouchableItem from './TouchableItem';
 
 export default class AccountCard extends React.PureComponent<{
-  title: ?string,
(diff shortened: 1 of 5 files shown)


src/components/AccountCard.js
```diff
@@ -25,27 +25,34 @@ import AccountIcon from './AccountIcon';
 import TouchableItem from './TouchableItem';
 
 export default class AccountCard extends React.PureComponent<{
-  title: ?string,
   address: string,
   chainId: string,
-  onPress: () => any
+  onPress: () => any,
+  title: ?string,
```
yjkimjunior: `title?:`

src/components/AccountCard.js
```diff
@@ -25,27 +25,34 @@ import AccountIcon from './AccountIcon';
...
   };
 
   render() {
-    const { address, chainId, style, onPress } = this.props;
+    const { address, chainId, onPress, seedType, style } = this.props;
     let { title } = this.props;
     title = title.length ? title : AccountCard.defaultProps.title;
+    let seedTypeDisplay = ''
+
+    if (seedType && seedType!== ''){
```
yjkimjunior: `seedType !== ''`

src/util/account.js
```diff
@@ -52,6 +46,13 @@ export function validateSeed (seed) {
     }
   }
 
+  if (!validBip39Seed) {
+    return {
+      valid: false,
+      reason: `This recovery phrase will be treated as a "legacy Parity brain wallet"`
```
yjkimjunior: Some wording suggestions (but please take with grain of salt):

 - Parity brain wallets are no longer supported.
 - The recovery phrase you entered is not a valid BIP39 phrase.
 - The recovery phrase you entered is a legacy Parity brain wallet seed, and is no longer supported.

src/components/AccountCard.js
Tbaut: if `seedType` is undefined `seedType!== ''` will be true. Since `seedType` is an optional `prop` I should check that it's defined isn't it?

src/components/AccountCard.js
Tbaut: I was just moving the props around for alphabetical order, I wonder if this is understood correctly. Will change it.

src/components/AccountCard.js
```diff
@@ -25,27 +25,34 @@ import AccountIcon from './AccountIcon';
...
   render() {
-    const { address, chainId, style, onPress } = this.props;
+    const { address, chainId, onPress, seedType, style } = this.props;
     let { title } = this.props;
     title = title.length ? title : AccountCard.defaultProps.title;
+    let seedTypeDisplay = ''
+
+    if (seedType && seedType !== ''){
+      seedTypeDisplay = 'seed: '+ seedType;
+    }
```
maciejhirsz: TBH I'd rather drop the whole `seed: `, it looks super noisy to me. Having simple visual indicator that says "bip39" or "brain wallet" should be fine.

That said:

```
const seedTypeDisplay = seedType || '';
```

src/util/account.js
maciejhirsz: https://www.reddit.com/r/UnnecessaryQuotes/ :P

src/components/AccountCard.js
Tbaut: Sure, non tech users will be lost with this jargon anyways

PR #241, opened 2019-06-08
Update js-yaml and handlbars

Tbaut:
Self explanatory.
I've tested successfully account recovery and transfer on android.


maciejhirsz:
We shouldn't be adding dependencies that we don't directly use, that's weird. If there is a need to upgrade dependencies-of-dependencies, it should be done in the lock file, or by bumping the version of whatever is pulling them in. Both of these are only used by `jest`, bumping `jest` to latest version `24.8.0` bumps `handlebars` up and removes `js-yaml`.

Tbaut:
thanks, that was a dumb one.

PR #227, opened 2019-02-26
Menu refactoring using popup

Tbaut:
- closes #225

screenshots:

PR #243, opened 2019-06-11
Bump jest to 24.8.0

Tbaut:
Self explanatory, and more sensible this time :)
Tested successfully the build and functionnalities on android.


Tbaut:
Do you mean I should remove the `yarn.lock`? I've had the package-lock.json in [previous PRs](https://github.com/paritytech/parity-signer/pull/227/files)`

yjkimjunior:
As long as it's consistent it should be no problem. Dug a little into past PR's where just the `yarn.lock` is committed. e.g. https://github.com/paritytech/parity-signer/pull/232/files

maciejhirsz:
I don't have any strong feelings for or against yarn/npm, so adding whichever lock file on .gitignore and then using the other is fine with me. `npm` might be a bit easier since we can then remove `yarn` from required dependencies for the setup, but I reckon it isn't really a blocker for anyone.

Tbaut:
I went for removing the yarn.lock and adding it to .gitignore,

yjkimjunior:
ah I just realized the `setup_macos.sh` script actually runs yarn at the :/

I will change this to `npm install` in a separate PR

fckt:
> I went for removing the yarn.lock and adding it to .gitignore,

Consider to read this:
https://yarnpkg.com/blog/2018/06/04/yarn-import-package-lock/

>> We highly recommend you to delete the package-lock.json file if you decide to use yarn in order to avoid future confusion and possible consistency issues.

PR #244, opened 2019-06-14
Fix CI builds

ddorgan:
sudo not required


Tbaut:
I've signed and tested the unsiged `apk` that was generated, but due to the problem we have since https://github.com/paritytech/parity-signer/pull/200#issuecomment-468723460 I couldn't generate a working apk :(. Force closing at launch.

Issue #261, opened 2019-07-19
Existing recovery account recovery on iPhone Xr


jacogr:
Ok, so latest toy, latest Signer.

Steps -

1. Empty accounts screen went to "recover account"

2. Spent the next 5 mins typing in my 34-word Parity-style recovery phrase

3. While entering this, around half-way the words started being behind the keyboard, so I couldn't actually see what I was typing. (Hence spending 5mins, I'm a slow phone typer and needed to check that auto-correct didn't do anything stupid with each word)

4. Once done, I couldn't move out of the box - the keyboard covers half my phrase and the remainder of the screen. No way to close the keyboard to press the button.

5. So... I just went out.

6. Then decided, maybe I will try again.

7. Clicked recover.

8. The recovery screen opened... and well, there is a bug (but it helped me recover from the first issue) - the old phrase I spent 5 mins typing in was still visible. Great! No keyboard, so could press next and add the PIN.

TL;DR Seems to be 2 issues, but the second one helped me recover from the first.

jacogr:
Actually don't even need to type in the whole phrase, just starting is enough to let the keyboard get in the way and overlap the box.

![image](https://user-images.githubusercontent.com/1424473/61526032-e8506d00-aa19-11e9-8118-10b44bf51fd7.png)

So from there, just have no way to minimize the keyboard to move on (Tried "Return")

PR #265, opened 2019-07-20
Make sure the recovery phrase is emptied upon process abortion

Tbaut:
- closes https://github.com/paritytech/parity-signer/issues/261(the second bug mentioned)

Users could set a seed for the `new` account in `AccountRecover`. This seed remained in the `AccountStore` state upon unmount. It needed to be manually removed in case a user doesn't finish the process.

### DIFF
--- a/src/screens/AccountRecover.js
+++ b/src/screens/AccountRecover.js
@@ -17,7 +17,7 @@
 'use strict';
 
 import React from 'react';
-import { Alert, findNodeHandle, KeyboardAvoidingView, StyleSheet, ScrollView, Text, View } from 'react-native';
+import { Alert, KeyboardAvoidingView, StyleSheet, ScrollView, Text } from 'react-native';
 import { Subscribe } from 'unstated';
 import colors from '../colors';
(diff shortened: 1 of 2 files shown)


Tbaut:
Please give it a shot, it's a small one with big impact. You could end up with an empty account in the account list if you really wished.

hanwencheng:
The PR LGTM, but one more thing about the workflow of recovering an account, since create a empty account will result in nothing in the storage, I think we should prevent the user enter the pin creation page with empty seed. also according to the [previous discussion](https://github.com/paritytech/parity-signer/issues/127#issuecomment-454610060), related code [here](https://github.com/paritytech/parity-signer/blob/master/src/util/account.js#L32-L47), basically these situation need to be prevented from further steps, probably we should create a new issue to solve this problem.

Tbaut:
You're right, I'll take care of this in this PR, it shouldn't be a big deal. I'll separate the other small bugfix into its own PR.

Tbaut:
On this PR is now as it should be, fixing one problem.

Issue #252, opened 2019-07-17
Account refactor to support multi-protocol


Tbaut:
We need to refactor the AccountsStore and accounts in general to support Substrate. The mapping looking like `account` -> `networkIndex`, the array of network should look similar to:

```js
[
  {
    name: 'kovan',
    color: '#rtzui',
    protocol: 'ethereum',
    ethereumChainId: 2
  },
  {
    name: 'kusama',
    color: '#34567',
    protocol: 'substrate',
    ss58Prefix: 2,
    balanceModuleId: 123
  }
]
```

We should be able to use the same AES encryption.

PR #275, opened 2019-07-22
fix: refactor the network list and add lodash

hanwencheng:
#252 
* refactor constants.js.
* add Substrate endpoint in constants.js
* add lodash as a dependency.
* create NetworkButton Component.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "bignumber.js": "^4.0.0",
     "debounce": "^1.0.0",
     "hoist-non-react-statics": "^2.5.0",
+    "lodash": "^4.17.15",
     "prop-types": "^15.6.1",
     "react": "16.8.6",
     "react-native": "0.60.3",
--- a/src/components/AccountCard.js
+++ b/src/components/AccountCard.js
@@ -20,7 +20,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import { StyleSheet, Text, View, ViewPropTypes } from 'react-native';
 import colors from '../colors';
-import { DEFAULT_NETWORK_COLOR, NETWORK_COLOR, NETWORK_TITLES } from '../constants';
+import { NETWORK_LIST } from '../constants';
 import AccountIcon from './AccountIcon';
 import TouchableItem from './TouchableItem';
 
@@ -29,7 +29,7 @@ export default class AccountCard extends React.PureComponent<{
   chainId: string,
   onPress: () => any,
   title?: string,
-  seedType?: string,
+  seedType?: string
 }> {
   static propTypes = {
     address: PropTypes.string.isRequired,
@@ -37,7 +37,7 @@ export default class AccountCard extends React.PureComponent<{
     onPress: PropTypes.func,
     seedType: PropTypes.string,
     style: ViewPropTypes.style,
-    title: PropTypes.string,
+    title: PropTypes.string
   };
 
   static defaultProps = {
@@ -49,6 +49,7 @@ export default class AccountCard extends React.PureComponent<{
     let { title } = this.props;
     title = title.length ? title : AccountCard.defaultProps.title;
     const seedTypeDisplay = seedType || '';
+    const network = NETWORK_LIST[chainId];
 
     return (
       <TouchableItem
@@ -77,7 +78,7 @@ export default class AccountCard extends React.PureComponent<{
             style={[
               styles.footer,
               {
-                backgroundColor: NETWORK_COLOR[chainId] || DEFAULT_NETWORK_COLOR
+                backgroundColor: network.color
               }
             ]}
           >
@@ -85,9 +86,7 @@ export default class AccountCard extends React.PureComponent<{
               style={[
                 styles.footerSeedType,
                 {
-                  color: NETWORK_COLOR[chainId]
-                    ? colors.card_bg
-                    : colors.card_text
+                  color: network.secondaryColor
                 }
               ]}
             >
@@ -97,13 +96,11 @@ export default class AccountCard extends React.PureComponent<{
               style={[
                 styles.footerNetwork,
                 {
-                  color: NETWORK_COLOR[chainId]
-                    ? colors.card_bg
-                    : colors.card_text
+                  color: network.secondaryColor
                 }
               ]}
             >
-              {NETWORK_TITLES[chainId]}
+              {network.title}
             </Text>
           </View>
         </View>
--- a/src/components/AccountDetailsCard.js
+++ b/src/components/AccountDetailsCard.js
@@ -20,7 +20,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import { StyleSheet, Text, View } from 'react-native';
 import colors from '../colors';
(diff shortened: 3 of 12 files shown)


hanwencheng:
[clabot:check]

src/components/AccountDetailsCard.js
```diff
@@ -20,7 +20,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import { StyleSheet, Text, View } from 'react-native';
 import colors from '../colors';
-import { DEFAULT_NETWORK_COLOR, NETWORK_COLOR, NETWORK_TITLES } from '../constants';
+import { NETWORK_LIST } from '../constants';
 import AccountIcon from './AccountIcon';
 import TouchableItem from './TouchableItem';
```
Tbaut: This class isn't used anywhere, we can remove the whole file I guess.

package.json
```diff
@@ -23,6 +23,7 @@
     "bignumber.js": "^4.0.0",
     "debounce": "^1.0.0",
     "hoist-non-react-statics": "^2.5.0",
+    "lodash": "^4.17.15",
```
yjkimjunior: +1 for removing lodash.

src/constants.js
```diff
@@ -1,40 +1,99 @@
...
+  [EthereumNetworkIds.CLASSIC]: {
+    title: 'Ethereum Classic',
+    color: '#8C7166',
+    secondaryColor: colors.card_bg,
+    available: true
+  },
+  [EthereumNetworkIds.EXPANSE]: {
+    title: 'Expanse'
+  },
+  [EthereumNetworkIds.ROPSTEN]: {
```
yjkimjunior: we should probably give the testnets their own color too? I'm not too fussed about this though

src/constants.js
```diff
@@ -1,40 +1,99 @@
...
-  [NETWORK_ID.frontier]: 'Ethereum',
-  [NETWORK_ID.classic]: 'Ethereum Classic',
-  [NETWORK_ID.ropsten]: 'Ropsten Testnet',
-  [NETWORK_ID.kovan]: 'Kovan Testnet',
-  [NETWORK_ID.goerli]: 'Görli Testnet'
+const substrateNetworkRaw = {
+  [SubstrateNetworkIds.KUSAMA]: {
+    name: 'kusama',
+    ss58Prefix: 2,
+    balanceModuleId: 123
```
yjkimjunior: question: is this the actual balanceModuleId for Kusama? if not, could you please leave a comment in case someone else is trying to figure that out in the future :)

src/constants.js
```diff
@@ -1,40 +1,99 @@
...
+export const ETHEREUM_NETWORK_LIST = _.mapValues(
+  ethereumNetworkRaw,
+  (ethereumNetwork, networkId) =>
+    _.defaults(ethereumNetwork, {
+      protocol: NetworkTypes.ETHEREUM,
+      color: '#F2E265',
+      secondaryColor: colors.card_text,
+      available: false,
+      title: `Ethereum_${networkId}`,
+      ethereumChainId: networkId
```
yjkimjunior: could you please also rename `chainId` to `ethereumChainId` everywhere where it's appropriate?

src/components/AccountCard.js
```diff
@@ -49,6 +49,7 @@ export default class AccountCard extends React.PureComponent<{
     let { title } = this.props;
     title = title.length ? title : AccountCard.defaultProps.title;
     const seedTypeDisplay = seedType || '';
+    const network = NETWORK_LIST[chainId];
```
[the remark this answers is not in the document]
hanwencheng: good point also as @yjkimjunior mentioned

src/constants.js
hanwencheng: No, I do not know what the balanceModuleId is, I will leave the comment

src/constants.js
hanwencheng: also as @Tbaut mentioned

src/components/AccountDetailsCard.js
hanwencheng: True, probably we should set up the `lint` and `prettier` in `pre-commit`

src/constants.js
hanwencheng: Screenshot 2019-07-23 at 11 18 16

How do you like this one? what about in parity Ethereum wallet, what the color schema is used there?

src/components/AccountCard.js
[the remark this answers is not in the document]
maciejhirsz: Since this is indexing into a mapping, I'd suggest `networkKey` instead.

src/constants.js
maciejhirsz: I think yellow for all testnets is okay. There are only so many easily distinguishable hues, and UX-wise we should mostly care about differentiating between different value-bearing networks. I would give Kusama a different color though (or a different default for all Substrate networks in general).

hanwencheng:
Also one comment about the `lodash` as a dependency. There are several reasons to use it.
1. The explicit function name, I agree with @Tbaut, functions like `mapValues` or `defaults` could actually all be written with vanilla `reduce` or `assign` function. and most of the collection operation functions could be easily written with `reduce`.
But compare the following code with vanilla js, with the lodash function the code gains its readability, which could be understood at the first sight, without split the code to create new function name (e.g. in the following example to create a `createNetworkWithDefaults` with `Object.assign({}, defaultProps, substrateNetwork )` for readability):
Vanilla:
```javascript
const defautProps = {
  protocol: NetworkTypes.SUBSTRATE,
  color: '#E6007A',
  secondaryColor: colors.card_bg,
  available: false
};

const SUBSTRATE_NETWORK_LIST = substrateNetworkRaw.reduce((acc, substrateNetwork, networkId)=>
 Object.assign(acc, {networkId:  { networkId: Object.assign({}, defaultProps, substrateNetwork )}})
, {});
```
With Lodash:
```javascript
const SUBSTRATE_NETWORK_LIST = _.mapValues(
  substrateNetworkRaw,
  (substrateNetwork, networkId) =>
    _.defaults(substrateNetwork, {
      protocol: NetworkTypes.SUBSTRATE,
      color: '#E6007A',
      secondaryColor: colors.card_bg,
      available: false
    })
);
```

2. the composer function, although it is not used here, it could also improve the readability, and to make the code clean. `lodash/fp` could make it cleaner if using functional programming styles.

3. some functions are not identical in Vanilla and Lodash, for example, the `map` function only works for Array in Vanilla, but intuitively we suppose it could also work for an object.

I do think readability is important and that's why I introduced lodash here, but I agree with @Tbaut that we should keep the code consistent, without using the similar function like `_.assign` or `Object.assign` or  `Array.map` and `_.map` in the same file.

yjkimjunior:
fair enough, I was more concerned about how much it'd bloat the bundle size. i think cherrypicking lodash module imports helps, e.g. `import map from lodash/map`

hanwencheng:
The `chainId` renaming requires to change a lot of files, in case of the conflict with other PRs and make it easy for all of us to resolve, I will submit another PR focus on that.

src/components/AccountDetailsCard.js
hanwencheng: This one is still in use.

PR #279, opened 2019-07-23
Substrate stuff in Rust

### LINKED ISSUE #155: Add ed25519 and sr25519 support for Substrate
- Expose rust crypto for sr25519 createAddressFromSeedUri (supporting SURI aka derivation path just like in rust and JS) to create addresses.
- Expose a new hash function to hash Tx data.
- Expose 2 substrateSign functions to sign data blobs (with ed25519 and sr25519).
- Expose a new qrCode function to be able to generate QR code with binary data (UOS support).

maciejhirsz:
Closes #155

+ [x] Expose a new qrCode function to be able to generate QR code with binary data (UOS support).
+ [x] Expose a new blake2s hash function to hash substrate payloads.
+ [x] Expose rust crypto for sr25519 createAddressFromSeedUri (supporting SURI aka derivation path just like in rust and JS) to create addresses.
+ [x] ~Expose 2 substrateSign functions to sign data blobs (with ed25519 and sr25519)~. Expose sr25519 signing (we can add ed when/if we need it).


ios/NativeSigner/EthkeyBridge.swift
```diff
@@ -65,6 +65,17 @@ class EthkeyBridge: NSObject {
     resolve(hash)
   }
 
+  @objc func blake2s(_ data: String, resolve: RCTPromiseResolveBlock, reject: RCTPromiseRejectBlock) -> Void {
+    var error: UInt32 = 0
+    var data_ptr = data.asPtr()
+    let hash_rust_str = blake2s(&error, &data_ptr)
```
yjkimjunior: @maciejhirsz maybe a dumb question, not sure if this is a bug on my end but when trying to compile this atm I'm getting an error: `Use of 'blake2s' nearly matches global function 'blake2s' in module '__ObjC' rather than instance method 'blake2s(_:resolve:reject:)'`?

maciejhirsz: Most likely my end since I haven't tried to compile it, I'll rename stuff.

maciejhirsz:
Sr25519 signing is done. I need to look into signing with Ed25519.

On SURI support: I'm currently only accepting phrases, but it should be fairly straightforward to add SURI support - basically you just append the phrase string with extra `/soft` or `//hard` derivation segments, and optionally `///password` at the end (I need to read up how password is actually used in substrate code, and whether it's something that's applicable here at all). Will have to think on how to add support for that in the UI.

Another thing is the signing of extrinsic, and extracting data from them. Right now it just takes a byte array to sign, and as far as I understand you sign either the blake2 hash of the extrinsic (hashing function exposed) or if the extrinsic is fewer bytes than 32 bytes. We might have to handle this in JS, but I'll continue reading on how all of this works in substrate code, so we don't make any dumb mistakes.

For now, I'd be happy if someone can check if this compiles for iOS and if the new functions are callable and produce non-error results on both platforms. CC @hanwencheng @yjkimjunior @Tbaut. [Some test vectors for the address](https://github.com/paritytech/parity-signer/blob/5276807809e4de0f8d6cfae41d191329586461ad/rust/signer/src/lib.rs#L248-L257).

yjkimjunior:
>  Right now it just takes a byte array to sign, and as far as I understand you sign either the blake2 hash of the extrinsic (hashing function provided in this PR) or if the extrinsic is fewer bytes than 32 bytes. 
I was not aware of there being a special way of handling the case where extrinsic is fewer than 32 bytes? FWICT, https://github.com/polkadot-js/api/blob/master/packages/types/src/primitive/Extrinsic/util.ts

Also a small aside, `Ethkeybridge.blake2` was not exposed in `native.js` in this branch, though I was able to do so and use it in my WIP PR.

Tbaut:
It didn't compile for me for Android:

I get the following when running `cd rust/signer && make android`
```
   Compiling rlp v0.3.0
   Compiling signer v0.1.0 (/home/thib/github/paritytech/parity-signer/rust/signer)
error: linker `/home/thib/paritytech/parity-signer/NDK/arm64/bin/aarch64-linux-android-gcc` not found
  |
  = note: No such file or directory (os error 2)
error: aborting due to previous error
error: Could not compile `signer`.
```

Would it be possible to return a boolean whether or not the passed seed is a valid bip39, just like we do for `brainWalletAddress`?

maciejhirsz:
> I was not aware of there being a special way of handling the case where extrinsic is fewer than 32 bytes?

You are right, it's 256 bytes.

> Also a small aside, `Ethkeybridge.blake2` was not exposed in `native.js` in this branch, though I was able to do so and use it in my WIP PR.

Added it now!

> Would it be possible to return a boolean whether or not the passed seed is a valid bip39, just like we do for `brainWalletAddress`?

Substrate supports bip39 phrases **only**, so there is no need for a boolean.

Tbaut:
> Substrate supports bip39 phrases only, so there is no need for a boolean.

but but.. I am able to create something with a random see using the UI. That's maybe a trick.

yjkimjunior:
I am able to `yarn run build-rust-ios` and android, the rust stuff passes `cargo test`, but trying to use `substrateAddrress` for example, I get `_NativeModules.EthkeyBridge.substrateAddress is not a function`.

hanwencheng:
It compiled successfully on both iOS and iPhone, probably I do not need to test it on both as they are rust binding?

yjkimjunior:
okay, as discussed and confirmed with Hanwen just now, `EthkeyBridge.substrateAddress` is not callable from js, while the native function itself behaves as expected i.e. there is something wonky with bridging.

yjkimjunior:
@maciejhirsz EthkeyBridge.m requires a couple exports
```
RCT_EXTERN_METHOD(substrateAddress:(NSString*)seed prefix:(NSString*)prefix resolve:(RCTPromiseResolveBlock)resolve reject:(RCTPromiseRejectBlock)reject)
RCT_EXTERN_METHOD(substrateSign:(NSString*)seed message:(NSString*)message resolve:(RCTPromiseResolveBlock)resolve reject:(RCTPromiseRejectBlock)reject)
```

maciejhirsz:
@yjkimjunior thanks, added in.

burdges:
This is miss-using the chaincode in the same way that subkey miss-uses it.  hmm

burdges:
We sadly cannot fix it for kusama however, so..

Issue #258, opened 2019-07-18
Can't see the selected identicon at account creation


Tbaut:
We used to see clearly which account was selected:
![image](https://user-images.githubusercontent.com/33178835/61480961-5b64cf80-a997-11e9-9bf7-15269827b560.png)

now you can't see anymore:
![image](https://user-images.githubusercontent.com/33178835/61481042-78999e00-a997-11e9-872b-70a1be32f6b8.png)

PR #280, opened 2019-07-23
Refactor new account state management

Tbaut:
- closes https://github.com/paritytech/parity-signer/issues/258

While digging into this bug, I realized that the way we managed new account info in the `AccountNew` screen was a little heavy. Calling `updateNew` every time anything changes is demanding for the device. In master when you click quickly between identicons, the device is lagging so much that weird things appear (this is introduced by persisting each change). I chose to store the new account first in a local state and to submit it to the account store only when tapping the *next step* button.

This required a very small change in `NetworkChooser` component, it doesn't access the AccountStore anymore.

What it fixes:

I do exactly the same in each example, I click on new account and click on the 4 first icons **consecutively**. You can see the selection is jumping.
Before :-1: 
image

After :+1: 
image

### DIFF
--- a/src/screens/AccountNetworkChooser.js
+++ b/src/screens/AccountNetworkChooser.js
@@ -18,11 +18,9 @@
 
 import React from 'react';
 import { ScrollView, StyleSheet, Text } from 'react-native';
-import { Subscribe } from 'unstated';
 import colors from '../colors';
 import TouchableItem from '../components/TouchableItem';
 import { DEFAULT_NETWORK_COLOR, NETWORK_COLOR, NETWORK_LIST, NETWORK_TITLES } from '../constants';
-import AccountsStore from '../stores/AccountsStore';
 
 export default class AccountNetworkChooser extends React.PureComponent {
   static navigationOptions = {
@@ -31,20 +29,7 @@ export default class AccountNetworkChooser extends React.PureComponent {
   };
 
   render() {
-    return (
-      <Subscribe to={[AccountsStore]}>
-        {accounts => (
-          <AccountNetworkChooserView {...this.props} accounts={accounts} />
-        )}
-      </Subscribe>
-    );
-  }
-}
-
-class AccountNetworkChooserView extends React.PureComponent {
-  render() {
     const { navigation } = this.props;
-    const { accounts } = this.props;
     return (
       <ScrollView style={styles.body} contentContainerStyle={{ padding: 20 }}>
         <Text style={styles.title}>CHOOSE NETWORK</Text>
@@ -59,7 +44,7 @@ class AccountNetworkChooserView extends React.PureComponent {
               }
             ]}
             onPress={() => {
-              accounts.updateNew({ chainId });
+              navigation.getParam('onChange')(chainId);
               navigation.goBack();
             }}
           >
--- a/src/screens/AccountNew.js
+++ b/src/screens/AccountNew.js
@@ -29,7 +29,6 @@ import TextInput from '../components/TextInput';
 import TouchableItem from '../components/TouchableItem';
 import { DEFAULT_NETWORK_COLOR, NETWORK_COLOR, NETWORK_TITLES } from '../constants';
 import AccountsStore from '../stores/AccountsStore';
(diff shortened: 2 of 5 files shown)


src/screens/AccountNew.js
```diff
@@ -92,13 +100,13 @@ class AccountNewView extends React.Component {
               <Button
                 buttonStyles={styles.nextStep}
                 title="Next Step"
-                disabled={ !validateSeed(selected.seed, selected.validBip39Seed).valid }
```
Tbaut: I don't think it makes sense here to validate the seed since it's not a use input, unlike for the recovery screen.

src/screens/AccountNetworkChooser.js
```diff
@@ -63,7 +48,7 @@ class AccountNetworkChooserView extends React.PureComponent {
               }
             ]}
             onPress={() => {
-              accounts.updateNew({ chainId: network.ethereumChainId });
```
hanwencheng: I am not sure if this part of refactoring does not affect renderings. I generally do not think to pass a function as params in react-navigation is not a good idea, because this will involves `this` pitfall. Could you double-check that, since I think the performance issue is mainly caused by the icon selection part? Another thing is that if we need this part, could we change the function name `onChange` to something like `updateNetwork`, because the name is implicit?

src/screens/AccountNew.js
```diff
@@ -47,13 +46,21 @@ export default class AccountNew extends React.Component {
...
+    super(props);
+
+    this.state = { ...this.props.accounts.getNew() };
+  }
+
   render() {
-    const { accounts } = this.props;
-    const selected = accounts.getNew();
+    const { accounts, navigation } = this.props;
+    const selected = this.state
```
hanwencheng: a trailing comma here 🙂

hanwencheng:
Here is how is shows on my simulator on both ios and android (on `master` branch), I can rarely see the lagging, so I leave some comments, hope they are still useful.
out

src/screens/AccountNetworkChooser.js
Tbaut: the `this` binding problem is handled nicely by the fat arrow (`=>`) function. It works pretty well as far as I could test.

The full story about the performance issue is the following:

You change the `selected.seed` and `selected.address` when selecting an icon. This triggers an `updateNew` that does a bunch of things, one of it being and `await brainWalletAddress(seed);`  (that we actually don't need) and this inside a `debounce( .., 200)`. 

Once this is done, this triggers a re-rendering of the view because the accounts (props) from the store have changed, where `selected = accounts.getNew()` is actualized. Then because `selected,address` is the `value` of `AccountIconChooser` this triggers a re-rendering of it with what was returned by the update. Long story short, if you log `AccountIconChooser` when the icon is selected and what `address` (called value) it receives from the props you'll see. if you click on the 3 or 4 first icons:
```js
selected 1;
selected 2;
selected 3;
get 1;
get 2;
get 3;
```

> Another thing is that if we need this part, could we change the function name onChange to something like updateNetwork, because the name is implicit?

Sure why not

Tbaut:
click on 3 or 4 icons consecutively just like my example, you'll see the lag for sure. Clicking on 1, 2, 1 doesn't allow to see it clearly.

src/screens/AccountNetworkChooser.js
Tbaut: Although I agree that passing an onChange function to the navigation params is probably not the best practice, I also think that this component should probably be refactored into a modal or a simple dropdown rather than its own navigation screen with access to the `unstated` accounts.

Tbaut:
So a better re-factoring might be to do it on the `updateNew` function?

hanwencheng:
I change the debounce delay to 500, and now I clearly see the problem, I think it is a expected behavior of [debouce](https://github.com/component/debounce#readme).
We may add `true` to the immediate flag, tried on the simulator, that could solve the problem.

Tbaut:
ok I'll close this and rather change the `updateNew` then to bypass this seed check for new accounts.

Issue #203, opened 2019-01-16
Account with empty recovery can't sign or get deleted


Tbaut:
For some reason, the address that gets generated with an empty seed can't sign or be deleted (0x00a329c0648769a73afac7f9381e08fb43dbea72). it simply shows the error "wrong pin". Even though this bug should not happen after https://github.com/paritytech/parity-signer/issues/154 users that have generated it should be able to delete it.

PR #282, opened 2019-07-24
Don't allow empty seed phrase recovery

Tbaut:
- closes https://github.com/paritytech/parity-signer/issues/203
- alphabetical order cleanup

An empty account never got persisted in storage (since https://github.com/paritytech/parity-signer/pull/235/files) however it was still added to the `AccountStore` state and ended up being listed on the account overview. This doesn't happen anymore.

We now also don't allow users to recover an account with an empty seed.

### DIFF
--- a/src/screens/AccountRecover.js
+++ b/src/screens/AccountRecover.js
@@ -105,7 +105,7 @@ class AccountRecoverView extends React.Component {
                 const validation = validateSeed(selected.seed, selected.validBip39Seed);
                 if (!validation.valid) {
                   if (validation.accountRecoveryAllowed){
-                    Alert.alert(
+                    return Alert.alert(
                       'Warning:',
                       `${validation.reason}`,
                       [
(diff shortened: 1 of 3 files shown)


src/screens/AccountRecover.js
```diff
@@ -104,29 +104,43 @@ class AccountRecoverView extends React.Component {
...
+                      'Error:',
+                      `${validation.reason}`,
+                      [
+                        {
+                          text: 'Back',
+                          style: 'cancel'
+                        }
+                      ]
+                    );
+                    return;
```
hanwencheng: could be written in the same line with Alert.alert.

hanwencheng:
Looks good, tested on iOS.

src/screens/AccountRecover.js
```diff
@@ -104,29 +104,43 @@ class AccountRecoverView extends React.Component {
...
-                        style: 'default',
-                        onPress: () => {
-                          this.props.navigation.navigate('AccountPin', {
-                            isWelcome: this.props.navigation.getParam(
-                              'isWelcome'
-                            ),
-                            isNew: true
-                          });
+                  if (validation.accountRecoveryAllowed){
+                    Alert.alert(
```
hanwencheng: and here

PR #283, opened 2019-07-24
Get back the icon selector

### LINKED ISSUE #258: Can't see the selected identicon at account creation

Tbaut:
- closes #258
- remove stray console.log
- move `renderItem` into its own function to clean the `render`
- remove unused variables


src/components/AccountIconChooser.js
```diff
@@ -73,38 +73,40 @@ export default class AccountIconChooser extends React.PureComponent<{
     }
   };
 
-  render() {
+  renderIcon = ({ item, index }) => {
```
hanwencheng: Nice, I think it will contribute to rendering optimization.

PR #284, opened 2019-07-25
Refactor updateNewAccount

Tbaut:
- debounce is now debouncing.
- icon selection isn't lagging anymore.
- removed a direct dependency to `debounce`

What it effectively solves:

On account recovery, debounce is working, the device used to calculate the address at every key stroke. Now it's effectively debouncing (after/before):
debounce

On new account, the icon selection doesn't re-render.
I do exactly the same in each example, I click on new account and click on the 4 first icons **consecutively**. You can see the selection is jumping. (before/after)
image

### DIFF
--- a/src/screens/AccountNew.js
+++ b/src/screens/AccountNew.js
@@ -74,7 +74,7 @@ class AccountNewView extends React.Component {
               <AccountIconChooser
                 value={selected && selected.seed && selected.address}
                 onChange={({ address, seed, bip39 }) => {
-                  accounts.updateNew({ address, seed, validBip39Seed: bip39 }, false);
+                  accounts.updateNew({ address, seed, validBip39Seed: bip39 });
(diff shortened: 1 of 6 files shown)


hanwencheng:
LGTM

src/screens/AccountNew.js
```diff
@@ -74,7 +74,7 @@ class AccountNewView extends React.Component {
               <AccountIconChooser
                 value={selected && selected.seed && selected.address}
                 onChange={({ address, seed, bip39 }) => {
```
hanwencheng: Could you please also rename this function here? I forget in which PR we discussed before.

yjkimjunior:
@Tbaut tested on iphone 6, works as expected!

yjkimjunior:
ezgif-1-9483c7328b65

Issue #266, opened 2019-07-22
Shows "Secure" when connected to cellular internet


phillux:
v2.2.1 / iOS / iPhone X

I'm connected to cellular internet and it shows "Secure" in the upper right hand corner. When wifi is connected, it shows "Not secure"

Tbaut:
Related to https://github.com/paritytech/parity-signer/issues/174
We're using a dep ([rn comminuty netInfo](https://github.com/react-native-community/react-native-netinfo)) for this feature that seems unreliatble and will not be maintained long term. IIRC it asks for wifi permissions (at least on android), which is a cause of friction.

I'm for removing this non essential feature altogether. What do you think? (also ping for opinions @maciejhirsz @yjkimjunior  @hanwencheng )

hanwencheng:
I think it would be helpful if we can automatically turn off all the internet connection first when we receive a signing request (with the `SCAN` button), and after sending the signing information the internet connection would start again, because the secure sign makes the most sense when the user starts to receive the request.
But it needs some research to see if it could be done easily and make it behave the same with both Android and iOS, some promising library may have done that like [react offline](https://github.com/rgommezz/react-native-offline).

Tbaut:
Parity Signer is meant to be used offline from the start, and it should remain so. We will not implement any feature to force users to go offline because this is was users must do in order to keep their funds safe. The phone should **never** be online.

Although we keep repeating this, we thought originally that it could be a nice to have, to show if users have forgotten to turn off any connectivity feature on their phone. The reality is that we don't check for bluetooth or NFC, only internet connectivity, which makes it loose its appeal IMO

yjkimjunior:
Agreed with @Tbaut for the most part. IMHO ideally we would block the user from doing anything at all period when any of the above (wifi, bluetooth, nfc) are detected as enabled, or more easier would be to check if airplane mode is enabled. Only then should we let them get to the AccountsList screen.

a) It's unrealistic to expect users to have a "burner" smart phone that is always offline.
b) even if we notify users whether they're secure or not, if someone f's something up using Signer while connected to a network, the burden still falls on us. Saying, "oh but didn't you notice the Not Secure message?" after someone lost money won't gain us any fans. Best to dummy proof from the start.

Tbaut:
> when any of the above (wifi, bluetooth, nfc) are detected as enabled, or more easier would be to check if airplane mode is enabled. Only then should we let them get to the AccountsList screen.

[...]

Issue #174, opened 2018-10-15
Show only "Not secure" warning if internet connection is detected. Never show "Secure" in the top right corner.


fckt:
### Motivation

Originally, the idea of the `Secure` message on the right corner was to state that device have no extra capabilities enabled (no internet, no bluetooth, no NFC, no IrDA, no USB cable connected), but only the `touch screen` and the `camera`. As it turns out we check only for an internet connection, as to check for other capabilities would require as to extend permissions for the app and will require to add libraries to work (to check in our case) with these capabilities. 

Showing the `Secure` message in case of checking only for the one type of connection (internet) could (and will) mislead user.

Related issues:
https://github.com/paritytech/parity-signer/issues/121
https://github.com/paritytech/parity-signer/issues/138
https://github.com/paritytech/parity-signer/issues/122

Tbaut:
Hey @fckt you assigned that one to yourself a while back, I'll do it if that's ok.

PR #286, opened 2019-07-25
Security header refresh

Tbaut:
- closes https://github.com/paritytech/parity-signer/issues/266
- closes https://github.com/paritytech/parity-signer/issues/174

- "Secure" will never be shown again, as we can't be sure and it's a strong word.
- update to the new [RNCNetinfo api](https://github.com/react-native-community/react-native-netinfo#usage) (this removes a deprecation warning)
- get rid of SecurityStore and `unstated` shared state and consolidate the little logic in one screen
- remove another android permission that seems unneeded by NetInfo

Todo

- [x] Triple check that the removed permission doesn't affect how things work even on an apk. Done, built an apk and it launched and worked without trouble. :white_check_mark:


hanwencheng:
out
Meet weird problem on iOS simulator, the state does not change when user click back.
I also set the breakpoint there and the listener is not called.

Need to be checked on real iOS device, probably is because simulator's problem

hanwencheng:
ignore it since it is a [known issue](https://github.com/react-native-community/react-native-netinfo#issues-with-the-ios-simulator)

yjkimjunior:
LGTM!

YljP3AEUYI

Tbaut:
Real iPhone test successful by @yjkimjunior :)

PR #287, opened 2019-07-25
chore(): update headers

Tbaut:
self explanatory


hanwencheng:
What about this one: `Parity Technologies Deutschland GmbH `?

Tbaut:
Parity Tech is mostly a UK company, just check any [other header](https://github.com/paritytech/substrate/blob/master/subkey/src/main.rs), there's no gmbh mentioned.

Issue #268, opened 2019-07-22
Some text fields don't have "done" option for exiting text input (iOS)


phillux:
v2.2.1 / iOS / iPhone X

Some text fields, such as the "Account name" and "Enter recovery words" fields, do not have the typical "done" button normally found on above the keyboard. This forces the user to tap an area of the screen that may cause future input errors. Scrolling is also not often possible, prohibiting the user from going to the next step.

Tbaut:
![image](https://user-images.githubusercontent.com/33178835/61623669-62316200-ac77-11e9-8297-ce0ca6d2cd10.png)

Tbaut:
Closing as this shouldn't be needed anymore. Also #278 will probably improve the keyboard situation a bit more.

PR #291, opened 2019-07-29
Fix keyboard aware scroll view related problem

### LINKED ISSUE #278: Double tap needed in scrollview
How to reproduce (on Android not sure about iOS):
- recover an account
- tap the "name" field to enter a name.
- top the "next step" button..  you need to tap it twice.

It happens on account creation as well and sometimes when choosing network.

This has been introduced here https://github.com/paritytech/parity-signer/pull/272#pullrequestreview-264780795 and is linked to the use of `ScrollView` and potentially `KeyboardAwareScrollView`. 

I was unlucky trying things like described [here](https://github.com/APSL/react-native-keyboard-aware-scroll-view#android-support), [here](https://github.com/APSL/react-native-keyboard-aware-scroll-view/issues/233#issuecomment-379583908) or everywhere on the internet with `keyboardDismissMode` and `keyboardShouldPersistTaps="always"`

This scroll things needs love.

hanwencheng:
fix #278 

* 96c3539: fix double tap problem on Android and half covered text input on iOS related to

 #268
* 10e97b4: every click on seed input field will trigger scroll 
* 0d29f30: align seed input to top on android, so iOS and Android are same now
* b40b6da: prettier the code and remove useless import

Screenshots here (too big cannot upload on Github):  
https://drive.google.com/file/d/1g-o_XmShquI9U52agOQUl4VBX9jBujOE/view

### DIFF
--- a/src/components/AccountSeed.js
+++ b/src/components/AccountSeed.js
@@ -132,6 +132,7 @@ export default class AccountSeed extends Component {
           style={[styles.input, invalidStyles]}
           multiline
           autoCapitalize="none"
+          textAlignVertical="top"
           onSelectionChange={this.handleCursorPosition}
           {...this.props}
         />
--- a/src/screens/AccountRecover.js
+++ b/src/screens/AccountRecover.js
@@ -33,6 +33,7 @@ import AccountsStore from '../stores/AccountsStore';
 import { validateSeed } from '../util/account';
 import { debounce } from '../util/debounce'
 import { brainWalletAddress } from '../util/native';
+import KeyboardScrollView from "../components/KeyboardScrollView";
 
 export default class AccountRecover extends React.Component {
   static navigationOptions = {
@@ -62,36 +63,25 @@ class AccountRecoverView extends React.Component {
     brainWalletAddress(seed)
     .then(({ address, bip39 }) => accounts.updateNew({address, seed, validBip39Seed: bip39}))
     .catch(console.error);
-  }
+  };
 
   debouncedAddressGeneration = debounce(this.addressGeneration, 200)
   
   componentWillUnmount = function () {
     // called when the user goes back, or finishes the whole recovery process
     this.props.accounts.updateNew({seed : ''});
-  }
+  };
 
   render() {
     const { accounts } = this.props;
     const selected = accounts.getNew();
     const networkKey = selected.networkKey;
     const network = NETWORK_LIST[networkKey];
-
     return (
       <SafeAreaView style={styles.safeAreaView}>
-        <KeyboardAwareScrollView style={styles.bodyContainer}>
+        <KeyboardScrollView style={styles.bodyContainer}
+                            contentContainerStyle={{ justifyContent: 'flex-end' }}>
           <Background />
-          <ScrollView
-            contentContainerStyle={{ justifyContent: 'flex-end' }}
-            style={{ flex: 1 }}
-            enableOnAndroid
-            scrollEnabled
-            keyboardShouldPersistTaps="always"
-            extraHeight={230}
-            innerRef={ref => {
-              this.scroll = ref;
-            }}
-          >
             <Text style={styles.titleTop}>RECOVER ACCOUNT</Text>
             <Text style={styles.title}>CHOOSE NETWORK</Text>
             <NetworkButton network={network}/>
@@ -169,8 +159,7 @@ class AccountRecoverView extends React.Component {
                 });
               }}
             />
-          </ScrollView>
-        </KeyboardAwareScrollView>
+        </KeyboardScrollView>
       </SafeAreaView>
     );
   }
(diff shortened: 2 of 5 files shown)


src/screens/AccountRecover.js
```diff
@@ -17,8 +17,13 @@
...
 
 import React from 'react';
-import { Alert, SafeAreaView, ScrollView, StyleSheet, Text } from 'react-native';
-import { KeyboardAwareScrollView } from 'react-native-keyboard-aware-scroll-view'
+import {
+  Alert,
+  SafeAreaView,
+  StyleSheet,
+  Text,
+  findNodeHandle
```
Tbaut: (nit) alphabetical order, `findNodeHandle` should be placed right after `Alert`

src/screens/AccountRecover.js
```diff
@@ -31,8 +36,9 @@ import TextInput from '../components/TextInput';
 import { NETWORK_LIST } from '../constants';
 import AccountsStore from '../stores/AccountsStore';
 import { validateSeed } from '../util/account';
-import { debounce } from '../util/debounce'
+import { debounce } from '../util/debounce';
 import { brainWalletAddress } from '../util/native';
+import KeyboardScrollView from '../components/KeyboardScrollView';
```
Tbaut: (nit) alphabetical order, because the import is from `../components/KeyboardScrollView`, it should be placed after `./components/Button`

src/components/AccountSeed.js
```diff
@@ -132,6 +132,7 @@ export default class AccountSeed extends Component {
           style={[styles.input, invalidStyles]}
           multiline
           autoCapitalize="none"
+          textAlignVertical="top"
```
Tbaut: Is this really needed? I think it looks better when the text is centred.

src/screens/AccountRecover.js
```diff
@@ -53,124 +59,121 @@ class AccountRecoverView extends React.Component {
...
+
+  componentWillUnmount = function() {
     // called when the user goes back, or finishes the whole recovery process
-    this.props.accounts.updateNew({seed : ''});
-  }
+    this.props.accounts.updateNew({ seed: '' });
+  };
 
   render() {
     const { accounts } = this.props;
```
Tbaut: (nit), it's not from your PR, but as I see it..
```suggestion
    const { accounts, navigation } = this.props;
```

src/screens/AccountRecover.js
```diff
@@ -53,124 +59,121 @@ class AccountRecoverView extends React.Component {
...
+              );
+
+              if (!validation.valid) {
+                if (validation.accountRecoveryAllowed) {
+                  return Alert.alert('Warning:', `${validation.reason}`, [
+                    {
+                      text: 'I understand the risks',
+                      style: 'default',
+                      onPress: () => {
+                        this.props.navigation.navigate('AccountPin', {
```
Tbaut: not needed anymore (see previous comment)
```suggestion
                        navigation.navigate('AccountPin', {
```

src/screens/AccountRecover.js
```diff
@@ -53,124 +59,121 @@ class AccountRecoverView extends React.Component {
...
+                          isWelcome: this.props.navigation.getParam(
```
Tbaut: not needed anymore (see previous comment)
```suggestion
                          isWelcome: navigation.getParam(
```

src/screens/AccountRecover.js
```diff
@@ -53,124 +59,121 @@ class AccountRecoverView extends React.Component {
...
-                  isWelcome: this.props.navigation.getParam('isWelcome'),
-                  isNew: true
-                });
-              }}
-            />
-          </ScrollView>
-        </KeyboardAwareScrollView>
+              }
+
+              this.props.navigation.navigate('AccountPin', {
```
Tbaut: same here
```suggestion
              navigation.navigate('AccountPin', {
```

src/screens/AccountRecover.js
```diff
@@ -53,124 +59,121 @@ class AccountRecoverView extends React.Component {
...
+                isWelcome: this.props.navigation.getParam('isWelcome'),
```
Tbaut: and here
```suggestion
                isWelcome: navigation.getParam('isWelcome'),
```

src/components/AccountSeed.js
hanwencheng: Basic idea is to keep the design keep the same in both Android and iOS. Tried "alignment" to the middle, it will cause the scroll view to first move up, and move a little down after the first keystroke. I do not have a preference regarding the position, so I will just choose the one without issues then.

hanwencheng:
YJ gave the iphone to me, I will test it later today.

hanwencheng:
out

Tested on iPhone, looks good.

PR #292, opened 2019-07-29
fix: change button default color

### LINKED ISSUE #276: Scan/Next step buttons remain greyed out irrespective of they are activated or not (Android)
Going by an ideal mobile design guideline, the button should be greyed out only if it is not activated i.e clicking on them carries no action. Once they are capable of carrying out an action the color of button should be changed from grey to some solid color (example green).

This has not been followed in Parity-signer, making the user confused if the button is supposed to work or not.

hanwencheng:
resolve #276.

Just use the existed neon green. Supposed the whole UI will be designed in the future.

Screenshot 2019-07-29 at 18 39 15


Tbaut:
What about something like `#445a6d` https://coolors.co/1d1d1d-445a6d-9e77fe-f2e364-f9f9f9 ? It's not grey, dark yet visible on black

hanwencheng:
I am up for that bright one #48acf0, tried on device and the dark one is really confusing with the disable button.

src/colors.js
```diff
@@ -16,6 +16,7 @@
 
 export default {
   bg: '#1D1D1D',
+  bg_button: '#48acf0',
   bg_text: '#FFFFFF',
   bg_text_sec: '#B4B5B0',
   bg_positive: '#72E762',
```
Tbaut: Can we remove this `bg_positive` if not used anywhere anymore?

hanwencheng: I guess you mean `bg_text_positive`, `bg_positive` is in use now

hanwencheng: wait a minute, let me check where are these `bg_positive` are used...

src/screens/TxDetails.js
```diff
@@ -106,7 +106,7 @@ export class TxDetailsView extends React.PureComponent {
           networkKey={this.props.recipient.networkKey || ''}
         />
         <Button
-          buttonStyles={{ backgroundColor: colors.bg_positive, height: 60 }}
+          buttonStyles={{ height: 60 }}
           title="Sign Transaction"
           textStyles={{ color: colors.card_text }}
```
Tbaut: found another nit!
the "Sign transaction" button is written black on blue whereas all others are white on blue.  Can we make sure to have the same styling?

src/screens/MessageDetails.js
```diff
@@ -105,7 +105,7 @@ export class MessageDetailsView extends React.PureComponent {
             : this.props.data}
         </Text>
         <Button
-          buttonStyles={{ backgroundColor: colors.bg_positive, height: 60 }}
+          buttonStyles={{ height: 60 }}
           title="Sign Message"
           textStyles={{ color: colors.card_text }}
```
Tbaut: I imagine that this makes the Text black here as well. Could we have it white?

hanwencheng: nice catch!

PR #293, opened 2019-07-30
Substrate accounts management

Tbaut:
-  Allow creation and recovery of substrate accounts
- Show `/soft` and `//hard` derivation path on account backup screen
- Allow to check the derivation path`password` for any `///password`

demo
demo2

todo (another PR):
- [ ] make sure we lock accounts as soon as we're done using the seed of an account (sign message or tx)


src/util/account.js
```diff
@@ -1,28 +1,34 @@
...
+  protocol = NetworkProtocols.SUBSTRATE,
+  publicKey,
+  networkKey = SubstrateNetworkKeys.SUBSTRATE
 }) {
   if (typeof address !== 'string' || address.length === 0) {
-    throw new Error(`Couldn't create an accountId, missing address`);
+    throw new Error(`Couldn't create an accountId, address missing`);
+  }
+  if (protocol === NetworkProtocols.SUBSTRATE){
+    return `${protocol}:${publicKey}`;
```
yjkimjunior: @Tbaut what do you think about we format this instead to return `${protocol}:${ss58Address}:${genesisHash}`?

Tbaut: yup that's what needs to be done now that we have the `genesisHash`.

src/screens/MessageDetails.js
```diff
@@ -92,23 +92,23 @@ export class MessageDetailsView extends React.PureComponent {
...
           }}
         />
         <Text style={styles.title}>MESSAGE</Text>
         <Text style={styles.message}>
-          {isAscii(this.props.message)
-            ? hexToAscii(this.props.message)
-            : this.props.data}
+          {isAscii(message)
+            ? hexToAscii(message)
+            : data}
```
yjkimjunior: i'm confused here. If message is Ascii, treat is as a hex and do a conversion on it to Ascii, otherwise show the raw hex data? Seems like this bit has been around since pre v2.2 so I'm surprised it hasn't been caught but should it be 
```
isAscii(message) ? message : hexToAscii(data)
```
?

Tbaut: yup, it def has been around for a while. Never tried to signed data that's not a Ascii though.
you suggest `hexToAscii(data)`, but that wouldn't look good if the hex isn't hex encoded ascii right? why not `isAscii(message) ? message : data` ?

yjkimjunior: that makes more sense

src/components/AccountIconChooser.js
```diff
@@ -31,49 +31,66 @@ import Address from './Address'
...
+
+    if (protocol === NetworkProtocols.ETHEREUM){
+      try {
+        const icons = await Promise.all(
+          Array(4)
+            .join(' ')
+            .split(' ')
+            .map(async () => {
+              const seed = await words();
+              const { address, bip39 } = await brainWalletAddress(seed);
```
yjkimjunior:
```suggestion
              const { address, bip39 } = protocol === NetworkProtocols.ETHEREUM ? await brainWalletAddress(seed) : await substrateAddress(seed, prefix);
```

Tbaut: Thanks, I'll use an if because `substrateAddress` doesn't return an object (no `bip39`).

src/components/DerivationPathField.js
```diff
@@ -0,0 +1,90 @@
...
+  Text,
+  TouchableOpacity,
+  View
+} from 'react-native';
+import Icon from 'react-native-vector-icons/MaterialIcons';
+
+import keyExtract from '../util/keyExtract'
+import TextInput from './TextInput';
+
+  export default function DerivationPathField(props) {
```
yjkimjunior: weird indentation :/

src/screens/AccountRecover.js
```diff
@@ -22,7 +22,7 @@ import {
   findNodeHandle,
```
yjkimjunior: when entering a recovery phrase, the field shows green as long as the words are bip39, and it'll let me continue to create pin and confirm, but crashes if the entire prhase was not actually valid. 
Screenshot 2019-08-20 at 16 13 43

src/screens/AccountBackup.js
```diff
@@ -110,10 +120,15 @@ class AccountBackupView extends React.PureComponent {
...
           <Text style={styles.seedText}>
-            {selected.seed}
+            {seedPhrase || seed}
           </Text>
         </TouchableItem>
-        {(isNew) &&
+        {!!derivationPath &&
+          <Text style={styles.derivationText}>
+            {derivationPath}
+          </Text>}
```
yjkimjunior: Not sure that this derivatio password verify should be here or not

Screenshot 2019-08-20 at 16 31 05

src/screens/AccountRecover.js
```diff
@@ -98,11 +144,11 @@ class AccountRecoverView extends React.Component {
           <Background />
           <Text style={styles.titleTop}>RECOVER ACCOUNT</Text>
           <Text style={styles.title}>CHOOSE NETWORK</Text>
-          <NetworkButton network={network} />
+          <NetworkButton network={selectedNetwork} />
```
yjkimjunior: small thing, but i believe now we should change the order of these networks, i.e. Kusama, Polkadot at the top, and the rest as is.

src/screens/AccountBackup.js
Tbaut: we should not show the password in the UI, but we need to show a hint and a way for users to verify it. I thought about another screen for that, but ended up doing it on the same to avoid more navigation.

yjkimjunior: let's log this as an issue for another PR then

src/screens/AccountRecover.js
Tbaut: This is actually linked to the networkKey, ethereuem with 1, 2.. is getting first. There will be only Kusama for v3 and it'll be the default, so I guess it's ok.

src/screens/AccountRecover.js
Tbaut: should be solved by the last commit.

src/constants.js
```diff
@@ -17,9 +17,9 @@ export const EthereumNetworkKeys = Object.freeze({
 
 // genesisHash is used as Network key for Substrate networks
 export const SubstrateNetworkKeys = Object.freeze({
-  // POLKADOT: '123',
   KUSAMA: '456',
```
yjkimjunior: @Tbaut I'm fine with merging as is and coming back to update this, unless there are objections.

Tbaut: right, we need several updated here and there for the readme, the T&Cs as well, so we can do that later.

burdges:
We might eventually want to distinguish kusama keys in which the derivation got missused https://github.com/paritytech/substrate/issues/3396 although we're likely to choose to do some ugly hacks so that this can stay the same.

Issue #295, opened 2019-07-31
Icon selection revamp for Substrate


Tbaut:
To have something consistant and workable with the substrate different addresses, I propose to the the following revamp. No big change, just that we won't pre-calculate 10 addresses and icons but rather 1 (the advantage is that the icon will load much faster) and we add a refresh button to change the icon and addresses if needed.

With the substrate several addresses it would look something like:
![image](https://user-images.githubusercontent.com/33178835/62200821-76b3df80-b386-11e9-9195-0fdbf667f081.png)

needed as part of https://github.com/paritytech/parity-signer/pull/293

Issue #179, opened 2018-10-22
Address's end is not displayed


Tbaut:
The full address isn't visible on a 1080 x 1920 pixels, 16:9 ratio (~401 ppi density) device:
![image](https://user-images.githubusercontent.com/33178835/47228990-e132fa80-d3c6-11e8-9d48-796da41f787c.png)

Low-end and older devices (the targeted phones) have potentially a lower resolution. [Here is a subjective view](https://deviceatlas.com/blog/most-used-smartphone-screen-resolutions#worldwide) of common device resolutions per country. We could make sure the full address is correctly displayed on a 480px wide screen. We should also probably show the 4/6 first and last digits on the main accounts overview screen, while make sure to show the full address on the account screen, something along the line:

![image](https://user-images.githubusercontent.com/33178835/47230811-86e86880-d3cb-11e8-8f29-fa62c3ec4653.png)

ltfschoen:
Agreed, this is a major user annoyance. I just created an account and was going to send ETH to it, but I have to import it into Fether just to be able to see my Ethereum address to send the amount to.

Tbaut:
Should be better now.

Issue #230, opened 2019-03-04
Allow screenshot in development mode


Tbaut:
Since https://github.com/paritytech/parity-signer/pull/187 it's not possible to take screenshots anymore.

I'd find it very convenient in development mode to be able to:
- be able to stream my screen (for presentations)
- take screen shots

PR #296, opened 2019-07-31
Tbaut icon rotation

Tbaut:
- closes https://github.com/paritytech/parity-signer/issues/295
- introduces the `Address` component that will cater for showing Ethereum or Substrate address(es) and added a `short` prop, helpful to tackle https://github.com/paritytech/parity-signer/issues/179 later on.

How it looks like (this is a real device with mirrored screen)
rotation


src/components/AccountIconChooser.js
```diff
@@ -16,143 +16,106 @@
...
     }
   };
 
-  renderIcon = ({ item, index }) => {
-    const { value, onSelect } = this.props;
-    const { address, bip39, seed } = item;
-    const iSelected = address.toLowerCase() === value.toLowerCase();
-    const style = [styles.icon];
+  renderIcon = () => {
+    const { address, bip39, seed } = this.state;
```
hanwencheng: `bip39` and `seed` are unused variables here.

src/components/AccountIconChooser.js
```diff
@@ -16,143 +16,106 @@
...
-  };
+import AccountIcon from './AccountIcon';
+import Address from './Address'
 
+export default class AccountIconChooser extends React.PureComponent {
   constructor(props) {
     super(props);
 
-    this.state = { icons: [] };
+    this.state = {};
```
hanwencheng: IMHO, put all the state even the `null` one in the constructor will give a quick overview of the whole component's structure. For example, I am a little bit surprised when I see that line in the later code.
```javascript
const { address, bip39, seed } = this.state;
```

hanwencheng:
LGTM, leave some trivial comments.

maciejhirsz:
Not sure if I should comment here or on the issue. What is the exact UX issue with having multiple icons show up at the same time? I think this is a UX downgrade tbh, eyes are random access and can scan multiple icons.

On Ethereum the UI is responsive with multiple icons, even on old hardware. Are Polkadot icons slow to render?

Tbaut:
The idea is keep a simple interface, although we'll add more info on the screen (with the Polkadot addresses per prefix). Having all the icons + the several addresses looks messy.

The icons load in a reasonable amount of time on Ethereum the first time you load it, but I wouldn't call it snappy.
It's too early to say regarding Polkadot icons since I don't have the lib working on RN yet, however having the whole icons reload if we change the network is, again, not nice. The current blockies are rendered natively (with binding), I expect Polkadot to be slower if done in JS.

I'd argue that if you care about having a really nice icon, you won't mind clicking on the refresh button a couple time. However in the previous version, if the ones selected for you didn't match your tastes, there was no refresh button, you needed to go back and start the flow over.

maciejhirsz:
I had to double check as it's literally instant on my iPhone with the app stock from the AppStore. I think this got worse when we switched to BIP39, as even with parity phrases the bulk of the time was spent doing phrase -> private key -> public key -> address derivation, not actually rendering the icon, and that is way more computing intensive with BIP39 than it was with parity phrases. So that's fair, I don't know if that alone warrants kicking off multiple icons but eh :P.

As for #179 - this layout actually further reduces the horizontal space available for the address in creation. Currently on the iPhone you can actually read the entire address during creation, with this change you won't be able to.

I don't have any suggestions on how to improve the layout atm. I can think on this more, but I understand we are being rushed, so I'm not going to veto the PR :stuck_out_tongue:. The mockup at #295 basically requires a redesign the moment we have more than one prefix anyway.

Tbaut:
I'll try to hack a compromise
image

Tbaut:
here we go with the following:

rotation

android/app/src/main/java/io/parity/signer/MainActivity.java
```diff
@@ -19,13 +19,6 @@ protected String getMainComponentName() {
         return "NativeSigner";
     }
 
-    @Override
-    protected void onCreate(Bundle savedInstanceState) {
-        super.onCreate(savedInstanceState);
-        getWindow().setFlags(WindowManager.LayoutParams.FLAG_SECURE,
-                           WindowManager.LayoutParams.FLAG_SECURE);
-    }
-
```
hanwencheng: Why we re-enable screenshot functionality of the app? Is it an expected feature?

src/components/AccountIconChooser.js
```diff
@@ -66,93 +60,116 @@ export default class AccountIconChooser extends React.PureComponent<{
...
     } catch (e) {
       console.error(e);
     }
-  };
+  }
+
+  renderAddress = () => {
+    const {value} = this.props;
 
+    if (!!value) {
```
hanwencheng: Why using double esclamation here? should be same as `value` I think.

android/app/src/main/java/io/parity/signer/MainActivity.java
Tbaut: Ooops, forgot about that, I'm removing this part every time I want to share a screenshot with you guys because I'm on a real phone. Super annoying, looking forward for https://github.com/paritytech/parity-signer/issues/230

hanwencheng:
LGTM again.

PR #297, opened 2019-07-31
Fix typeface and maintain interface consistency

### LINKED ISSUE #270: Typeface inconsistency
v2.2.1 / iOS / iPhone X

There are currently 2 different fonts being used, and about 5 different styles.

- Helvetica/San Francisco (iOS default)
  - Normal, Bold
- Manifold (Parity font)
  - Normal, Bold, All Caps

"Accounts" screen:
- Top menu bar:
  - Logo and brandmark: Looks good, but there is just a bit too much space between the logo and type
  - "Not Secure" is in San Francisco font, I think this should be in Manifold as well
- Account element
  - Account name should be in Manifold, and perhaps bold

Another small copy suggestion:
- Should we put "Scan QR Code" on the button for added clarity?

hanwencheng:
resolve #270 

Generally using two kinds of the font in an app is not a good practice, so I change all the place with  font-family `Roboto` to `Manifold CF`, across the app now we have one font with three styles: `Manifold CF`, `Manifold CF Bold` and `Manifold CF DemiBold`

Changes are:
* the original font in Android does not work, changed the font name and make it work.
* Create `fonts.js` as util.
* Consistently use `Manifold CF` as the only default font of the app.
* Use `bold`, `demibold` font instead with software-rendered `font-weight`
 * fix some unfunctional font styles.
* change default Android font color from grey to black.
* reduce the padding between the logo and `Parity` in the left header according to #270 

Some problem needs to be solved in the future:

* Too many different font sizes, once we have a clear design, we may create an `AppText` component instead of React Native's `Text`
* Alert is still using the default system alert and thus the font can not be changed there, create a customized alert instead in the future, a promising library is [React-Native-Modal-Box](https://github.com/maxs15/react-native-modalbox).


Tbaut:
Thanks a lot for that cleanup.

I find the Manifold cool for titles, and eventually for things like the network card, however I find it hard to read as soon as the text is a little small or we write a couple words.

e.g the security text, the "enter a new account name", or any address is particularly hard to read (B, 8, 6 and 9 are similar).

I don't think it's a problem to have different fonts for titles and for content. I remember reading recommendations [here](https://www.smashingmagazine.com/2010/11/best-practices-of-combining-typefaces/) and there to use "a sans-serif font for your heading and serif font for your body text".

Ping @phillux since he raised the issue.

Pardon the yellowish color of my phone in night mode. 
image

PR #301, opened 2019-08-01
Polkadot identicons

Tbaut:
- bring in `@polkadot/reactnative-identicon` and the needed wasm deps. Striped down as much as I could. The building time is actually ok the second time (still last a couple min the first time)
- Set `--max_old_space_size=8192` in scripts to avoid out of memory problem
- upgrade `react-native` (0.60.4),` bignumber.js`, `hoist-non-react-statics`
- remove some unused deps: `react-native-simple-picker`, `yarn`

- added a sample of identicon, should be removed before merging. It should look like:
image


hanwencheng:
find one `jest` related problem on iOS, which block the building process:

Screenshot 2019-08-02 at 13 26 01

Solved with [root directory specification](https://github.com/facebook/jest/issues/6801#issuecomment-489702164) in `jest.config.js`

following is a simple config file:

```javascript
module.exports = {
  verbose: true,
  roots: ['<rootDir>/src'],
};
```

src/screens/AccountList.js
```diff
@@ -122,15 +121,35 @@ class AccountListView extends React.PureComponent {
...
                 { value: 'AccountNew', text: 'New Account' },
                 { value: 'AccountRecover', text: 'Recover Account' },
-                { value: 'About', text: 'About' }]}
+                { value: 'About', text: 'About' },
+              ]}
             />
           </View>
         </View>
-        { hasNoAccount && this.showOnboardingMessage()}
+        {hasNoAccount && this.showOnboardingMessage()}
```
Tbaut: Should be removed before merging

Tbaut:
ok I was able to create a signed and fully working apk.. although it takes ages, the building of the metro bundle is long the first time but quite snappy after.

@hanwencheng if you say it's working well on iphone, I'd remove my identicon samples and merge.

hanwencheng:
Got error on iOS

Screenshot 2019-08-05 at 22 55 35

It seems the `crypto` package is not indexed.

shim.js
```diff
@@ -0,0 +1,8 @@
+process.browser = true
+
+if (typeof global.Buffer !== 'undefined') {
+  // running on VSCode debugger
+  global.Buffer = undefined
+}
+
```
hanwencheng: As we are not using `nodify`, does these `Buffer` related code still needed?

yjkimjunior: +1, IIRC the shim.js was generated by the nodify --HACK script right?

Tbaut: Good point. It seems that `process.browser = true;` is still needed though. I removed the rest and things still work.. on android at least :+1:

(I get a "`process.split` undefined" if I remove the `process.browser = true;`)

Tbaut:
As we talked about in person with @hanwencheng, it seems that the 'crypto' override done [here](https://github.com/paritytech/parity-signer/pull/301/files#diff-fc10a6682269c1fefc2abe05d1500b8aR8) isn't picked-up correctly.

I'd advise to try in this order:

- pull the lastest changes and try again, I removed the crypto import in `shim.js`, this might help.

if things still don't work:

- make sure to do a `yarn clean`, and that all the commands in there work properly (install watchman if needed).

if things still don't work:

- play with `babel.rc` instead of `babel.config.js` and the rewrite require plugin. Have a look here https://github.com/polkawallet-io/polkawallet-RN/pull/142/files, it seems to work for them.

package.json
```diff
@@ -11,32 +11,43 @@
...
     "react-native-secure-storage": "git+https://github.com/paritytech/react-native-secure-storage.git",
-    "react-native-simple-picker": "^2.1.0",
+    "react-native-svg": "^9.5.3",
     "react-native-tabs": "^1.0.9",
     "react-native-vector-icons": "^6.6.0",
     "react-navigation": "^3.11.1",
+    "readable-stream": "^3.4.0",
     "unstated": "^2.1.1",
+    "vm-browserify": "1.1.0",
     "yarn": "^1.17.3"
```
yjkimjunior: yarn as a dependency?

shim.js
```diff
@@ -0,0 +1 @@
+process.browser = true;
```
yjkimjunior: this isn't a shim anymore, so maybe just set `process.browser = true` directly in App.js now?

src/screens/AccountList.js
```diff
@@ -30,10 +29,11 @@ import Button from '../components/Button';
 import AccountsStore from '../stores/AccountsStore';
 import { accountId } from '../util/account';
 import PopupMenu from '../components/PopupMenu';
+import Identicon from '@polkadot/reactnative-identicon';
```
yjkimjunior: import ordering

src/screens/AccountList.js
yjkimjunior: +1

shim.js
Tbaut: :man_shrugging: happy to remove that.

package.json
Tbaut: :man_shrugging: happy to remove that.

src/screens/AccountList.js
Tbaut: Sure, I'll remove that when removing the identicons from the main screnn

shim.js
Tbaut: Copying it in Apps throws
>undefined is not and object (evaluating process.version.split)

Note renaming it after discussion because it still is a black magic shim.

PR #309, opened 2019-08-07
fix react native warnings

### LINKED ISSUE #271: componentWillReceiveProps is deprecated and should be removed
We need to move to replace it with `getDerivedStateFromProps`

### LINKED ISSUE #277: componentWillMount is deprecated and should be removed
Use ComponentDidMount instead

hanwencheng:
close #271 , closes #277

* Remove all the `componentWillReceiveProps`, `componentWillMount`, `componentWillUpdate` method in our code base, whereas the warning with the dependency [react-native-markdown-renderer](https://github.com/mientjan/react-native-markdown-renderer). So I use React Native `ignoreWarning` method in the entry point of the app, i.e. `App.js`. in 
e036a5d

* cc53f25 , a68295b Refactored QrView and AccountIcon component with React hooks.

* 0df5839 DataDetail component is not used anywhere, and global `styles.js` file does not existed, so remove all of them.

* 4fd58b7 : Remove another warning of the iOS Native code threading problem


Tbaut:
Thanks a lot for that. We are re-factoring quite a lot with YJ right now in #293 and #279 and this will conflict quite a lot I suppose. I'll cherry pick your changes and add them to my branch, I think it'll be a headache otherwise.

hanwencheng:
I will merge this PR after these bigger ones are merged.

src/App.js
```diff
@@ -57,6 +57,18 @@ import TermsAndConditions from './screens/TermsAndConditions';
 import TxDetails from './screens/TxDetails';
 
 export default class App extends Component {
+
+  constructor(){
+    super();
+    if (__DEV__) {
+      YellowBox.ignoreWarnings([
```
Tbaut: Not sure it's a good idea to ignore them. I think it's ok to keep them

src/components/AccountIcon.js
```diff
@@ -17,54 +17,35 @@
 'use strict';
 
 import PropTypes from 'prop-types';
```
Tbaut: I've used hooks now in the substrate re-factoring for this file here: https://github.com/paritytech/parity-signer/pull/293/files#diff-4f18afcbf235db6a111ab00e2a6bee19L24

src/App.js
hanwencheng: It is just annoying, have to click the `dismiss` button everytime, plus now we have cleaned our code and know that warnings only comes from the dependencies. Let's keep eyes on new code of `componentWill*` method, I think there is little probabilities that we will use these lifecycles methods since we now have React Hooks to use.

src/components/AccountIcon.js
hanwencheng: Nice!

Tbaut:
All the major PRs have been now so that this one can go in :)

Tbaut:
Needs review before merging :disappointed:

Issue #313, opened 2019-08-12
Fix iphone splash screen


Tbaut:
On iphone, the apps shows a weird splash screen upon launch (the pause button is not from the splash screen, it's from my screenshot):
![image](https://user-images.githubusercontent.com/33178835/62858575-ff256f00-bcfa-11e9-8bd1-0bba69608fdd.png)

We should have a simple image containing the Signer logo:
![app-icon](https://user-images.githubusercontent.com/33178835/62858711-4e6b9f80-bcfb-11e9-9245-3ef9c4e86118.png)

hanwencheng:
Closed with #316 .

PR #316, opened 2019-08-12
chore: update splash screen

geastwood:
PR for #313 

Probably making it the background black is better.

Screenshot 2019-08-12 at 22 32 00


geastwood:
[clabot:check]

geastwood:
Updated the color.

Tbaut:
Thanks a lot, I'll let @hanwencheng test on an iphone (I don't have a mac :)) before merging

hanwencheng:
Tested on iOS devices. It all looks good, I just remove a duplicated image file in 54cf3a94d49c66691223dacd87c516d01c7c5ce2.

geastwood:
Thanks for merging!

What value does it bring to introduce `react-native-splash-screen`?

hanwencheng:
It hides the white react native bridging splash screen on both iOS and Android. 

Screenshot 2019-08-14 at 11 25 45

geastwood:
The screenshot is only for development. And RN uses for iOS the *LaunchScreen* which is the essential way for implementing splash screen. Afaik, the lib makes sense when you want to show a splash screen programmatically, e.g. show an advertisement while the user is back to the app, namely hooking into the lifecycle of user interaction.

hanwencheng:
Thanks for the explanation, we will ignore that screen then.

PR #328, opened 2019-08-23
Parse and construct function together

Tbaut:
Address the concerns raised in our discussions to have a parsing and constructing function together.


src/util/suri.js
```diff
@@ -36,6 +36,10 @@ export function parseSURI (suri) {
  */
 
 export function constructSURI ({ derivePath = '', password = '', phrase = '' }) {
+
+  if(derivePath === null || phrase === null) {
```
Tbaut: I think we should allow to have the `derivePath` to be null. Either way the fact that it's initialized to an empty string, It actually can't be null at this stage can it?

yjkimjunior: undefined would mean the value wasn't set, but null is an explicit assignment, so it wouldn't use the default empty string values. But yeah derivePath could be null, I'll fix that

src/components/AccountIconChooser.js
```diff
@@ -66,7 +66,14 @@ export default class AccountIconChooser extends React.PureComponent {
               Object.assign(result, await brainWalletAddress(result.seed));
             } else {
               try {
-                result.address = await substrateAddress(`${result.seed}${derivationPath}///${derivationPassword}`, prefix);
+
+                const suri = constructSURI({
+                  phrase: result.seed,
+                  derivePath: derivationPath,
+                  password:derivationPassword
```
maciejhirsz: formatting

src/screens/AccountNew.js
```diff
@@ -101,11 +102,17 @@ class AccountNewView extends React.Component {
               derivationPath={derivationPath}
               onSelect={({ newAddress, isBip39, newSeed }) => {
                 if (isSubstrate) {
+                  const suri = constructSURI({
+                    phrase: newSeed,
+                    derivePath: derivationPath,
+                    password:derivationPassword
```
maciejhirsz: formatting

src/screens/AccountRecover.js
```diff
@@ -96,7 +97,11 @@ class AccountRecoverView extends React.Component {
         )
         .catch(console.error);
     } else {
-      const suri = `${seedPhrase}${derivationPath}///${derivationPassword}`
+      const suri = constructSURI({
+        phrase: seedPhrase,
+        derivePath: derivationPath,
+        password:derivationPassword
```
maciejhirsz: formatting

src/components/DerivationPathField.js
```diff
@@ -58,13 +58,22 @@ export default function DerivationPathField(props) {
...
-            const derivationPath = parseSURI(text);
-    
-            onChange({
-              derivationPassword: derivationPath.password || '',
-              derivationPath: derivationPath.derivePath || ''
-            });
-            setIsValidPath(!!derivationPath.password || !!derivationPath.derivePath);
+            try {
+              const derivationPath = parseDerivationPath(text);
+              console.log('derivationPath', derivationPath)
```
Tbaut: stray comment

Tbaut:
Sorry @maciejhirsz for the extra work, I needed to refactor a bit to be able to keep relevant tests.
Please review my last 2 commits (the one before the last being the most critical.)

src/screens/AccountRecover.js
```diff
@@ -43,6 +43,7 @@ import AccountsStore from '../stores/AccountsStore';
 import { empty, validateSeed } from '../util/account';
 import { debounce } from '../util/debounce';
 import { brainWalletAddress, substrateAddress } from '../util/native';
+import {constructSURI} from "../util/suri";
```
yjkimjunior: single quote

src/screens/AccountNew.js
```diff
@@ -32,6 +32,7 @@ import { NETWORK_LIST, NetworkProtocols } from '../constants';
 import fonts from '../fonts';
 import AccountsStore from '../stores/AccountsStore';
 import { empty, validateSeed } from '../util/account';
+import {constructSURI} from "../util/suri";
```
yjkimjunior: singel quote

src/screens/AccountRecover.js
```diff
@@ -85,26 +87,50 @@ class AccountRecoverView extends React.Component {
...
+          password: derivationPassword,
+          phrase: seedPhrase
+        });
+
+        substrateAddress(suri, prefix)
+          .then((address) => {
+            accounts.updateNew({ address, derivationPath, derivationPassword, seed: suri, seedPhrase, validBip39Seed: true })
+          })
+          .catch(() => {
+            //invalid phrase
```
hanwencheng: This one error catch may be redundant?

hanwencheng:
Tested on iOS, LGTM, only one minor comment.

src/screens/AccountRecover.js
Tbaut: I don't think so since the catch is specific to `substrateAddress` whereas the other is for `constructSURI`

PR #332, opened 2019-08-26
QR Code Scanner Fixes

### LINKED ISSUE #247: Show a friendly message if an address is scanned
from #245 

Users might be used to wallet applications that allow to scan a receiver's address to send funds to. Instead of showing a general error message, we can show a dedicated  message if we detect that the QR code is in the form `0x...` (20 bytes) or `ethereurm:0x..` such as:

> **Ethereum address detected**
You need to create a transaction using a software such as MyCrypto or Fether so that Parity Signer can sign it.

hanwencheng:
fix #247 

mainly 3 fixes:

a15259d: fix a scanning user UX issue, after unavailable QR code, only trigger the scan when the user clicks the button.

out mp4

b5e239f: add address detected warning.

eaf9f52: remove an unhandled promise rejection warning, code should return when error detected.

### DIFF
--- a/src/screens/QrScanner.js
+++ b/src/screens/QrScanner.js
@@ -162,7 +162,7 @@ export class QrScannerView extends React.PureComponent {
 
 const text = {
   ADDRESS_ERROR_TITLE: 'Address detected',
-  ADDRESS_ERROR_MESSAGE: 'Please create a raw transaction using a software such as MyCrypto or Fether so that Parity Signer can sign it.',
+  ADDRESS_ERROR_MESSAGE: 'Please create a transaction using a software such as MyCrypto or Fether so that Parity Signer can sign it.',
   PARSE_ERROR_TITLE: 'Unable to parse transaction'
 };
 
--- a/src/screens/QrScanner.js
+++ b/src/screens/QrScanner.js
@@ -26,7 +26,7 @@ import colors from '../colors';
 import fonts from '../fonts';
 import AccountsStore from '../stores/AccountsStore';
 import ScannerStore from '../stores/ScannerStore';
-import { isJsonString, rawDataToU8A } from '../util/decoders';
+import {isAddressString, isJsonString, rawDataToU8A} from '../util/decoders';
 
 export default class Scanner extends React.PureComponent {
   static navigationOptions = {
@@ -37,6 +37,19 @@ export default class Scanner extends React.PureComponent {
   constructor(props) {
     super(props);
     this.state = { enableScan: true };
+  }
+
+  showErrorMessage(scannerStore, title, message) {
+    this.setState({ enableScan: false });
+    Alert.alert(title, message, [
+      {
+        text: 'Try again',
+        onPress: () => {
+          scannerStore.cleanup();
+          this.setState({ enableScan: true });
+        }
+      }
+    ]);
   }
 
   render() {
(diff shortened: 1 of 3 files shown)


src/util/decoders.js
```diff
@@ -189,28 +205,34 @@ export function decodeToString(message: Uint8Array): string {
 
 export function asciiToHex(message: string): string {
   var result = [];
-	for (let i = 0; i < message.length; i++) {
-		var hex = Number(message.charCodeAt(i)).toString(16);
-		result.push(hex);
+  for (let i = 0; i < message.length; i++) {
+    var hex = Number(message.charCodeAt(i)).toString(16);
```
yjkimjunior: grumbling at myself: 
```suggestion
    let hex = Number(message.charCodeAt(i)).toString(16);
```

src/util/decoders.js
```diff
@@ -189,28 +205,34 @@ export function decodeToString(message: Uint8Array): string {
...
+  return result.join('');
 }
 
 export function hexToAscii(hexBytes: Uint8Array): string {
-	var hex  = hexBytes.toString();
-	var str = '';
-	for (var n = 0; n < hex.length; n += 2) {
-		str += String.fromCharCode(parseInt(hex.substr(n, 2), 16));
-	}
+  var hex = hexBytes.toString();
```
yjkimjunior:
```suggestion
  let hex = hexBytes.toString();
```

src/util/decoders.js
```diff
@@ -189,28 +205,34 @@ export function decodeToString(message: Uint8Array): string {
...
+  var str = '';
```
yjkimjunior:
```suggestion
  let str = '';
```

src/util/decoders.js
```diff
@@ -189,28 +205,34 @@ export function decodeToString(message: Uint8Array): string {
...
+  for (var n = 0; n < hex.length; n += 2) {
```
yjkimjunior:
```suggestion
  for (let n = 0; n < hex.length; n += 2) {
```

src/screens/QrScanner.js
```diff
@@ -143,6 +160,12 @@ export class QrScannerView extends React.PureComponent {
   }
 }
 
+const text = {
+  ADDRESS_ERROR_TITLE: 'Address detected',
+  ADDRESS_ERROR_MESSAGE: 'Please create a raw transaction using a software such as MyCrypto or Fether so that Parity Signer can sign it.',
```
Tbaut:
```suggestion
  ADDRESS_ERROR_MESSAGE: 'Please create a transaction using a software such as MyCrypto or Fether so that Parity Signer can sign it.',
```

src/util/decoders.js
hanwencheng: Here we can use const I think, each loop is a separate block (scope), so they do not conflict with each other.

src/util/decoders.js
hanwencheng: also `const` here?

hanwencheng:
@Tbaut I agree, before I make the commit, I just do `prettier --write` with the file. It would be nice if we have a lint standard and to integrate the lint into CI test.

src/util/decoders.js
yjkimjunior: yeah :[

yjkimjunior:
works well on iOS! thank you!

PR #337, opened 2019-08-27
fix: add Linter and fix caught errors

### LINKED ISSUE #335: Add linter and enforce rules accross the code
We use different linter with different rules.
We should enforce one for the project and fix most issues in a PR.

hanwencheng:
close #335 

`yarn` is needed before test the code. I have pointed out some critical code modification part in the comments.

Though the diff on Github may show the tabs in 8 spaces, the file view should be correct.

It includes following fixes:
* Init the linter configuration and add the linter into our CI.
* add [husky](https://github.com/typicode/husky) to configure pre-commit with lint.
* fix caught errors by Eslint.
* Fix the indentation from space to tab in project-scope.


package.json
```diff
@@ -14,7 +14,7 @@
     "clean": "watchman watch-del-all && rm -rf /tmp/metro-bundler-cache-* && rm -rf /tmp/haste-map-react-native-packager-* && rm -rf node_modules/ && yarn cache clean --force",
     "commit": "commit-wizard",
     "ios": "yarn run build-rust-ios && react-native run-ios",
-    "lint": "eslint",
+    "lint": "./node_modules/.bin/eslint . --fix",
```
Tbaut: what about `npx eslint`. Looks cleaner to me.

```suggestion
    "lint": "npx eslint . --fix",
```

maciejhirsz: Unrelated to that, I wouldn't expect that calling `lint` would trigger a `fix` by default, that should rather be another script, like `(npm run|yarn) lint:fix`.

yjkimjunior: ^ +1, seems the "fix" is getting tripped up with the slashes in `suri.spec.js` and making the build fail.

yjkimjunior:
I personally only have these minor pet peeves:
https://eslint.org/docs/rules/
```
"comma-dangle": ["error", "never"],
"indent": ["error", "tab"],
"object-curly-spacing": "always",
"quotes": "single",
"sort-keys": ["error", "asc", {"caseSensitive": true, "natural": false, "minKeys": 2}]
```

package.json
hanwencheng: I think `npx` wouldn't work in CI, they do not have global context. I will use `lint:fix` for that

hanwencheng: So true, updated!

Tbaut:
Is this ready to review? Please change the title and labels accordingly.

hanwencheng:
It is still under work, as long as I finished, will change labels.

hanwencheng:
Fixed all of the errors, and most of the warnings, but still not sure why the Github diff is displayed the tabs [incorrectly](https://github.com/isaacs/github/issues/170#issuecomment-150489692) with 8 spaces in some file like `App.js`, the indentation should all looks good with local Editors. Tried to fix it in #357 . I already got crazy with this indentation display.

Force pushed the commits, now they should look clear.
0f8b2ea: Most of the configurations related to ESLint.
79f6c42: ESLint fixes. (Too much files, maybe just test the app).

hanwencheng:
The `Missing trailing comma.eslint(comma-dangle)` conflict should be resolved with the rules set in `.eslintrc`by YJ's suggestion. @Tbaut

Tbaut:
Thanks a lot for that. Since there are tons of changes hard to review, represent no added value for our users and potential to break things, we should not merge this PR before the release of v3.

yjkimjunior:
@hanwencheng could we also have `no-unused-vars` enabled please?

hanwencheng:
As v3.0 is released, I will remove the `on ice` label and move this into milestone v3.1.

Tbaut:
Please don't merge. V3.0 **is not** released. @hanwencheng

Tbaut:
@hanwencheng I'd be happy to merge this now before it conflicts more :)

hanwencheng:
@yjkimjunior `no-unused-vars` is already included in the `eslint-plugin-prettier`, so it is not needed to add it again. 🙂

src/screens/AccountNetworkChooser.js
```diff
@@ -20,104 +20,104 @@ import React from 'react';
...
-    const { accounts } = this.props;
+	render() {
+		const { navigation } = this.props;
+		const { accounts } = this.props;
 
-    return (
-      <ScrollView style={styles.body} contentContainerStyle={{ padding: 20 }}>
-        <Text style={styles.title}>CHOOSE NETWORK</Text>
-        { Object.entries(NETWORK_LIST)
-          .filter(([networkKey]) => networkKey !== UnKnownNetworkKeys.UNKNOWN )
```
hanwencheng: This name is undefined, so change it,

src/util/decoders.js
```diff
@@ -47,239 +51,266 @@ import { NETWORK_LIST, SUBSTRATE_NETWORK_LIST, SubstrateNetworkKeys } from '../c
...
+		if (e) {
+			throw new Error(e);
+		} else {
+			throw new Error('we cannot handle the payload: ', bytes);
+		}
+	}
 }
 
 export function decodeToString(message) {
-  const decoder = new TextDecoder('utf8');
```
hanwencheng: TextDecoder is in Web API, so it is changed to Javascript native function here.

src/util/suri.js
```diff
@@ -51,21 +53,21 @@ export function parseSURI (suri) {
...
-  const RE_CAPTURE = /^((?:\/\/?[^/]+)*)(?:\/\/\/(.*))?$/;
-  const matches = input.match(RE_CAPTURE);
-  let derivePath, password;
+export function parseDerivationPath(input) {
+	const RE_CAPTURE = /^((?:\/\/?[^/]+)*)(?:\/\/\/(.*))?$/;
+	const matches = input.match(RE_CAPTURE);
+	let derivePath, password;
 
-  if (matches) {
-    [_,derivePath = '', password = ''] = matches;
```
hanwencheng: `_` can be removed, it also trigger an `unused-var` error

src/util/decoders.js
```diff
@@ -47,239 +51,266 @@ import { NETWORK_LIST, SUBSTRATE_NETWORK_LIST, SubstrateNetworkKeys } from '../c
...
-        action =
-          firstByte === '00' || firstByte === '01'
-            ? 'signData'
-            : firstByte === '01'
-            ? 'signTransaction'
-            : null;
-        address = uosAfterFrames.substr(4, 44);
-
-        data['action'] = action;
-        data['data']['account'] = account;
```
hanwencheng: The account value should be assigned to `address` rather than `account`

Tbaut: Not sure I understand the comment.
Do you mean we should rename it? If yes then let's do it in another PR.

hanwencheng: This `account` variable is undefined in the file scope, I think it was a mistake, it should be `address` variable.

yjkimjunior:
@hanwencheng  Could we also resolve the `no-unused-vars` warnings please?

src/util/decoders.js
yjkimjunior: thanks for catching that. exactly why we need a linter. IMO this is fine to merge now, as only Ethereum UOS will be affected (which is not supported by anyone at the moment).

hanwencheng:
@yjkimjunior Thanks for remind me again, I just checked with `no-unused-vars`, it did find more errors here, seems it is not fully covered by `eslint-plugin-prettier`, sorry for not checking the previous comment.

yjkimjunior:
while we're on the topic of type checking, I don't believe our current `flow` setup is helping anyone at all. I'm not super familiar with `flow` but I expect it to red underline when, for example, I have an extra semi-colon (`var a = 2;;;`) or something dumb like that.

hanwencheng:
@yjkimjunior  The flow is mostly used in the Stores, I am not pretty sure how much it could help since I am not familiar with that. What do you suggestion? should we remove flow, the problem like `var a = 2;;;` is linted by `prettier/prettier`,  so I do not quite understand.

yjkimjunior:
> @yjkimjunior The flow is mostly used in the Stores, I am not pretty sure how much it could help since I am not familiar with that. What do you suggestion? should we remove flow, the problem like `var a = 2;;;` is linted by `prettier/prettier`, so I do not quite understand.

ok bad example, that would indeed be prettier's job. However what about if we're trying to sum a `number` and a `Uint8Array`? this is where flow should be red squiggly-ing but I don't see it (maybe something wrong with my personal setup?)

the `flowconfig` isn't even setup with an existing version, which is why I don't think it's ever worked before.

hanwencheng:
So since it does not work, let us open another issue to remove the flow type checker and use typescript instead, probably first in these stores? @yjkimjunior @Tbaut

yjkimjunior:
migrating to typescript will be quite cumbersome for a mediocre payoff at best, I'm in favor of just getting ourselves familiar with flow and incrementally improving the flow setup as things come up.

yjkimjunior:
happy to merge now, unless you have any final comments @Tbaut

hanwencheng:
There are still there annoying `non-unused-var` error.

yjkimjunior:
we should either escalate those `warning`'s to `error`'s so the CI marks them as build failing, or modify the `travis.yml` to do that.

yjkimjunior:
LGTM!

Issue #339, opened 2019-08-28
Can't sign payload with Substrate


Tbaut:
How to reproduce

- recover an account for Kusama with the following seed
>melt bleak cruel expect process mistake harbor very program fringe first escape spend convince park nephew phone obvious feel amazing borrow occur disorder clinic

This should generate the account E2Wv...94BT4e
- go to https://uostestqrs.herokuapp.com and copy/paste the address `E2WvmoNEWM8LUKQnw4VKXqbqTJKj2Md8tbpzs7NRt94BT4e` in the address field.

- scan a Substrate payload or blake2hash or oversized payload.
- It shows a black empty screen

yjkimjunior:
the `txRequestData.data` can be null, so `isAddressString` should check for that. More generally we should also add an alert in case anything fails so we don't fall to a black screen again. I will make a PR for that

hanwencheng:
After some check, get some clues here, hope that could help.

Firstly, some undefined error comes from this [line](https://github.com/paritytech/parity-signer/blob/master/src/util/account.js#L8), and throw an error if the `protocol` is not defined. Here we should prove that the target `networkKey` exists in the list `NETWORK_LIST`.

The above function is called by the following method as @yjkimjunior said:
![Screenshot 2019-08-28 at 14 42 28](https://user-images.githubusercontent.com/6014309/63856631-66514d80-c9a2-11e9-9996-a14b185cc8dc.png)
Why the `networkKey` is set to `456`? `456` is not defined in the `constants.js`.

yjkimjunior:
^ yes, also addressing this with the udpated genesis hash, and upating the genesis hash used in the testqr payloads as well.

PR #340, opened 2019-08-28
Fix Signing Substrate Payload, Better UX for Errors

yjkimjunior:
#339

also added the ios production build script to package.json


src/util/db.js
```diff
@@ -54,12 +54,14 @@ function txKey(hash) {
 export const deleteAccount = async account =>
   SecureStorage.deleteItem(accountId(account), accountsStore);
 
-export const saveAccount = account =>
+export const saveAccount = account => {
+  console.log('account -> ', account);
```
```suggestion
```

yjkimjunior:
Hmm I will take a look

src/util/account.js
```diff
@@ -4,14 +4,16 @@ export function accountId({
...
-  const { ethereumChainId, protocol, genesisHash } = NETWORK_LIST[networkKey];
-
-  if (typeof address !== 'string' || address.length === 0 || !networkKey) {
-    throw new Error(`Couldn't create an accountId, address or networkKey missing`);
+  if (typeof address !== 'string' || address.length === 0 || !networkKey || !NETWORK_LIST[networkKey]) {
+    throw new Error(`Couldn't create an accountId. Address or networkKey missing, or network key was invalid.`);
   }
 
-  if (protocol === NetworkProtocols.SUBSTRATE){ 
+  const { ethereumChainId='', protocol, genesisHash } = NETWORK_LIST[networkKey];
```
hanwencheng: Here I guess `genesisHash` may also be undefined for Ethereum transactions.

Tbaut: indeed, since it's not used anyway it doesn't crash, but it should probably be initialized as well.

Tbaut: addressed in https://github.com/paritytech/parity-signer/pull/346

PR #349, opened 2019-09-02
onScan find Address in Store with Appropriate Prefix

### LINKED ISSUE #348: Address not found on 3.0
After requesting a signature on a "Polkadot Dev" account, the . following appears -

ima_0d1527f

- the displayed account is not encoded with prefix 42 (the address shows in 42 on the UI)
- when actually decoding the address displayed and re-coding it with the correct prefix, the 2 matches - so it would seem the lookup is incorrect

yjkimjunior:
fixes #348

This will match the account prefix based on the payload genesis hash.

In case user is signing a message or a hash, and the address provided couldn't be found in store, it'll do one more pass with the dev chain prefix encoding to search.

### DIFF
--- a/src/stores/ScannerStore.js
+++ b/src/stores/ScannerStore.js
@@ -17,6 +17,7 @@
 // @flow
 import { GenericExtrinsicPayload } from '@polkadot/types';
 import { hexStripPrefix, isU8a, u8aToHex } from '@polkadot/util';
+import { checkAddress, decodeAddress, encodeAddress  } from '@polkadot/util-crypto';
 import { Container } from 'unstated';
 
 import { NETWORK_LIST, NetworkProtocols } from '../constants';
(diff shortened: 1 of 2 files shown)


src/stores/ScannerStore.js
```diff
@@ -81,6 +82,24 @@ export default class ScannerStore extends Container<ScannerState> {
   async setParsedData(strippedData, accountsStore) {
     const parsedData = await constructDataFromBytes(strippedData);
 
+    const chainName = checkAddress(parsedData.data.account, 2)[0] ? 'Kusama' : checkAddress(parsedData.data.account, 42)[0] ? 'Polkadot' : null;
```
Tbaut: I'm worried that this logic and what follows isn't generic and needs to be extended as soon as there's a new network.
Can't we use `NETWORK_LIST[genesisHash].prefix` ?

yjkimjunior: I agree it's not ideal, but updates will be needed when there's a new network regardless.

By the way this is for the payloads without a genesisHash such as signing a message or a hash.

src/util/decoders.js
```diff
@@ -143,38 +144,54 @@ export async function constructDataFromBytes(bytes) {
...
+              network = NETWORK_LIST[extrinsicPayload.genesisHash.toHex()];
+              data['data']['account'] = encodeAddress(publicKeyAsBytes, network.prefix); // encode to the prefix;
+
               break;
             case '01': // data is a hash
               data['action'] = 'signData';
               data['oversized'] = false;
               data['isHash'] = true;
               data['data']['data'] = rawPayload;
+              data['data']['account'] = encodeAddress(publicKeyAsBytes, 2); // default to Kusama
```
Tbaut: can we use the prefix here as well rather than have it hardcoded directly.
This suggestion needs to import `SubstrateNetworkKeys` (also double check it, just typed it here)
```suggestion
              const kusamaNetworkKey = SubstrateNetworkKeys.KUSAMA;
              const kusamaPrefix = NETWORK_LIST[kusamaNetworkKey].prefix;
              data['data']['account'] = encodeAddress(publicKeyAsBytes, kusamaPrefix); // default to Kusama
```

src/util/decoders.js
```diff
@@ -183,6 +200,7 @@ export async function constructDataFromBytes(bytes) {
               data['data']['data'] = isOversized
                 ? await blake2s(u8aToHex(rawPayload))
                 : u8aToString(rawPayload);
+              data['data']['account'] = encodeAddress(publicKeyAsBytes, 2); // default to Kusama
```
Tbaut: here as well, (would need a `const` earlier in the code, and just used)

Issue #350, opened 2019-09-02
v2 and v3 don't have the same package name


Tbaut:
when installing v3 (apk) after a v2.2 (from the store)
I have 2 different apps with 2 different Bundle names:
![image](https://user-images.githubusercontent.com/33178835/64115942-3e425f80-cd91-11e9-8d4b-7694490d8061.png)
![image](https://user-images.githubusercontent.com/33178835/64115950-439faa00-cd91-11e9-9c4d-1b98bc24bf4f.png)
for v2, the name is io.parity.signer.
This must be a side effect of https://github.com/paritytech/parity-signer/pull/300/

The name **must** stay the same for the android store.

PR #351, opened 2019-09-02
Make release flag implicit

Tbaut:
closes : https://github.com/paritytech/parity-signer/issues/350

tested:
- build an apk and add to the google playstore console --> accepted (unlike before).
- install with `yarn start` and `yarn android` --> can do screenshots.
- install release apk --> can't do screenshots.


hanwencheng:
Thanks!

PR #352, opened 2019-09-02
Migrate v2

### LINKED ISSUE #347: Accounts missing after 3.0
After the 3.0 upgrade my Ethereum accounts disappeared.

On 3.0.1 each time I open the app, it has lost my previously added substrate accounts.

Tbaut:
- closes #347 
- Will migrate v2 accounts (ethereum accounts until v2.2) at first load, without deleting them in the old store.
- bump Terms and Conditions so that you will get older accounts migrated (as well as being shown the ToCs) for sure at first launch.
- refactor the whole thing.. still not 100% sure what the v1 migration is really doing..


android/app/build.gradle
```diff
@@ -144,13 +144,9 @@ android {
     }
     buildTypes {
         release {
-            applicationIdSuffix "release"
```
Tbaut: Removed this so that the v3 is being installed on top of v2 and not a separated app on android

Tbaut:
The way I did it (a little annoying):
- checkout an old version git checkout c9b6451666628edacf36484b2c8dcc3e6e07d9ba <-- that's the commit for v2.2.2
- launch and create an account
- checkout this branch `git checkout tbaut-migrave-v2`
- launch and see if you see the account (without warning of unhandled promise rejection)


This is on android. I tested the following without success:

- get v2.2.2 from the store and create an account
- install master in debug --> can't install it, another app with the same name exist
- try to install an apk of v3.0.2 --> can't install, same name but not the same signature (the store signs it differently it seems)

src/util/db.js
```diff
@@ -20,27 +20,25 @@ import { AsyncStorage } from 'react-native';
...
-  SecureStorage.deleteItem(account.address, accountsStore_v1);
-
-export async function loadAccounts_v1() {
+export async function loadAccounts( version = 3 ) {
   if (!SecureStorage) {
     return Promise.resolve([]);
   }
 
-  return SecureStorage.getAllItems(accountsStore_v1).then(accounts =>
+  const accountStoreVersion = version == 1 ? 'accounts' : `accounts_v${version}`
```
```suggestion
  const accountStoreVersion = version === 1 ? 'accounts' : `accounts_v${version}`
```

src/util/db.js
```diff
@@ -112,10 +100,10 @@ async function storagePushValue(key, value) {
...
 
 export async function loadToCAndPPConfirmation() {
-  const result = await AsyncStorage.getItem('ToCAndPPConfirmation');
+  const result = await AsyncStorage.getItem('ToCAndPPConfirmation_v3');
   return !!result;
 }
 
 export async function saveToCAndPPConfirmation() {
-  await AsyncStorage.setItem('ToCAndPPConfirmation', 'yes');
+  await AsyncStorage.setItem('ToCAndPPConfirmation_v3', 'yes');
```
yjkimjunior: why is this called ToC btw? I don't think we need to change since we all know it means Terms and Conditions but just curious

src/util/db.js
```diff
@@ -20,27 +20,25 @@ import { AsyncStorage } from 'react-native';
...
+
+  return SecureStorage.getAllItems(accountsStore).then(accounts =>
     Object.values(accounts).map(account => JSON.parse(account))
   );
 }
 
 const accountsStore = {
-  keychainService: 'accounts_v2',
-  sharedPreferencesName: 'accounts_v2'
+  keychainService: "accounts_v3",
```
yjkimjunior: single quotes

src/screens/Loading.js
```diff
@@ -63,10 +55,33 @@ export default class Loading extends React.PureComponent {
...
+    
+    await loadAccounts()
+    this.props.navigation.dispatch(tocActions);
+  }
+
+  async migrateAccount_v1 () {
+    const oldAccounts_v1 = await loadAccounts(1);
+    const accounts = oldAccounts_v1.map(empty).map(a => ({ ...a, v1recov: true }));
+    accounts.forEach(saveAccount);
+  }
```
yjkimjunior: so this loads v1 accounts from secure storage, resets all the values to something false-y, and adds a new field `v1recov` to set to `true` the saves them back to v2 secureStorage? why does it map `empty`, shouldn't it save it into `v2` storage as is?

src/screens/Loading.js
```diff
@@ -63,10 +55,33 @@ export default class Loading extends React.PureComponent {
...
+
+  // only ethereum account with chainId and networkType properties
+  async migrateAccount_v2 () {
+    const oldAccounts_v2 = await loadAccounts(2);
+    const accounts = oldAccounts_v2.map(empty).map(a => {
+      let result = a
+      if (a.chainId) {
+        result = { ...a, networkKey: a.chainId, v2recov: true };
```
yjkimjunior: result is already pointing to a, `result = { networkKey: result.chainId, v2recov: true, ...result }` would be more readable, or just keep this but set result to `{}` initially.

src/util/db.js
Tbaut: It's the flag to know if users have accepted the ToCs and our privacy policies (PP) and that will either tell the app to show the Account list view, or the ToCs screen.
So bumping it here allows us to :
- show the ToCs and PP screen again, even if users have accepted the v2 of those ToCs and PPs
- migrate v2 accounts if any

src/screens/Loading.js
Tbaut: Just in case you hadn't seen, I copied this code from before.
>  why does it map empty

not sure either.. but my guess is just to make sure we initialize all the new values.

> shouldn't it save it into v2 storage as is?

I can't tell what's the diff between v1 and v2.. so I can't tell what would make sense or not. We would need to test an older version. I'll do that.

src/util/db.js
yjkimjunior: yes, but..shouldn't it be `TaC` or `TnC`?

Tbaut: didn't even realize it. I think it's an app wide typo!

src/screens/Loading.js
```diff
@@ -63,10 +53,32 @@ export default class Loading extends React.PureComponent {
...
+    // v2 accounts are only ethereum account 
+    // with deprectaded `chainId` and `networkType: 'ethereum'` properties
+    // no networkKey property
+    const oldAccounts_v2 = await loadAccounts(2);
+    const oldAccounts = [...oldAccounts_v1, ...oldAccounts_v2]
+
+    const accounts = oldAccounts.map(a => {
+      let result = {}
+      if (a.chainId) {
+        result = { ...a, networkKey: a.chainId, recovery: true };
```
Tbaut:
```suggestion
        result = { ...a, networkKey: a.chainId, recovered: true };
```

Tbaut:
- reworked the migration:
  - not using `empty` any more
  - migrating v1 similarly as v2

- changed the `empty` function to only return an empty account.

src/screens/Loading.js
```diff
@@ -63,10 +53,32 @@ export default class Loading extends React.PureComponent {
...
+    const accounts = oldAccounts.map(a => {
+      let result = {}
+      if (a.chainId) {
+        result = { ...a, networkKey: a.chainId, recovered: true };
+        delete result.chainId;
+        delete result.networkType;
+      }
+      return result
+    })
+    accounts.forEach(saveAccount);
```
hanwencheng:
```suggestion
    accounts.forEach(account => {
      if(account.networkKey && NETWORK_LIST[account.networkKey]){
        saveAccount(account)
      }
    });
```
This prevent a throw-out error and reduce one warning. In my previous account list I have some account with `networkKey:456`, after I add these code, more accounts are shown.

Tbaut: > In my previous account list I have some account with networkKey:456

Right but this is came from master, not a release so I don't think this is a case I want to handle.

hanwencheng: Then I suggest we handle the accountID parsing error here. This is the entry point for account list initialization, all the account processing errors will firstly be catched here.
```suggestion
    accounts.forEach(account => {
      try{
        saveAccount(account);
      } catch(e){
        console.error(e);
      }
    });
```

hanwencheng:
LGTM, leave comments on one point here. Also one question: is there anyway to wipe out the previous account data on iOS devices?

Tbaut:
addressed comments, tested again with instructions given in https://github.com/paritytech/parity-signer/pull/352#issuecomment-527437669 and it's looking good to review again :)

PR #353, opened 2019-09-03
feat: add ECCrypto native module with hardware-back key storage / secure enclave .

### LINKED ISSUE #222: Secure Enclave to encrypt accounts on iOS
Following #217. We cannot use the private keys or the encryption scheme for signing transactions, but we can use the private key generated by the enclave to encrypt and decrypt the account keystore files.

Use https://github.com/paritytech/react-native-secure-storage

hanwencheng:
This is the first PR for #222 According to [Parity-Signer Secure Key Storage Proposal](https://gist.github.com/hanwencheng/736cc28834a06d9cbd4198c03a7c6033)

To avoid the PR become too large, and also to be available to switch to another high priority issues, I would rather split this PR into multiple parts.

As development goes by, the `react-native-ecc` is not 100% fits to our use case, I have implement an `ECCrypto` module which both works in iOS (with Secure Enclave) and Android (with Hard-ware backed Keystore).

So in this PR the following parts are finished:

- [x] iOS Encryption and decryption function.
- [x] Android Encryption and decryption function.
- [x] Integrate with iOS touchID.
- [ ] ~~Integrate with Android Fingerprint.~~
- [x] Add Integration Test for the encryption and decryption bridge

Rest parts are:

- [ ] Refactor Account to store the encrypted account data with secure storage
- [ ] Save the hashed pin in to system normal storage.

I am using Google [Tink](https://github.com/google/tink) library on Android for simplify the code and workflow but currently the BiometricPrompt is not integrated yet, there is an [issue](https://github.com/google/tink/issues/211#issue-445852940) discussing it.

### Usage

```javascript

import { ECCrypto } from 'NativeModules';

const encryptAndThenDecrypt = async () => {
  try {
    //Encrypt
    const cipherText = await ECCrypto.encrypt({
      data: 'I am a seed data',
      label: '0x5Cc5dc62be3c95C771C142C2e30358B398265deF' //any identical string, e.g. public address 
    });
    //Decrypt
    const clearText = await ECCrypto.decrypt({
      data: cipherText,
      label: '0x5Cc5dc62be3c95C771C142C2e30358B398265deF' //the same identical string
    });
    console.log('decrypt result is ', clearText);
  } catch(e) {
    console.log(e);
  }
};
```

------ORIGINAL DESCRIPTION BELOW------

[React-native-ecc](https://github.com/tradle/react-native-ecc/) has implement secure enclave with ecc sign and verify function. We could use to replace `react-native-secure-storage`, I am currently creating an [PR](https://github.com/tradle/react-native-ecc/pull/6) to add the encryption and decryption function to fulfill our goal.

On the Android system we will also need to store the encrypted data into Keystore so I will also implement the encryption and decryption function on Android based on React-native-ecc as well.

### DIFF
--- a/e2e/testIDs.js
+++ b/e2e/testIDs.js
@@ -3,11 +3,17 @@ const testIDs = {
 		tacView: 'tac_view',
 		agreeTacButton: 'tac_agree',
 		agreePrivacyButton: 'tac_privacy',
+		nativeModuleTestButton: 'tac_native_test',
 		nextButton: 'tac_next'
 	},
 	AccountListScreen: {
-		accountList: 'accountList',
-	}
+		accountList: 'accountList'
+	},
+	NativeTestScreen: {
+		nativeTestView: 'native_test_view',
+		startButton: 'native_test_start',
+		succeedView: 'native_test_success'
+	},
 };
 
 export default testIDs;
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -179,6 +179,8 @@ dependencies {
     implementation 'androidx.appcompat:appcompat:1.0.0'
     implementation 'com.facebook.react:react-native:0.60.3'
 
+    compile 'com.google.crypto.tink:tink-android:1.3.0-rc1'
+
     if (useIntlJsc) {
         implementation 'org.webkit:android-jsc-intl:+'
     } else {
--- a/android/app/src/main/java/io/parity/signer/ECCryptoModule.java
+++ b/android/app/src/main/java/io/parity/signer/ECCryptoModule.java
@@ -1,0 +1,120 @@
+// Copyright 2015-2019 Parity Technologies (UK) Ltd.
+// This file is part of Parity.
+
+// Parity is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Parity is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+package io.parity.signer;
+
+import android.os.Build;
+import android.util.Base64;
+import android.util.Log;
+import com.facebook.react.bridge.*;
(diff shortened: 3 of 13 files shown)


Tbaut:
Thanks for the PR. I'm not entirely sure how to review this PR, or why it has the "grumble" label.

>So the main roadmap would be:
- [x] iOS Encryption and decryption function.
- [x] Android Encryption and decryption function.
- [x] Integrate with iOS touchID.
- [ ] Integrate with Android Fingerprint.
- [ ] Add Unit Test for the encryption and decryption function
- [ ] Refactor Account to store the encrypted account data with secure storage
- [ ] Save the hashed pin in to system normal storage.

This should rather live in an issue IMO. It can surely be split into several PRs, I see mainly 2, with the first 5 elements (up to the unit tests) and the last 2 elements with the integration. What do you think?

hanwencheng:
The "grumble" label is because according to the previous talk with @sjeohp, and his PR on #416 , we have bunch of stuff to be merged.

For better review, I will create a test for the PR.

And I agree with you to move the tasks in to the issue.

sjeohp:
@hanwencheng I think we can merge with master separately and it won't be a problem.

hanwencheng:
Native test part is finished, will refresh this PR after the CI integration merged, then we will have a clear view of it.

hanwencheng:
Integration test is added as [here](http://github.com/paritytech/parity-signer/pull/353/files#diff-631404afaba023d63642e27678c82257R23-R35)

e2e/testIDs.js
```diff
@@ -3,11 +3,17 @@ const testIDs = {
...
 		agreePrivacyButton: 'tac_privacy',
+		nativeModuleTestButton: 'tac_native_test',
 		nextButton: 'tac_next'
 	},
 	AccountListScreen: {
-		accountList: 'accountList',
-	}
+		accountList: 'accountList'
+	},
+	NativeTestScreen: {
```
sjeohp: Is this just a dummy-UI for testing non-UI features?

sjeohp:
I'm not sure it makes sense to use a UI testing framework to test non-UI features.  Surely encryption methods and such can be tested more easily without it...

e2e/testIDs.js
hanwencheng: yeah, it is a dummy UI only display on the development environment. 

I have not an expert to do the testing on Native modules. I thought about how to test about them, the reason I choose to use E2E for testing it is:
1. a central place for testing all the native modules together.
2. one command could run the test without mouse interactions.

If there is a native way to fulfill these goals, then I am happy change to that :)

sjeohp:
Happy to merge once the conflict is resolved

PR #357, opened 2019-09-04
chore: show tab correctly

hanwencheng:
As we are using "tabs" instead of "spaces" on Javascript file, Github show the indentation of tab always as 4 spaces, it will make confusion when we check the diff on Github, especially for #337.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,16 +1,18 @@
 root = true
+
 [*]
-indent_style=space
-indent_size=2
-tab_width=2
-end_of_line=lf
-charset=utf-8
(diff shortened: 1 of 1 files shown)


hanwencheng:
If you scroll down to check the diffs on Github here: https://github.com/paritytech/parity-signer/pull/337/files . The indentations are quite confusion because Github default display of tab for  4 or 8 spaces. 

According [here](https://stackoverflow.com/questions/8833953/how-to-change-tab-size-on-github)
>   When you have a [.editorconfig](https://editorconfig.org/) in your repository it will respect it when viewing code on GitHub.

PR #358, opened 2019-09-06
fix: Remove substrate dev

Tbaut:
As discussed, this will not be useful in the released version.

**MERGE before release only**


Tbaut:
Should probably add a proper check for `_DEV_` similar to https://github.com/paritytech/parity-signer/pull/365/files#diff-82ce97d8934ca42f2bca37e403683393R101

I'll log that.

Issue #158, opened 2018-10-09
Add ability to change the network on account edition screen


maciejhirsz:
This is necessary for legacy accounts that are all imported as mainnet Ethereum, and some edge cases such as claiming testnet ETH using the same address on different networks.

fckt:
For now user can just create an another account (through "Recover Account") with the same address, but for a different network.. But I'm agree we should add an ability to choose a network for the existing account.

Slesarew:
We need to balance convenience with friction here; this requires proper research

Slesarew:
Ok, looks like it **has** to be difficult.

PR #367, opened 2019-09-09
fix: clear account state when changing network

### LINKED ISSUE #363: Copy behavior inconsistently for Substrate and Ethereum accounts
On account backup screen, we can copy the seed phrases by long pressing on it.

But this currently only works on Substrate accounts, for Ethereum accounts it would be an empty string.

Probably we should disable this copy feature by the release version. It would be safe if we force user to write the seed phrase down instead of copying it. 

Related code here: https://github.com/paritytech/parity-signer/blob/master/src/screens/AccountBackup.js#L111

Screenshot 2019-09-06 at 17 09 55

hanwencheng:
close #363 .

1. Clear the state when choosing the new network. (Which may cause more bugs if user switch network between Substrate and Ethereum).
2. Copy the seed when creating an Ethereum account in DEV mode. 

Fixed behaviour here:

out

### DIFF
--- a/src/screens/AccountBackup.js
+++ b/src/screens/AccountBackup.js
@@ -28,7 +28,7 @@ import Button from '../components/Button';
 import TouchableItem from '../components/TouchableItem';
 import DerivationPasswordVerify from '../components/DerivationPasswordVerify';
 import AccountsStore from '../stores/AccountsStore';
-import {NETWORK_LIST, NetworkProtocols} from "../constants";
+import { NetworkProtocols, NETWORK_LIST } from '../constants';
 
 
 export default class AccountBackup extends React.PureComponent {
--- a/src/screens/AccountBackup.js
+++ b/src/screens/AccountBackup.js
@@ -28,6 +28,7 @@ import Button from '../components/Button';
 import TouchableItem from '../components/TouchableItem';
 import DerivationPasswordVerify from '../components/DerivationPasswordVerify';
 import AccountsStore from '../stores/AccountsStore';
+import {NETWORK_LIST, NetworkProtocols} from "../constants";
 
 
 export default class AccountBackup extends React.PureComponent {
@@ -75,7 +76,8 @@ class AccountBackupView extends React.PureComponent {
     const {navigate} = navigation;
     const isNew = navigation.getParam('isNew');
     const {address, derivationPassword, derivationPath, name, networkKey, seed, seedPhrase} = isNew ? accounts.getNew() : accounts.getSelected();
-    
+    const {protocol} = NETWORK_LIST[networkKey];
+
     return (
       <ScrollView
         style={styles.body}
@@ -108,7 +110,11 @@ class AccountBackupView extends React.PureComponent {
                     text: 'Copy anyway',
                     style: 'default',
                     onPress: () => {
-                      Clipboard.setString(`${seedPhrase}${derivationPath}`);
+                      if (protocol === NetworkProtocols.SUBSTRATE) {
+                        Clipboard.setString(`${seedPhrase}${derivationPath}`);
+                      } else {
+                        Clipboard.setString(seed)
+                      }
                     }
                   },
                   {
@@ -117,7 +123,7 @@ class AccountBackupView extends React.PureComponent {
                   }
                 ]
               );
-            } 
+            }
           }}
         >
           <Text style={styles.seedText}>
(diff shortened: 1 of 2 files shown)


src/screens/AccountBackup.js
```diff
@@ -28,6 +28,7 @@ import Button from '../components/Button';
 import TouchableItem from '../components/TouchableItem';
 import DerivationPasswordVerify from '../components/DerivationPasswordVerify';
 import AccountsStore from '../stores/AccountsStore';
+import {NETWORK_LIST, NetworkProtocols} from "../constants";
```
yjkimjunior: single quote

src/screens/AccountNetworkChooser.js
```diff
@@ -24,6 +24,7 @@ import fonts from "../fonts";
 import TouchableItem from '../components/TouchableItem';
 import { NETWORK_LIST } from '../constants';
 import AccountsStore from '../stores/AccountsStore';
+import { empty } from "../util/account";
```
yjkimjunior: single quote

yjkimjunior:
@Tbaut sorry but I'm quite confused by what you mean above,

It seems this PR is just making the clippy copy the appropriate seed for the network that is chosen, and that's it.

> Rather than having yet another if (isSubstrate) I would actually advise to remove the if here and handle Ethereum accounts just like Substrate accounts.

how? seed is separate from seedPhrase and derivationPath so they have to be treated separately.

>  If things are set correctly, the derivation path and password will be set to empty string for Ethereum so that you don't have to change anything else

FWICT that was the source of the bug being fixed here, where long pressing the seed for Ethereum would copy an empty string.

> No need to empty anything with the networkChooser either

this seems to have been the other bug, where the wrong seed for the wrong network would be copied...

>  I think it's better to isolate each component and the Network chooser should not have to deal with side effect happening elsewhere.

? what is this referring to?

sorry if my quetsions don't make sense, but like i said, i'm quite confused here

Tbaut:
This PR does solve the problem, I just think there's a better way to do so.
I propose to use have the same comportment between Ethereum and Substrate accounts. So I propose to have a `seedPhrase` for Ethereum accounts as well so that we don't have to add conditions for Ethereum or Substrate. The `DerivationPath` and DerivationPassword` will likely be empty for Ethereum until we implement the feature.

> how? seed is separate from seedPhrase and derivationPath so they have to be treated separately.

Sure I agree.

> this seems to have been the other bug, where the wrong seed for the wrong network would be copied...

This is a side effect that doesn't happen with my suggestion.

> I think it's better to isolate each component and the Network chooser should not have to deal with side effect happening elsewhere.

I don't see why we would need to totally empty an account if we change the network. This logic should not need to be there as it's a fix for something else. If we want to reuse this component for e.g https://github.com/paritytech/parity-signer/issues/158 this logic would be in the way.

Happy to do a PR to show you what I mean. I just think there's a more elegant way to fix this bug.

hanwencheng:
Thanks for the Q&A session, which make things clear.

> In the near future both will have derivation paths. 

If that is happened (and also `seedPhrase`) I think it is would make sense to use a uniform data structure for Substrate and Ethreum accounts.

Another question, after I check the difference of:
- `seed`:  `${phrase}${derivePath}///${password}` , see [code](https://github.com/paritytech/parity-signer/blob/master/src/util/suri.js#L83)
- Copied seed clipboard Here: `${seedPhrase}${derivationPath}` , see [code](https://github.com/paritytech/parity-signer/blob/master/src/screens/AccountBackup.js#L111)

Is there a special reason that we do not include `password` in the seed for Substrate account?

Tbaut:
> The only difference is whether password is included, is there a special reason that we do not include password in the account backup for Substrate account?

The idea is that the password should remain secret, so we give a way to test it, without ever revealing it. A malicious actor with the seed phrase can't get access to the fund. A malicious actor with the seed and the derivation path can still not get access to the fund since they need the password on top of that.

> - seed: ${phrase}${derivePath}///${password} , see code

This is one of the things that broke with my suggestion, because `brainwalletSign` ([here](https://github.com/paritytech/parity-signer/blob/master/src/util/native.js#L79)) doesn't support derivation path (and isn't happy with those "///" at the end of most seed phrases). Even after making this changes the signing was still broken, not sure where it comes from. Here's the suggestion I had: https://github.com/paritytech/parity-signer/compare/tbaut-suggestion?expand=1

PR #372, opened 2019-09-09
shows the assets on iOS

### LINKED ISSUE #364: Missing Parity logo on iphone
I just realized from a screenshot given by @hanwencheng that there's no Parity Signer logo on iphone (at the top left of the account list), and no back arrow on any other screen:
image
image

whereas on Android (it's normal to have no title on Android):
image

On iphone, the v2 used to have both the arrow and the title:
image

hanwencheng:
close #364 

If any new static files like images and  is required in the JS codebase, if order to link these static files to bundle for iOS devices, we need run `yarn build-prod:ios` in order to generate the updated assets folder.

**Reason**:
Images cannot be find in the release bundle, need to:
1. extract the required static assets into `ios/assets` folder.
2. add the references to `ios/assets` folder in Xcode


ios/NativeSigner.xcodeproj/project.pbxproj
```diff
@@ -1746,7 +1724,7 @@
 				CLANG_WARN_SUSPICIOUS_MOVE = YES;
 				CLANG_WARN_UNREACHABLE_CODE = YES;
 				CLANG_WARN__DUPLICATE_METHOD_MATCH = YES;
-				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
+				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer: hanwen@parity.io (Y66P73ZL35)";
```
Tbaut: Not sure this should stay there

ios/NativeSigner/AppDelegate.m
```diff
@@ -34,7 +34,7 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
 - (NSURL *)sourceURLForBridge:(RCTBridge *)bridge
 {
 #if DEBUG
-  return [[RCTBundleURLProvider sharedSettings] jsBundleURLForBundleRoot:@"index" fallbackResource:nil];
+  return [NSURL URLWithString:@"http://192.168.1.220:8081/index.ios.bundle?platform=ios&dev=true"];
```
yjkimjunior: I'm not able to get my metro bundler to connect and run, and I suspect this is the issue...Ran the build-prod:ios script after a cleaning cache as well. @hanwencheng

yjkimjunior: just did a clean build reverting this line and it works :)

hanwencheng: My fault, this line cause problem, and should not be here.

ios/NativeSigner.xcodeproj/project.pbxproj
hanwencheng: good catch, this should be removed here, but use the automatic code signing

PR #377, opened 2019-09-10
align height in Account new screen

### LINKED ISSUE #373: Height not aligned with address text and "select icon"
In the account create screen.

Height not aligned with address text and "select icon" , so the UI also changes.

out mp4

hanwencheng:
close #373 

Self explained.

### DIFF
--- a/src/components/AccountCard.js
+++ b/src/components/AccountCard.js
@@ -44,7 +44,7 @@ export default class AccountCard extends React.PureComponent {
   };
 
   render() {
-    const { address, networkKey, onPress, seedType, shortAddress = false, style } = this.props;
+    const { address, networkKey, onPress, seedType, style } = this.props;
     let { title } = this.props;
     title = title.length ? title : AccountCard.defaultProps.title;
     const seedTypeDisplay = seedType || '';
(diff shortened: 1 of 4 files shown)


Tbaut:
The problem with a bigger font size is that the address isn't fully visible on devices with a lower resolution (since we target older devices I'll likely be the case).

Tbaut:
Here's what I see (there's still a little jump but that's not a big deal)
address-cut

hanwencheng:
Will refactor it according to #296

hanwencheng:
I use similar way as proposed in #296 to show the address, one update in 8d77cda is that we use system's [ellipsesizeMode](https://facebook.github.io/react-native/docs/text.html#ellipsizemode) to automatically shrink the address. with more information displayed,  I think this could replace our hardcoded string ellipses [function](https://github.com/paritytech/parity-signer/blob/master/src/components/Address.js#L34-L36) @Tbaut.

Here is the result images on iOS and Android:
out

One more thing I noticed that there is a small jump when we refresh the list of icons on Ethereum Network.

Tbaut:
This is great, didn't know about it, it perfectly fits our needs. I'll review it tomorrow.

hanwencheng:
ed56b0d resolve the icon refresh jump problem mentioned above.

hanwencheng:
This PR is ready for the review now. @Tbaut @yjkimjunior

PR #379, opened 2019-09-10
fix: add pin check

### LINKED ISSUE #374: Prevent user use decimal in Pincode
Current we use `numeric` prop for TextInput, but the display on Android and iOS devices are [different](https://lefkowitz.me/visual-guide-to-react-native-textinput-keyboardtype-options/)

Standard comparison is as following, but user may also use custom input application, though on my Xiaomi there is no decimal, I think we still need to use regular expression to dismiss the non-number input.

Screenshot 2019-09-09 at 16 27 29
 
out mp4

hanwencheng:
close #374 .

Also clean one place in `AccountNetworkChooser.js` to make the code clear.

### DIFF
--- a/src/screens/AccountNetworkChooser.js
+++ b/src/screens/AccountNetworkChooser.js
@@ -26,7 +26,6 @@ import { NETWORK_LIST } from '../constants';
 import AccountsStore from '../stores/AccountsStore';
 import { empty } from '../util/account';
 
-
 export default class AccountNetworkChooser extends React.PureComponent {
   static navigationOptions = {
     title: 'Choose a network',
--- a/src/screens/AccountNetworkChooser.js
+++ b/src/screens/AccountNetworkChooser.js
@@ -26,6 +26,7 @@ import { NETWORK_LIST } from '../constants';
 import AccountsStore from '../stores/AccountsStore';
 import { empty } from '../util/account';
 
+
 export default class AccountNetworkChooser extends React.PureComponent {
   static navigationOptions = {
     title: 'Choose a network',
@@ -61,7 +62,7 @@ class AccountNetworkChooserView extends React.PureComponent {
               }
             ]}
             onPress={() => {
-              accounts.updateNew({ ...empty('', networkKey) });
+              accounts.updateNew(empty('', networkKey));
               navigation.goBack();
             }}
           >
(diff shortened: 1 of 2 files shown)


src/screens/AccountPin.js
```diff
@@ -96,6 +96,12 @@ class AccountPinView extends React.PureComponent {
     return (<Text style={styles.hintText}>Choose a PIN code with 6 or more digits</Text>)
   }
 
+  onPinInputChange = stateName => pinInput => {
```
Tbaut: This is a weird notation that I'm not used to.
Any way to make it a little easier to read?

src/screens/AccountPin.js
```diff
@@ -111,14 +117,14 @@ class AccountPinView extends React.PureComponent {
...
-          onChangeText={pin => this.setState({ pin: pin, pinMismatch: false, pinTooShort: false })}
+          onChangeText={this.onPinInputChange('pin')}
           value={this.state.pin}
         />
         <Text style={styles.title}>CONFIRM PIN</Text>
         <PinInput
           returnKeyType="done"
           focus={this.state.focusConfirmation}
-          onChangeText={confirmation => this.setState({ confirmation: confirmation, pinMismatch: false, pinTooShort: false })}
+          onChangeText={this.onPinInputChange('confirmation')}
```
yjkimjunior: agreed with @Tbaut above, why not have a separate handler for pin and confirmation rather than trying to make it general within `onPinInputChange` ?

src/screens/AccountNetworkChooser.js
```diff
@@ -26,6 +26,7 @@ import { NETWORK_LIST } from '../constants';
 import AccountsStore from '../stores/AccountsStore';
 import { empty } from '../util/account';
 
+
```
yjkimjunior:
```suggestion
```

hanwencheng:
It is a curry function, some interesting playground is [here](https://github.com/learn-javascript-courses/es6-curry).

It is practical in handle event listeners, but sure I can change it for the readability,.

hanwencheng:
It is a curry function, some interesting playground is [here](https://github.com/learn-javascript-courses/es6-curry).

It is practical in handle event listeners, I think it would be a little bit redundant if I change to two separate functions, please check out 80f43fe , any better ideas?

PR #380, opened 2019-09-10
fix #371: Format Balance

### LINKED ISSUE #371: Format Balance & Number on TxDetails for Kusama Payloads
should be able to just iter over the params, if param type is Balance then format as such, if param type is BN, number, uint, ..., etc. format as number.

yjkimjunior:
#371


src/util/units.spec.js
```diff
@@ -16,20 +16,67 @@
...
-  it('should return BigNumber for undefined values', () => {
-    expect(fromWei(null)).toEqual('0');
-    expect(fromWei(undefined)).toEqual('0');
-    expect(fromWei(0)).toEqual('0');
-    expect(fromWei('0')).toEqual('0');
-    expect(fromWei('')).toEqual('0');
+  describe('kusama', () => {
+    beforeAll(() => {
+      const metadata = new Metadata(kusamaData);
+
```
Tbaut:
```suggestion
```

yjkimjunior: ok I will just use `formatBalance` and do the same thing apps does then. The only thing is the "international" friendly space instead of commas, which I can make a PR to add that option for in the `common` library, but I don't think there's any international standard for that, so no matter which we choose, some people will be less than pleased

Tbaut: Sure, what I meant by international is that we shouldn't use comma as a thousand separator, anything else would be good.

yjkimjunior: cool, thankfully Jaco has already thought of this in his formatBalance method :)

yjkimjunior:
| Input:      | Input: 
| IMG_0031  | IMG_0030

PR #388, opened 2019-09-11
Update readme for Kusama screenshots and tutorials

Tbaut:
self explanatory


README.md
```diff
@@ -8,20 +8,20 @@
 
 Parity Signer is a mobile application that allows any smartphone to act as an air-gapped crypto wallet. This is also known as "cold storage".
 
-Once the app has been installed, users can create Kusama or Ethereum accounts, sign messages/transactions, and transfer funds to and from these accounts without any sort of connectivity enabled on the device.
+You can create Kusama and Ethereum accounts, sign messages/transactions, and transfer funds to and from these accounts without any sort of connectivity enabled on the device.
 
-Users can opt to shut off and even physically remove the smartphone's Wifi, Mobile Network, and Bluetooth to ensure that the mobile phone containing these accounts will not be exposed to any online threat.
+You must turn off or even physically remove the smartphone's Wifi, Mobile Network, and Bluetooth to ensure that the mobile phone containing these accounts will not be exposed to any online threat.
```
hanwencheng: `must` might give user the impression that the app is only capable in an offline mode, suggest to use `should`

Tbaut: I think the phrase is clear, You must... to ensure that the mobile phone containing these accounts will not be exposed to any online threat.

If this phrase scares ppl that didn't want to follow this rule, then I'm actually happy they don't use the app :)

Tbaut:
Merging as time sensitive with the release and insubstantial changes.

PR #390, opened 2019-09-14
Signer should fallback to bytes when method cannot be decoded with the available metadata..

yjkimjunior:


yjkimjunior:
Okay this should be better now. 
IMG_0032

src/components/PayloadDetailsCard.js
```diff
@@ -115,25 +115,41 @@ function ExtrinsicPart({ label, fallback, value }) {
...
-      }
 
-      setArgNameValue(result);
-      setSectionMethod(`${sectionName}.${methodName}`);
+        setArgNameValue(result);
+        setSectionMethod(`${sectionName}.${methodName}`);
+      } catch (e) {
+        Alert.alert(
+          'Could not decode method with available metadata',
+          `Falling back to raw bytes. For your safety, we advise you decode this carefully yourself before signing and submitting the extrinsic.`,
```
Tbaut:
```suggestion
          `Not being able to decode what you are signing is inherently unsafe.A newer version of Parity Signer would be most likely able to decode this extrinsic. If you are not sure what we are talking about, open an issue on github.com/paritytech/parity-signer`,
```

yjkimjunior:
ok able to decode after registering the polkadot session key

https://github.com/polkadot-js/api/blob/a72fbece9628f6fef9dc02a2330fa53a4176fa3d/packages/api/src/base/Init.ts#L28

IMG_A8C31BCAE470-1

PR #392, opened 2019-09-17
feat: Show accounts with unknown network

### LINKED ISSUE #389: Mechanism to prevent the worse
If for some reason we are not able to read an account id, we don't handle the promise error and the whole UI is broken, no account is shown.

This happens when we change the data structure between 2 versions, don't do the migration properly, and users don't follow our (annoying) update recommendations. 
This would also happen if a genesis hash is changed for a network (e.g after https://github.com/paritytech/parity-signer/issues/302) in which case this problem would happen without even a bad move from our users. 

We have the case right now that Kusama will change genesis hash. If our users are lazy or hadn't created relevant accounts yet (just to test) they would still be trapped.

Suggestion: 
Catch this even, run the UI in a degraded mode for this account, with a warning saying that one account (flag it) can't be read correctly, it should be deleted and recovered : 
- list of accounts (show the name and address if we can)
- allow to see the recovery phrase (without password) after a pin check
- allow to delete the account (after pin check)

Tbaut:
closes #389 

How to test:
- create a Kusama account (1)
- modify the genesisHash aka `networkKey` of Kusama in `src/constants.js`
- reload the App

Because the  genesisHash used to create (1) isn't known any more, we can't match the Kusama Network. Still this account can be seen, its backup retrieved with the pin and deleted.

edit:
This was quite a painful PR, trying to remove many weird things that were done. Now we use the same [key, value] mapping in the accountStore and in the db. This allows to show accounts which we have no corresponding `NetworkKey`. It also helps to seemlessly save or delete accounts (even if the network is not known) because the db and the store have the same keys.
 
This PR introduces the notion of `selectedKey` that corresponds to a `selectedAccount` (more broadly named `selected` in the app). I think it's quite straightforward, we should use this key as much as possible and only use `accountId()` when dealing with Data coming in (QR code scanning) or saving new accounts.

I tested the migration but please feel free to do it as well ([as explained here](https://github.com/paritytech/parity-signer/pull/352#issuecomment-527437669))

image
image

### DIFF
--- a/src/screens/AccountDetails.js
+++ b/src/screens/AccountDetails.js
@@ -104,7 +104,7 @@ This account can only be recovered with its associated recovery phrase.`,
     return (
       <View style={styles.warningView}>
         <Text style={{...styles.title, ...styles.warningTitle}}>Warning</Text>
-        <Text style={styles.warning}>
+        <Text>
           This account wasn't retrieved successfully. This could be because its network isn't supported,
           or you upgraded Parity Signer without wiping your device and this account couldn't be migrated.
           {'\n'}{'\n'}
--- a/src/screens/AccountDetails.js
+++ b/src/screens/AccountDetails.js
@@ -56,11 +56,11 @@ class AccountDetailsView extends React.Component {
     super(props);
   }
 
-  componentDidMount() {
-    this.subscription = this.props.navigation.addListener('willFocus', t => {
-      this.props.txStore.loadTxsForAccount(this.props.accounts.getSelected());
-    });
-  }
+  // componentDidMount() {
+  //   this.subscription = this.props.navigation.addListener('willFocus', t => {
+  //     this.props.txStore.loadTxsForAccount(this.props.accounts.getSelected());
+  //   });
+  // }
 
   onDelete = () => {
     const accounts = this.props.accounts
@@ -92,9 +92,9 @@ This account can only be recovered with its associated recovery phrase.`,
     );
   }
 
-  componentWillUnmount() {
-    this.subscription.remove();
-  }
+  // componentWillUnmount() {
+  //   this.subscription.remove();
+  // }
 
   onOptionSelect = (value) => {
     const navigate = this.props.navigation.navigate
@@ -127,7 +127,6 @@ This account can only be recovered with its associated recovery phrase.`,
   }
 
   render() {
-    console.log('render account details');
     const account = this.props.accounts.getSelected();
 
     if (!account) {
--- a/src/screens/AccountDetails.js
+++ b/src/screens/AccountDetails.js
@@ -155,8 +155,8 @@ This account can only be recovered with its associated recovery phrase.`,
         <View style={styles.qr}>
           {
             protocol !== NetworkProtocols.UNKNOWN
-            ? <QrView data={selectedKey} />
-            : this.renderWarningUnknownAccount()
+              ? <QrView data={selectedKey} />
+              : this.renderWarningUnknownAccount()
           }
         </View>
       </ScrollView>
--- a/src/screens/AccountDetails.js
+++ b/src/screens/AccountDetails.js
@@ -18,6 +18,7 @@
 
 import React from 'react';
 import { Alert, ScrollView, StyleSheet, Text, View } from 'react-native';
+import { NavigationActions, StackActions } from 'react-navigation';
 import { Subscribe } from 'unstated';
 
 import colors from '../colors';
@@ -28,6 +29,7 @@ import AccountsStore from '../stores/AccountsStore';
 import TxStore from '../stores/TxStore';
 import { accountId } from '../util/account';
 import PopupMenu from '../components/PopupMenu'
+import { NETWORK_LIST, NetworkProtocols } from '../constants';
 
 export default class AccountDetails extends React.Component {
   static navigationOptions = {
@@ -67,7 +69,7 @@ class AccountDetailsView extends React.Component {
 
     Alert.alert(
       'Delete Account',
-      `Do you really want to delete ${selected.name || selected.address}?
+      `Do you really want to delete ${selected.name || selected.address || 'this account'}?
 This account can only be recovered with its associated recovery phrase.`,
       [
         {
@@ -75,7 +77,12 @@ This account can only be recovered with its associated recovery phrase.`,
           style: 'destructive',
           onPress: () => {
             accounts.deleteAccount(selected);
-            this.props.navigation.navigate('AccountList');
+            const resetAction = StackActions.reset({
+              index: 0,
+              key: undefined, // FIXME workaround for now, use SwitchNavigator later: https://github.com/react-navigation/react-navigation/issues/1127#issuecomment-295841343
+              actions: [NavigationActions.navigate({ routeName: 'AccountList' })]
+            });
+            this.props.navigation.dispatch(resetAction);
           }
         },
         {
@@ -103,12 +110,32 @@ This account can only be recovered with its associated recovery phrase.`,
     }
   }
 
+  renderWarningUnknownAccount = function () {
+    return (
+      <View style={styles.warningView}>
+        <Text style={{...styles.title, ...styles.warningTitle}}>Warning</Text>
+        <Text style={styles.warning}>
+          This account wasn't retrieved successfully. This could be because its network isn't supported,
+          or you upgraded Parity Signer without wiping your device and this account couldn't be migrated.
+          {'\n'}{'\n'}
+          To be able to use this account you need to:{'\n'}
+          - write down its recovery phrase{'\n'}
+          - delete it{'\n'}
+          - recover it{'\n'}
+        </Text>
+      </View>
+    )
+  }
+
   render() {
+    console.log('render account details');
     const account = this.props.accounts.getSelected();
 
     if (!account) {
       return null;
     }
+
+    const protocol = account.networkKey && NETWORK_LIST[account.networkKey] && NETWORK_LIST[account.networkKey].protocol || NetworkProtocols.UNKNOWN ;
 
     return (
       <ScrollView
@@ -135,7 +162,11 @@ This account can only be recovered with its associated recovery phrase.`,
           title={account.name}
         />
         <View style={styles.qr}>
-          <QrView data={accountId(account)} />
+          {
+            protocol !== NetworkProtocols.UNKNOWN
+            ? <QrView data={accountId(account)} />
+            : this.renderWarningUnknownAccount()
+          }
         </View>
       </ScrollView>
     );
@@ -157,7 +188,7 @@ const styles = StyleSheet.create({
     backgroundColor: colors.card_bg
   },
   deleteText: {
-    color: 'red'
+    color: colors.bg_alert
   },
   header: {
     flexDirection: 'row',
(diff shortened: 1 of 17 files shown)


hanwencheng:
Thanks for the PR! I have not start the part to change the account store, but basically I will change some state to the account, and will use promise to pipeline the authentication and account getter functions. I agree with the new way to access the account data, and I think merging will not be a big problem.

src/screens/AccountList.js
```diff
@@ -58,37 +57,12 @@ export default class AccountList extends React.PureComponent {
...
+    accounts: PropTypes.object.isRequired,
     onAccountSelected: PropTypes.func.isRequired,
   };
 
   constructor(props) {
     super(props);
-    this.scrollToIndex = this.scrollToIndex.bind(this);
-  }
-
-  scrollToIndex() {
```
Tbaut: This is useless IMO, and barely working. Removing it (it was playing with accountIds as well)

src/screens/MessageDetails.js
```diff
@@ -51,10 +51,6 @@ export default class MessageDetails extends React.PureComponent {
                 message={isU8a(message) ? u8aToHex(message) : message}
                 dataToSign={isU8a(dataToSign) ? u8aToHex(dataToSign) : dataToSign}
                 isHash={scannerStore.getIsHash()}
-                onPressAccount={async account => {
```
Tbaut: This is an unwanted behaviour that is not related to this PR. Users shouldn't be able to click on an account card in the message details view.

src/stores/ScannerStore.js
```diff
@@ -126,7 +126,7 @@ export default class ScannerStore extends Container<ScannerState> {
     // all the frames are filled
     if (Object.keys(this.state.multipartData.length) === frameCount) {
       // fixme: this needs to be concated to a binary blob
-      const concatMultipartData = Object.keys(this.state.multipartData).reduce((result, data) => res.concat(this.state.multipartData[data]));
+      const concatMultipartData = Object.keys(this.state.multipartData).reduce((result, data) => result.concat(this.state.multipartData[data]));
```
Tbaut: also unrelated to this PR, and this code isn't used (yet) now that's a bug for sure :)

src/util/db.js
```diff
@@ -49,18 +55,16 @@ function txKey(hash) {
...
-export const saveAccount = account => 
+export const saveAccount = (accountKey, account) => 
   SecureStorage.setItem(
-    accountId(account),
+    accountKey,
     JSON.stringify(account, null, 0),
     accountsStore
   );
 
-export const saveAccounts = accounts => accounts.forEach(saveAccount);
```
Tbaut: never used

hanwencheng:
When create new account, there is still the UNKNOWN NETWORK
WechatIMG5

Need to be removed from `AccountNetworkChooser.js`

src/util/db.js
```diff
@@ -30,10 +30,16 @@ export async function loadAccounts( version = 3 ) {
     keychainService: accountStoreVersion,
     sharedPreferencesName: accountStoreVersion
   };
-
-  return SecureStorage.getAllItems(accountsStore).then(accounts =>
-    Object.values(accounts).map(account => JSON.parse(account))
-  );
+ 
+  return SecureStorage.getAllItems(accountsStore).then(accounts => {
+      let accountMap = new Map();
```
hanwencheng: could be `const`

src/stores/AccountsStore.js
```diff
@@ -118,36 +112,38 @@ export default class AccountsStore extends Container {
...
   }
 
 
-  async deleteAccount(account) {
+  async deleteAccount(accountKey) {
     const { accounts } = this.state;
 
-    accounts.delete(accountId(account));
+    accounts.delete(accountKey);
     this.setState({ accounts });
```
hanwencheng: Though currently how we deal the selectedKey is safe, I suggest to reset the `selectedKey` to `""` after we delete it to avoid potential problem.

src/screens/Loading.js
```diff
@@ -63,25 +64,26 @@ export default class Loading extends React.PureComponent {
...
     const oldAccounts_v1 = await loadAccounts(1);
-    // v2 (up to v2.2.2) are only ethereum accounts 
+    // get a map from old accounts
+    // v2 accounts (up to v2.2.2) are only ethereum accounts 
     // with now deprectaded `chainId` and `networkType: 'ethereum'` properties
     // networkKey property is missing since it was introduced in v3.
     const oldAccounts_v2 = await loadAccounts(2);
-    const oldAccounts = [...oldAccounts_v1, ...oldAccounts_v2]
-    const accounts = oldAccounts.map(a => {
+    const oldAccounts = [...oldAccounts_v1, ...oldAccounts_v2];
```
hanwencheng: I don't know it even works without `entries`

src/screens/AccountDetails.js
```diff
@@ -103,13 +100,34 @@ This account can only be recovered with its associated recovery phrase.`,
     }
   }
 
+  renderWarningUnknownAccount = function () {
+    return (
+      <View style={styles.warningView}>
+        <Text style={{...styles.title, ...styles.warningTitle}}>Warning</Text>
+        <Text style={styles.warning}>
```
hanwencheng: missing styles for warning

src/screens/AccountDetails.js
```diff
@@ -169,12 +191,19 @@ const styles = StyleSheet.create({
...
   title: {
     color: colors.bg_text_sec,
     fontSize: 18,
     fontFamily: fonts.bold,
     flexDirection: 'column',
     justifyContent: 'center',
+  },
+  warningTitle: {
+    color: colors.bg_alert,
+    fontSize: 20,
```
hanwencheng: missing `fontFamily`

src/screens/Loading.js
Tbaut: Not sure what you mean. `loadAccounts` retrurns a map, using the spread is allowed and documented https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Relation_with_Array_objects

hanwencheng: Thanks, I thought I can only use it with `[...oldAccounts_v1.entries()]`

src/screens/AccountNetworkChooser.js
```diff
@@ -50,32 +50,34 @@ class AccountNetworkChooserView extends React.PureComponent {
...
-              }
-            ]}
-            onPress={() => {
-              accounts.updateNew(empty('', networkKey));
-              navigation.goBack();
-            }}
-          >
-            <Text
+        { Object.entries(NETWORK_LIST)
+          .filter(([networkKey]) => networkKey !== UnKnownNetworkKeys.UNKNOWN )
```
Tbaut: The commit looks bigger than it is, all I did is add this filter to make sure `Unkown network` isn't showing up.

Tbaut:
Nice catch, thanks for the review. Your comments were addressed :)

src/screens/AccountDetails.js
```diff
@@ -135,7 +153,11 @@ This account can only be recovered with its associated recovery phrase.`,
           title={account.name}
         />
         <View style={styles.qr}>
-          <QrView data={accountId(account)} />
+          {
+            protocol !== NetworkProtocols.UNKNOWN
+            ? <QrView data={selectedKey} />
+            : this.renderWarningUnknownAccount()
```
yjkimjunior: nit 2 spaces

src/screens/AccountList.js
```diff
@@ -113,7 +87,7 @@ class AccountListView extends React.PureComponent {
   render() {
     const { accounts, navigation, onAccountSelected } = this.props;
     const hasNoAccount = accounts.length < 1;
-    const { navigate } = navigation;
+    const { navigate } = navigation;;
```
yjkimjunior: 1 semicolon

src/screens/AccountList.js
```diff
@@ -138,16 +112,16 @@ class AccountListView extends React.PureComponent {
             this.list = list;
           }}
           style={styles.content}
-          data={accounts}
-          keyExtractor={account => accountId(account)}
+          data={[...accounts.entries()]}
```
yjkimjunior: why not just `data={accounts.entries()}`?

src/screens/AccountNetworkChooser.js
```diff
@@ -22,7 +22,7 @@ import { Subscribe } from 'unstated';
 import colors from '../colors';
 import fonts from "../fonts";
 import TouchableItem from '../components/TouchableItem';
-import { NETWORK_LIST } from '../constants';
+import { NETWORK_LIST, UnKnownNetworkKeys } from '../constants';
```
yjkimjunior:
```suggestion
import { NETWORK_LIST, UnknownNetworkKeys } from '../constants';
```

src/screens/MessageDetails.js
```diff
@@ -85,7 +81,7 @@ export class MessageDetailsView extends React.PureComponent {
   };
 
   render() {
-    const {dataToSign, isHash, message, onNext, onPressAccount, sender} = this.props;
+    const {dataToSign, isHash, message, onNext, sender} = this.props;
```
yjkimjunior:
```suggestion
    const { dataToSign, isHash, message, onNext, sender} = this.props;
```

src/stores/AccountsStore.js
```diff
@@ -189,26 +187,24 @@ export default class AccountsStore extends Container {
...
+    throw new Error(`no account found for the address: ${address}`);
+}
 
   getSelected() {
-    return this.state.accounts.get(this.state.selected);
+    return this.state.accounts.get(this.state.selectedKey);
+  }
+
+  getSelectedKey() {
+    return this.state.selectedKey
```
yjkimjunior:
```suggestion
    return this.state.selectedKey;
```

src/stores/AccountsStore.js
```diff
@@ -189,26 +187,24 @@ export default class AccountsStore extends Container {
...
-      .sort((a, b) => {
-        if (a.name < b.name) {
-          return -1;
-        }
-        if (a.name > b.name) {
-          return 1;
-        }
-        return 0;
-      });
+    return this.state.accounts
```
yjkimjunior:
```suggestion
    return this.state.accounts;
```

src/util/db.js
```diff
@@ -30,10 +30,16 @@ export async function loadAccounts( version = 3 ) {
...
   };
-
-  return SecureStorage.getAllItems(accountsStore).then(accounts =>
-    Object.values(accounts).map(account => JSON.parse(account))
-  );
+ 
+  return SecureStorage.getAllItems(accountsStore).then(accounts => {
+      const accountMap = new Map();
+      for (let [key, value] of Object.entries(accounts)) {
+        const account = JSON.parse(value)
```
```suggestion
        const account = JSON.parse(value);
```

src/util/db.js
```diff
@@ -30,10 +30,16 @@ export async function loadAccounts( version = 3 ) {
...
+        accountMap.set(key, {...account})
```
```suggestion
        accountMap.set(key, {...account});
```

src/util/db.js
```diff
@@ -30,10 +30,16 @@ export async function loadAccounts( version = 3 ) {
...
+      }
+
+      return accountMap
```
```suggestion
      return accountMap;
```

src/screens/AccountList.js
Tbaut: Your suggestion gives:
> Error: Invariant violation: tried to get frame for out of range index NaN

Probably because `entries()` from a map returns a map iterator object that Flatlist can't handle. (cf: [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/entries#Syntax))

PR #394, opened 2019-09-23
feat: automatically adjust text height

### LINKED ISSUE #393: Message Details Field Should Expand On Overflow
Messages that go beyond the 4 lines of the text area just get cut off. Particularly important is that the `Pay KSMs to the Kusama account: 5C.......` message cuts off the bottom half of the address.

hanwencheng:
close #393 

Android | iOS
------------ | -------------
out mp4 | out mp4 ios

Introduce the [react-native-text-size](https://github.com/aMarCruz/react-native-text-size) to calculate the number of the lines of the text and thus the height of the text component.

Also refactor MessageDetailsView into a stateless component.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -164,6 +164,7 @@ android {
 }
 
 dependencies {
+    implementation project(':react-native-text-size')
     implementation project(':react-native-camera')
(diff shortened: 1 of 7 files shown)


hanwencheng:
Indeed, so much overkill.

PR #395, opened 2019-09-30
chore: Use Kusama cc2 genesis hash

Tbaut:
- uses the Kusama cc2 genesis hash (using polkascan or apps -> tools -> chain.getBlockHash(0))
"0xe3777fa922cafbff200cadeaea1a76bd7898ad5b89f7848999058b50e715f636"

Tested with apps, could sign the tx, got "error 1010", as expected since transfer aren't enabled.

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -17,7 +17,7 @@ export const EthereumNetworkKeys = Object.freeze({
 
 // genesisHash is used as Network key for Substrate networks
 export const SubstrateNetworkKeys = Object.freeze({
-  KUSAMA: '0x3fd7b9eb6a00376e5be61f01abb429ffb0b104be05eaff4d458da48fcd425baf', // https://polkascan.io/pre/kusama/block/0
+  KUSAMA: '0xe3777fa922cafbff200cadeaea1a76bd7898ad5b89f7848999058b50e715f636', // hhttps://polkascan.io/pre/kusama-cc2/block/0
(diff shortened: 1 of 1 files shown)


src/constants.js
```diff
@@ -17,7 +17,7 @@ export const EthereumNetworkKeys = Object.freeze({
 
 // genesisHash is used as Network key for Substrate networks
 export const SubstrateNetworkKeys = Object.freeze({
-  KUSAMA: '0x3fd7b9eb6a00376e5be61f01abb429ffb0b104be05eaff4d458da48fcd425baf', // https://polkascan.io/pre/kusama/block/0
+  KUSAMA: '0xe3777fa922cafbff200cadeaea1a76bd7898ad5b89f7848999058b50e715f636', // hhttps://polkascan.io/pre/kusama-cc2/block/0
```
Tbaut:
```suggestion
  KUSAMA: '0xe3777fa922cafbff200cadeaea1a76bd7898ad5b89f7848999058b50e715f636', // https://polkascan.io/pre/kusama-cc2/block/0
```

src/constants.js
```diff
@@ -27,7 +27,7 @@ const substrateNetworkBase = {
     decimals: 12,
     genesisHash: SubstrateNetworkKeys.KUSAMA,
     prefix: 2,
-    title: 'Kusama',
+    title: 'Kusama CC2',
```
hanwencheng: >  Anything you’ve done on Kusama so far will persist into Kusama CC2
See this from the CC2 release post here, correct me if I am wrong, we can still use name `Kusama`, since currently all the users on Kusama use the same network.

Tbaut: We can still use the name Kusama, but if there's a CC3 in the future, with a different genesis hash, and we only write "Kusama", things may be confusing, I think it's better to be precise.

If there's a Kusama CC3, any account created with CC2 will need to be migrated (manually or automatically).

hanwencheng: Agree on the naming.

PR #397, opened 2019-10-01
chore: remove git restrictions with pre-git and add semantic validator for pr

hanwencheng:
#### Description
1. remove any restrictions on commit
2. add semantic validator to add the following [types](https://github.com/paritytech/parity-signer/compare/hanwen-git-hook?expand=1#diff-ad4346b0676b8cc8daabacb6e255e307R4-R15) in front of the PR as what we do before with commit.  @goldsteinsveta

#### Additional Configurartions
Since I can not change the settings of the repo, to use it we still need to do followings administration with this repo:
Install Semantic Pull Request on Github, start [here](https://github.com/apps/semantic-pull-requests/installations/new/permissions?target_id=6014309).

Another thing is that I suggest we only allow [Squash Merging](https://help.github.com/en/articles/configuring-commit-squashing-for-pull-requests) under merge button.

#### Trouble Shooting 
If after this PR you still met the restrictions with commit, please delete git hook repository
```shell
$parity-signer    rm -rf  .git/hooks/
```

### DIFF
--- a/.github/semantic.yml
+++ b/.github/semantic.yml
@@ -1,0 +1,15 @@
+# Always validate the PR title, and ignore the commits
+titleOnly: true
+
+types:
+  - major
+  - feat
+  - fix
+  - chore
+  - docs
+  - style
+  - refactor
+  - test
+  - build
+  - ci
+  - revert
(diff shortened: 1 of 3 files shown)


Tbaut:
> Another thing is that I suggest we only allow Squash Merging under merge button.

Not entirely sure why we would use the others, but since squash-merge is the default I don't really mind keeping the other options in case it can help. There's no overhead.

PR #400, opened 2019-10-02
fix #398 - Format address decoded from method arguments to appropriate prefix, format tip as balance

### LINKED ISSUE #398: Destination address is encoded with prefix 42 although I'm on Kusama
How to reproduce:

Send some funds from any address on Kusama CC2
- the `dest` address in the transaction details is always encoded with 42 prefix (starting with 5..) although it's on Kusama.

image

I suspect that it's not on Signer side, (couldn't find anything when trying to debug quickly). @yjkimjunior can you confirm? I'd send that to apps and the extension, I tested on both using Kusama encoding.

yjkimjunior:
Closes #398

IMG_0036


src/components/PayloadDetailsCard.js
```diff
@@ -217,7 +233,9 @@ function ExtrinsicPart({ label, fallback, value }) {
             ? renderMethodDetails()
```
Tbaut: I find this type of structure particularly hard to read, can we use the good old if/else?

yjkimjunior: fair enough, i don't mind either way, ill chang eit

yjkimjunior: sorry @Tbaut unfortunately I think I'll keep this as it is actually :/

Tbaut:
Damn, I knew it. it's all right

PR #409, opened 2019-10-07
style: UI and style components clean up

goldsteinsveta:


goldsteinsveta:
[clabot:check]

Tbaut:
I didn't test on my device yet. Just one thing I wonder, why do we want to keep style inline rather than centralized at the end of each file? is that a deliberate choice?

Tbaut:
Should we close this?

Issue #322, opened 2019-08-21
UOS Multipart


yjkimjunior:
Following #273, multipart payloads need to be properly tested.

* [x] - Multipart:
   * [x] - Cold Signer MUST be able to start scanning the Multipart Payload at any frame.  
   * [x] - Cold Signer MUST NOT expect the frames to come in any particular order. 
   * [x] - Cold Signer SHOULD show a progress indicator of how many frames it has successfully scanned out of the total count.
   * [x] - `part_data` MUST be stored by the Cold Signer, ordered by frame number, until all frames are scanned. 
    * [x] - Once all frames are combined, the part_data must be concatenated into a single binary blob, and then interpreted as a completely new albeit larger Payload, starting from the prefix table above.

Issue #318, opened 2019-08-13
Inject Types from Metadata QR


yjkimjunior:
Following this: https://github.com/paritytech/parity-signer/pull/273

For users who don't want to ever upgrade their Signer, eventually payload will fall out of sync with latest metadata. The solution we discussed is to host a QR code with the minified metadata to inject the latest types on the fly for decoding.

Upon scanning a "add new metadata" QR code:

- verify that the payload has been signed by a Parity key
- the user should ideally not be asked for anything else than "be careful and double check everything".

The best would be to add the genesis hash of the network in the payload.

yjkimjunior:
ok so just tested this and for call only meta v8 for Kusama this will be about 32 frames so just over a minute of holding your hand steady.

yjkimjunior:
since we use the type registry from @polkadot/api, a better and reasonably implementable UX would be to allow registering specific types as is done in @polkadot/apps

hanwencheng:
A way to do that is maybe we have a similar website as https://react-native-community.github.io/upgrade-helper/ .

PR #411, opened 2019-10-09
feat: UOS Multipart

yjkimjunior:
#322

Starting with a utf8 message as this is most likely how a metadata blob will be transferred (https://github.com/paritytech/parity-signer/issues/318). ~It would however be rather useful to add an extension range specifically for metadata payloads, although on the other hand metadata injection would probably only be with regard to Substrate based chains. @maciejhirsz?~ 

Particularly because the UOS spec says that `7B` is a forbidden first byte of any `part_data`, but that is how a metadata struct/json is going to be transfered.

* [x] - Multipart:
  * [x] - Cold Signer MUST be able to start scanning the Multipart Payload at any frame.  
   * [x] - Cold Signer MUST NOT expect the frames to come in any particular order. 
   * [x] - Cold Signer SHOULD show a progress indicator of how many frames it has successfully scanned out of the total count.
   * [x] - `part_data` MUST be stored by the Cold Signer, ordered by frame number, until all frames are scanned. 
    * [x] - Once all frames are combined, the part_data must be concatenated into a single binary blob, and then interpreted as a completely new albeit larger Payload, starting from the prefix table above.

ezgif-4-38c22cf26ef4


src/util/decoders.js
```diff
@@ -218,12 +219,12 @@ export async function constructDataFromBytes(bytes) {
...
             throw new Error(e);
           } else {
-            throw new Error('we cannot handle the payload: ', bytes);
+            throw new Error('Error: we cannot handle this payload: ', bytes);
           }
         }
         break;
       default:
-        throw new Error('we cannot handle the payload: ', bytes);
+        throw new Error('we cannot handle the pasta: ', bytes);
```
Tbaut: the pasta :) ?
Not suggesting anything since I'm not entirely sure what you wanted to change here.

yjkimjunior: > Could you add the instructions to test it?

I will updating the heroku app with more test cases for this. FYI, the calls-only metadata from kusama v7 is 26 or so frames. It could have been as low as 13 but couldn't get the camera to scan at 2048 frame size or anything above 1024. https://github.com/polkadot-js/ui/issues/224#issuecomment-539438539

> the pasta :) ?

ahah I forgot about this, just changed it to see which block was throwing. 

> We somehow need to reset the isMultipart and/or frameCount at component unmount 

+1. I kept that whilst testing the 26 frames as losing that progress can be devastating, but
I'm thinking of adding an alert that asks whether you want to save your progress or not, if over 2 frames have been scanned. As Jaco mentioned most payloads will be in 2 or less frames, so that would only apply to metadata in practice.

Tbaut: >I'm thinking of adding an alert that asks whether you want to save your progress or not, if over 2 frames have been scanned. As Jaco mentioned most payloads will be in 2 or less frames, so that would only apply to metadata in practice.

:+1:

Tbaut:
Thanks, please ping me for review when you think it's good to go.

yjkimjunior:
testing contract upload on Substrate so will need to uncomment the Substrate stuff before testing. it is 15 frames to upload incrementer contract.

yjkimjunior:
hi, @Tbaut the multipart pr is ready for review again. You can test with the random message in the herokuapp, and you can test with `system.setCode` with `flipper.wasm` or something in polkadot/apps too (beware, it'll be 13 frames)

yjkimjunior:
| system.setCode(flipper.wasm) | |
|ezgif-6-eab9d360ae99| |

yjkimjunior:
Hey @Tbaut thanks for all the testing! As we discussed yesterday, I've now archived the UOS-QRs repo and heroku app as it was originally a placeholder while the QR support was being added to apps. 

From what I've tested, `system.setCode` with some small wasm file works on apps master, I'd appreciate if you tested on apps directly

yjkimjunior:
Took a while to figure out what `Invalid Trasaction: Custom(2)` meant, but it means [no permission](https://github.com/paritytech/polkadot/blob/c69ba7e2d4b4032dab11939129d890bfc88e210e/primitives/src/lib.rs#L84
) since only these modules are allowed. https://github.com/paritytech/polkadot/blob/05b1c16802ab9e6dfc0e2542c69adf5bd8e7d729/runtime/src/lib.rs#L134

Issue #410, opened 2019-10-09
bug: encode address in args for sudo proposals


yjkimjunior:
#400 needs to also account for sudo proposals

PR #413, opened 2019-10-10
fix: format nested args

yjkimjunior:
#410

Problem: Sometimes extrinsics can have a `Call` as an argument, which has a `Call` as an argument, which has a `Call` ... and so on. In the case of current SRML's AFAICT, it only goes one level deep with `sudo(proposal)`s

Solution: since we're getting to a point where we support various Substrate chains, this PR just does the above recursively to address the general case of that issue. The kvArray stuff is only necessary because Javascript objects are unordered, and the ordering matters in the final human readable render of the call.

Testing: In https://polkadot.js.org/apps/#/extrinsics, select the `sudo` extrinsic and make a proposal of any kind.


Tbaut:
Could you add a lil description with the problem and how to test things? I personally know it, but I don't think it's the case for anyone reading this issue :)

hanwencheng:
Tried sudo extrinsic with following arguments:

Screen Shot 2019-10-14 at 11 45 52

And get :
nestedCall

The arguments are not showed.

I think the problem is most related to the styling [here](https://github.com/paritytech/parity-signer/pull/413/files#diff-d2ba6932294ddb86cc948f24df49a974R226-R227), probably `flex` settings

src/components/PayloadDetailsCard.js
```diff
@@ -277,6 +298,7 @@ const styles = StyleSheet.create({
     textAlign: 'left',
     color: colors.card_bg_text,
```
hanwencheng: This value is not defined, could be removed

src/components/PayloadDetailsCard.js
```diff
@@ -259,9 +279,10 @@ const styles = StyleSheet.create({
   subLabel: {
     backgroundColor: null,
     color: colors.card_bg_text,
```
hanwencheng: not defined `card_bg_text`

src/components/PayloadDetailsCard.js
```diff
@@ -195,23 +212,26 @@ function ExtrinsicPart({ label, fallback, prefix, value }) {
...
-          <Text style={styles.secondaryText}>
-            You are calling <Text style={styles.secondaryText}>{sectionMethod}</Text> with the following arguments:
-          </Text>
+    if (formattedCallArgs) {
+      return Object.entries(formattedCallArgs).reverse().map((entry, index) => {
+        const sectionMethod = entry[0];
+        const paramArgs = entry[1];
+  
+        return (
+          <View key={index} style={{ display: 'flex', flexDirection: 'column', alignItems: 'flex-start', justifyContent: 'flex-start', paddingLeft  : 5, width: '100%' }}>
```
hanwencheng: It would be nice if we can put style into separate file.

src/components/PayloadDetailsCard.js
```diff
@@ -195,23 +212,26 @@ function ExtrinsicPart({ label, fallback, prefix, value }) {
...
-              Object.entries(argNameValue).map(([key, value]) => { return (
-                <View key={key} style={{ display: 'flex', flexDirection: 'row', flexWrap: 'wrap', padding: 5, alignItems: 'flex-start' }}>
-                  <Text style={{...styles.subLabel, flex: 1}}>{key}: </Text>
-                  <Text style={{...styles.secondaryText, flex: 3}}>{value}</Text>
-                </View>
-              )})
+              paramArgs.map(([param, arg]) => (
+                <React.Fragment key={param}>
+                  <Text style={{...styles.subLabel, flex: 1}}>{param}: </Text>
+                  <Text style={{...styles.secondaryText, flex: 3}}>{arg || undefined}</Text>
```
hanwencheng: If arg is `undefined`, then we return `undefined`, which is still not a string here.

yjkimjunior:
@hanwencheng i just tested setRemark without any changes to my source code, it was able to display correctly. I've addressed the other grumbles.

IMG_5F90AEB8F3CC-1

hanwencheng:
@yjkimjunior I still got the problem on my Android phone, forget to get the cable for iOS today, so I do not know if it is consistence or caused by something else.

src/components/PayloadDetailsCard.js
```diff
@@ -241,37 +260,31 @@ function ExtrinsicPart({ label, fallback, prefix, value }) {
...
+						<View key={index} style={styles.callDetails}>
+							<Text>
+								Call <Text style={styles.secondaryText}>{sectionMethod}</Text>{' '}
+								with the following arguments:
+							</Text>
+							{paramArgs.map(([param, arg]) => (
+								<React.Fragment key={param}>
+									<Text style={{ ...styles.subLabel, flex: 1 }}>{param}: </Text>
+									<Text style={{ ...styles.secondaryText, flex: 3 }}>
+										{arg}
```
hanwencheng: Or we do `{arg || "undefined"}` ?

src/components/PayloadDetailsCard.js
```diff
@@ -241,37 +260,31 @@ function ExtrinsicPart({ label, fallback, prefix, value }) {
...
+
+					return (
+						<View key={index} style={styles.callDetails}>
+							<Text>
+								Call <Text style={styles.secondaryText}>{sectionMethod}</Text>{' '}
+								with the following arguments:
+							</Text>
+							{paramArgs.map(([param, arg]) => (
+								<React.Fragment key={param}>
+									<Text style={{ ...styles.subLabel, flex: 1 }}>{param}: </Text>
```
hanwencheng: `flex` on  your iOS device also does not work, on Android the view with `flex` will disappear. I think it is because with `flex-direction: column` we have to set the `height` otherwise the `flex` will not work. 

Consider the height might be variant since:
1. nested call with "sudo"
2. the args could be super long, e.g. args for "setCode"

Probably we just do not set height, but gives some styling, and use ellipses for super long args.

src/components/PayloadDetailsCard.js
yjkimjunior: sometimes there is indeed no arguments, so maybe I'll add a check for that and show "... with no arguments" so it looks intentional and not error-y.

yjkimjunior:
@hanwencheng, if the payload gets to be over 256 bytes, then it must be hashed *and then* signed, so we will never run into the problem of absurdly long scrolling PayloadDetails. I couldn't reproduce the flex issue you mention, but i've set an explicit flex size for the call details row. 

below is 256 bytes, so the maximum amount of space it could possibly take up.

IMG_0021

Tbaut:
I've encountered UI issues by trying the following:
image

and I got:
image

hanwencheng:
Same problem as Thibaut, I tried remove the line:327 `flex:3`,

and now it works like: 

WechatIMG4

yjkimjunior:
ok, well that's what I had before the first grumble...

hanwencheng:
Screen Shot 2019-10-15 at 17 18 06

Tried this case occasionally and the method cannot be interpreted.

yjkimjunior:
sudoAs "occasionally" fails? will look into it.

hanwencheng:
I mean I just by chance tried the `sudoAs` method, as you try it, it will failed to be interpreted.

yjkimjunior:
ah reproduced it. So at least it's falling back to bytes as expected. Seems `sudoAs` was added recently. I've updated the metadata and now it's able to do the same here as anything else. Thank you for catching it @hanwencheng !

IMG_D507E40607D2-1

PR #414, opened 2019-10-11
fix: use keyboard scroll view for pin input screen

### LINKED ISSUE #403: Cannot see second PIN entry; blocked by keypad
DESCRIPTION:
When entering the second PIN, you cannot see the textbox as it is obscured by the numeric keypad.  Thus, when typing, you lack feedback on whether you entered a digit or not.  You are able to see how many numbers were entered after the keypad goes away (i.e. 

TO REPRODUCE:
1. Create an account (Ethereum or Kusama, it does not matter).  
2. Select icon, go past seed backup screen, until you get to the ACCOUNT PIN screen
3. Enter PIN in first (upper) textbox
4. Click on bottom textbox to enter

EXPECTED BEHAVIOR:
Can see textbox and thus verify that keys were pressed

OBSERVED BEHAVIOR:
Cannot see textbox on screen.  Users do not have feedback on whether or not a key was pressed.

SCREENSHOT:

pin_entry

NOTES:
This was on an iPhone 6 SE (Model Number MP7T2LL/A).

The number is visible after dismissing the keypad popup.

hanwencheng:
fix #403  

screenshot on iPhone SE @laboon

Also click on the blank place will hide the keyboard.

out

### DIFF
--- a/src/screens/AccountPin.js
+++ b/src/screens/AccountPin.js
@@ -121,7 +121,7 @@ class AccountPinView extends React.PureComponent {
 	render() {
 		const title = 'ACCOUNT PIN';
 		return (
-			<KeyboardScrollView style={styles.body}>
+			<KeyboardScrollView style={styles.body} extraHeight={120}>
 				<Background />
 				<Text style={styles.titleTop}>{title}</Text>
 				{this.showHintOrError()}
@@ -168,7 +168,7 @@ class PinInput extends Component {
 				multiline={false}
 				autoCorrect={false}
 				numberOfLines={1}
-				returnKeyType="next"
+				returnKeyType="done"
 				secureTextEntry
 				style={styles.pinInput}
 				{...this.props}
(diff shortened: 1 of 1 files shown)


yjkimjunior:
hmm, when I test what we have currently, pressing Done on the keyboard closes it, then I can press done the confirm the pin.

hanwencheng:
The done I think it is still useful to dismiss the keyboard once we can completely show the textInput. Here is the updated version:

out

src/screens/AccountPin.js
```diff
@@ -120,7 +121,7 @@ class AccountPinView extends React.PureComponent {
 	render() {
 		const title = 'ACCOUNT PIN';
 		return (
-			<View style={styles.body}>
+			<KeyboardScrollView style={styles.body} extraHeight={120}>
```
yjkimjunior: probably not a big deal but does this hardcoded `extraHeight` look okay with smaller devices?

hanwencheng:
IMO Iphone SE is the device with smallest screen on iOS.  Here is the tested on Android Nexus One:

out

Issue #404, opened 2019-10-06
Feat: Move to root phrase HDKD model of accounts


yjkimjunior:
Basically one seed phrase unlocks a root account under which there are many derived addresses (which have no phrase themselves). 

Each root has no network affiliation. The keys derived from each root will each have a network affiliation.

1. ### New Account Creation
* create a *root* from which the user can specify to create another `N` derivations.
* meaning, remove the derivation path field from this page. It should be moved to the AccountDetails page.

2. ### AccountsList (Home) 
* only show the *root* accounts
* prepopulate with defaults: `//kusama, //ethereum, //edgeware`
* since root accounts have no network affiliation these should not show the network indicator as we currently do.

3. ### AccountDetails 
* show a list of derivation paths, alphabetically sorted
* prepopulate with defaults `//funding, //stash, //controller`
* user would also need to specify the network affiliation and path, and press a button to generate an additional series of derived accounts, `N ... N + 4`, where `N` is the number of indexed child derivations from the account already. 
* each derived account can optionally have their own names, if they don't just display the path.
* should be able to tap on any of the derived keys to get to the QR screen to display its public key.

**N.B.** the *//funding* is not the actual account though it'll have a pub/priv keypair. The only accounts the user should be able to show public key for are the soft derivations from that come from this keypair.

##### Hard vs Soft key recap:
Hard - hash the existing private key and get a new private key (and hence a new public key as well).
Soft - derive a new public key directly from the existing private key.

hanwencheng:
From reading the BIP32 standard [here](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki/#extended_Keys) I found the definition about the soft keys and hard keys may be debatable here:

> The fact that they are equivalent is what makes non-hardened keys useful (one can derive child public keys of a given parent key without knowing any private key), and also what distinguishes them from hardened keys. The reason for not always using non-hardened keys (which are more useful) is security; 

From my understanding, both hard and soft derivation will create new keypair, but the difference is that:
hard - public key could also derived child public key
soft  - public key could not derive child public key

PR #415, opened 2019-10-11
feat: create derive account button in account details

hanwencheng:
First functional PR for #404 

In account detail screen, create a `derive account` button for root identity (Account without derivation path).

User may create derived account from the seed.


hanwencheng:
out

hanwencheng:
Closed, this should be implemented in #431

PR #416, opened 2019-10-14
feat: Biometric auth

sjeohp:


sjeohp:
The objective here is to allow longer alpha-numeric pins to be practical.

- Set a complex pin
- Enable biometric auth
  - this encrypts and stores the pin using a biometric key in secure-hardware
- Sign/backup/etc
  - secure-hardware decrypts the pin
  - pin is used to decrypt seed, etc.
  - if biometry fails for some reason (EG. if someone enrolls a new fingerprint or disables biometry in device settings, the existing keys will be invalidated), then fall back to regular pin 

Additionally: sign transactions while minimizing time the decrypted seed spends in memory, IE. decrypt and sign in rust without returning it to js.

ios/NativeSigner.xcodeproj/project.pbxproj
```diff
@@ -1661,7 +1634,7 @@
 				PRODUCT_BUNDLE_IDENTIFIER = io.parity.NativeSigner;
 				PRODUCT_NAME = NativeSigner;
 				SWIFT_OBJC_BRIDGING_HEADER = "NativeSigner/NativeSigner-Bridging-Header.h";
-				SWIFT_VERSION = 5.0;
+				SWIFT_VERSION = 4.2;
```
hanwencheng: Why downgrade swift version here?

hanwencheng:
Many thanks for the PR!

From what I see the biometric auth workflow is:  react native js code--> native code --> rust code, is that correct?

By the way I have not find the code for biometric auth part in native code, am I too early to review it now?

For the native secure functions: it would be nice if you could make some test code in rust part or on Native part (on #353 I am also working on test cases). Both for the future development and the current reviewing and testing process.

ios/NativeSigner.xcodeproj/project.pbxproj
sjeohp: I was testing with an old Xcode that didn't support 5.  This change isn't important, I actually forgot about it.

sjeohp:
> By the way I have not find the code for biometric auth part in native code, am I too early to review it now? Otherwise could you please point them out ?

https://github.com/sjeohp/secure-native 

> For the native secure functions: it would be nice if you could make some test code in rust part or on Native part (on #353 I am also working on test cases). Both for the future development and the current reviewing and testing process.

👍

hanwencheng:
I have the sample integration test for testing native modules, which is on `hanwen-native-test` branch, most related to this [commit](https://github.com/paritytech/parity-signer/commit/4ad0e1fc96386d3d0fe9bdc47d7c4fd94df483c7).

sjeohp:
> I have the sample integration test for testing native modules, which is on `hanwen-native-test` branch, you may add another test as [here](https://github.com/paritytech/parity-signer/commit/4ad0e1fc96386d3d0fe9bdc47d7c4fd94df483c7#diff-631404afaba023d63642e27678c82257R32).

I don't believe it will work in this case since the methods will fail without biometric input..

hanwencheng:
Ah, it is possible on iOS, first you start with [await device.setBiometricEnrollment(true)](https://github.com/wix/Detox/blob/master/docs/APIRef.DeviceObjectAPI.md#devicesetbiometricenrollmentbool-ios-only) and then use `await device.matchFinger()` to pass the biometric auth.

ios/NativeSignerTests/SecureNativeTests.swift
```diff
@@ -0,0 +1,101 @@
+//
+//  SecureNativeTests.swift
+//  NativeSignerTests
+//
+//  Created by Joseph Mark on 11/11/19.
+//  Copyright © 2019 Facebook. All rights reserved.
```
yjkimjunior: Not sure that we should give Facebook the copyright for this :p

sjeohp:
> @sjeohp would you mind just going back and adding the Parity license header in the changed files please? I think it's just in SecureNativeTests.swift and the BridgingHeader file

Sure, done

hanwencheng:
Would be happy to review and test it once the conflict are solved.

sjeohp:
> Would be happy to review and test it once the conflict are solved.

resolved!

android/app/src/androidTest/java/io/parity/signer/EthkeyBridgeTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+    public void resolve(@Nullable Object value) {
+        assertEquals(value, this.expectedValue);
+    }
+
+    public void reject(String code, String message) {
+        assertTrue(this.shouldBeRejected);
+    }
+
+    public void reject(String code, Throwable throwable) {
```
hanwencheng: Where do we use `reject` method? for simplification we could also do 
```
public void reject(Object... x) {
  assertTrue(this.shouldBeRejected)
}
```

sjeohp: Not sure I understand.. we don't use anything in this file, since I didn't find a way to successfully mock a ReactApplicationContext in the test environment.  I included the file in case anyone else wants to make another attempt, but it can be discarded.

package.json
```diff
@@ -98,15 +100,15 @@
         "build": "xcodebuild -project ios/NativeSigner.xcodeproj -scheme NativeSigner -configuration Debug -sdk iphonesimulator -derivedDataPath ios/build/NativeSigner",
         "type": "ios.simulator",
         "device": {
-          "type": "iPhone SE"
+          "type": "iPhone 8"
```
hanwencheng: for backward compatible reason better to keep small screen iphone as tester phone

rust/signer/src/lib.rs
```diff
@@ -89,11 +93,73 @@ fn qrcode_bytes(data: &[u8]) -> Option<String> {
...
+#[derive(Debug, PartialEq)]
+struct SURI {
+    phrase: String,
+    path: String,
+    password: String,
+}
+
+fn parse_suri(suri: &str) -> Result<SURI, String> {
+    fn parse_derivation_path(input: &str) -> Result<(String, String), String> {
+        let re = Regex::new(r"^((?:\/\/?[^/]+)*)(?:\/\/\/(.*))?$").expect("invalid regex");
```
hanwencheng: what is `parse_suri` function used for? is it only used for substrate seed?

what is the difference with the one in `suri.js`, 

If not, this regex does not catch the regex of brain wallet seed with extra white space in this [PR](https://github.com/paritytech/parity-signer/pull/467), e.g '  good//sdsadsa/sdsa'

sjeohp: It is identical to the one in `suri.js` (prior to your PR).  The point is to replicate logic so that the seed/suri can be used correctly in rust without returning to js.

> If not, this regex does not catch the regex of brain wallet seed with extra white space in this PR, e.g ' good//sdsadsa/sdsa'

I am updating it, along with other grumbles.

rust/signer/src/lib.rs
```diff
@@ -236,15 +296,83 @@ export! {
...
+    fn sn_substrate_brainwallet_sign(pin: Result<String, String>, path: String, password: String, message: String, encrypted: String, legacy: bool) -> Result<String, String> {
+        let seed = decrypt(&encrypted, &pin?)?;
+        let suri = if legacy { seed } else {
+            let phrase = parse_suri(&seed)?.phrase;
+            construct_suri(&phrase, &path, &password)
+        };
+        let keypair = sr25519::KeyPair::from_suri(&suri).ok_or(format!("KeyPair from suri"))?;
+        let message: Vec<u8> = message.from_hex().map_err(|e| format!("{}", e))?;
+        let signature = keypair.sign(&message);
+        Ok(signature.to_hex())
```
hanwencheng: Maybe a dumb question, at this point, does rust auto garbage collect `keypair`?

rust/signer/src/util.rs
```diff
@@ -239,15 +284,64 @@ impl<'jni, Inner: Return<'jni, Env = JNIEnv<'jni>> + Default> Return<'jni> for O
...
+                    .new_object(class, "()V", &[])
+                    .expect("Must be able to instantiate the Exception; qed")
+                    .into();
+
+                // !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+                // !!!!                                                        !!!!
+                // !!!! WE CAN NO LONGER INTERACT WITH JNIENV AFTER THIS POINT !!!!
+                // !!!!                                                        !!!!
+                // !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+                env.throw(exception)
```
hanwencheng: These error handling part seems duplicate with the following one, is it possible to wrap them into a function?

src/screens/AccountDetails.js
```diff
@@ -73,14 +73,48 @@ class AccountDetailsView extends React.PureComponent {
...
-				next: value,
-				onDelete: this.onDelete
-			});
+			if (accounts.getSelected().biometricEnabled) {
+				try {
+					await accounts.unlockAccountWithBiometric(accounts.getSelectedKey());
+					if (value === 'AccountDelete') {
+						this.onDelete();
+					} else if (value === 'AccountBiometric') {
+						await accounts.disableBiometric(accounts.getSelectedKey());
```
hanwencheng: better extract `accounts.getSelectedKey()` out to reduce duplicated code and improve readability.

src/screens/AccountDetails.js
```diff
@@ -73,14 +73,48 @@ class AccountDetailsView extends React.PureComponent {
...
+					} else {
+						navigate(value);
+					}
+				} catch (e) {
+					console.log(e);
+					Alert.alert('Biometric Error', e.message, [
+						{
+							onDismiss: () => {
+								navigate('AccountUnlock', {
```
hanwencheng: better wrap following  block into function
```
								navigate('AccountUnlock', {
									next: value,
									onDelete: this.onDelete
								});
```

and put the alert function into `alertUtils.js`

src/screens/AccountDetails.js
```diff
@@ -73,14 +73,48 @@ class AccountDetailsView extends React.PureComponent {
...
+	onOptionSelect = async value => {
 		const navigate = this.props.navigation.navigate;
+		const accounts = this.props.accounts;
 
 		if (value !== 'AccountEdit') {
-			navigate('AccountUnlock', {
-				next: value,
-				onDelete: this.onDelete
-			});
+			if (accounts.getSelected().biometricEnabled) {
```
hanwencheng: better use `match..case` instead of `if..else` to reduce nested blocks

src/screens/AccountUnlock.js
```diff
@@ -56,8 +56,8 @@ export class AccountUnlockAndSign extends React.PureComponent {
 									}),
 									NavigationActions.navigate({ routeName: next })
 								],
-								index: 1, // FIXME workaround for now, use SwitchNavigator later: https://github.com/react-navigation/react-navigation/issues/1127#issuecomment-295841343
-								key: undefined
+								index: 1,
+								key: undefined // FIXME workaround for now, use SwitchNavigator later: https://github.com/react-navigation/react-navigation/issues/1127#issuecomment-295841343
```
hanwencheng: I previously missed this block, use `navigateToLegacyAccountList` from `navigationHelpers.js`

src/screens/AccountUnlock.js
```diff
@@ -84,8 +84,26 @@ export class AccountUnlock extends React.PureComponent {
 								pin
 							);
 						}}
-						navigate={() => {
-							if (next === 'AccountDelete') {
+						navigate={async pin => {
+							if (next === 'AccountBiometric') {
+								navigation.goBack();
+								if (accounts.getSelected().biometricEnabled) {
+									await accounts.disableBiometric(accounts.getSelectedKey());
```
hanwencheng: extract `accounts.getSelectedKey()`

src/screens/AccountUnlock.js
```diff
@@ -84,8 +84,26 @@ export class AccountUnlock extends React.PureComponent {
 								pin
 							);
 						}}
-						navigate={() => {
-							if (next === 'AccountDelete') {
+						navigate={async pin => {
+							if (next === 'AccountBiometric') {
```
hanwencheng: extract this block of code out of `render()` into a function, to make the JSX part code clean and slightly performance improvement.

src/screens/AccountUnlock.js
```diff
@@ -84,8 +84,26 @@ export class AccountUnlock extends React.PureComponent {
...
+										.catch(error => {
+											// error here is likely no fingerprints/biometrics enrolled, so should be displayed to the user
+											console.log(error);
+											Alert.alert('Biometric Error', error.message, [
+												{
+													style: 'default',
+													text: 'Ok'
+												}
+											]);
+										});
```
hanwencheng: This block of code to handle `AccountBiometric` selection, better be rather wrapped into `onToggleBiometric` function like `onDelete` rather than put it here.

`AccountUnlock` Component as its name implied, it mainly contains the logic for user for unlocking the account, the action next, better be defined in where it is called, and keep this part straightforward.

src/screens/AccountUnlock.js
```diff
@@ -84,8 +84,26 @@ export class AccountUnlock extends React.PureComponent {
...
+							if (next === 'AccountBiometric') {
+								navigation.goBack();
+								if (accounts.getSelected().biometricEnabled) {
+									await accounts.disableBiometric(accounts.getSelectedKey());
+								} else {
+									await accounts
+										.enableBiometric(accounts.getSelectedKey(), pin)
+										.catch(error => {
+											// error here is likely no fingerprints/biometrics enrolled, so should be displayed to the user
+											console.log(error);
```
hanwencheng: remove this line?

src/screens/AccountUnlock.js
```diff
@@ -95,10 +113,13 @@ export class AccountUnlock extends React.PureComponent {
 											routeName: 'LegacyAccountList'
 										}),
 										NavigationActions.navigate({ routeName: 'AccountDetails' }),
-										NavigationActions.navigate({ routeName: next })
+										NavigationActions.navigate({
+											params: next === 'AccountBiometric' ? { pin: pin } : {},
```
hanwencheng: `next` wouldn't be 'AccountBiometric', there is already a `if` sentences before.

src/screens/MessageDetails.js
```diff
@@ -45,27 +45,34 @@ import { getIdentityFromSender } from '../util/identitiesUtils';
...
+					senderIdentity
+				);
+				await scannerStore.signDataWithSeedPhrase(
+					seedPhrase,
+					NETWORK_LIST[sender.networkKey].protocol
+				);
+				return navigateToSignedMessage(this.props.navigation);
+			} catch (e) {
+				scannerStore.setErrorMsg(e.message);
+			}
```
hanwencheng:
```suggestion
	async onSignMessage(scannerStore, accountsStore, sender) {
		try {
			if (sender.biometricEnabled) {
				await scannerStore.signDataBiometric(sender.isLegacy);
				return navigateToSignedMessage(this.props.navigation);
			}
			if (sender.isLegacy) {
				return this.props.navigation.navigate('AccountUnlockAndSign', {
					next: 'SignedMessage'
				});
			}
			const senderIdentity = getIdentityFromSender(
				sender,
				accountsStore.state.identities
			);
			const seedPhrase = await unlockSeedPhrase(
				this.props.navigation,
				senderIdentity
			);
			await scannerStore.signDataWithSeedPhrase(
				seedPhrase,
				NETWORK_LIST[sender.networkKey].protocol
			);
			return navigateToSignedMessage(this.props.navigation);
		} catch (e) {
			scannerStore.setErrorMsg(e.message);
		}
	}
```

At the same time, now there is no error handling if `signDataBiometric` returns false.

It is better to wrap `signDataBiometric` (in `ScannerStore.js`) into a promise, and throw error with like `e.message = "Biometric data not correct, please try again"`. So that the error could be catched by the block here and displayed by `scannerStore.setErrorMsg(e.message);`

src/screens/MessageDetails.js
```diff
@@ -90,9 +97,9 @@ export default class MessageDetails extends React.PureComponent {
 								}
 								prehash={scannerStore.getPrehashPayload()}
 								isHash={scannerStore.getIsHash()}
-								onNext={() =>
-									this.onSignMessage(scannerStore, accountsStore, sender)
-								}
+								onNext={async () => {
+									this.onSignMessage(scannerStore, accountsStore, sender);
```
hanwencheng: It is not needed to use `async` here, no `await` in the block.

src/screens/TxDetails.js
```diff
@@ -46,27 +46,34 @@ import { getIdentityFromSender } from '../util/identitiesUtils';
...
+					senderIdentity
+				);
+				await scannerStore.signDataWithSeedPhrase(
+					seedPhrase,
+					NETWORK_LIST[sender.networkKey].protocol
+				);
+				return navigateToSignedTx(this.props.navigation);
+			} catch (e) {
+				scannerStore.setErrorMsg(e.message);
+			}
```
hanwencheng: would suggest same as in `messageDetails.js`

src/screens/AccountDetails.js
```diff
@@ -73,14 +73,48 @@ class AccountDetailsView extends React.PureComponent {
...
+					await accounts.unlockAccountWithBiometric(accounts.getSelectedKey());
+					if (value === 'AccountDelete') {
+						this.onDelete();
+					} else if (value === 'AccountBiometric') {
+						await accounts.disableBiometric(accounts.getSelectedKey());
+					} else {
+						navigate(value);
+					}
+				} catch (e) {
+					console.log(e);
```
hanwencheng: debugging code..

src/stores/AccountsStore.js
```diff
@@ -284,6 +333,37 @@ export default class AccountsStore extends Container<AccountsStoreState> {
...
+					pinKey: v4()
+				});
+				return this.save(accountKey);
+			});
+	}
+
+	async updateBiometric(accountKey, account, pin) {
+		return secureDelete(APP_ID, account.pinKey)
+			.catch(e => {
+				console.log(e);
```
hanwencheng: debugging code..

src/stores/ScannerStore.js
```diff
@@ -406,6 +409,59 @@ export default class ScannerStore extends Container<ScannerState> {
...
+
+			if (!isEthereum) {
+				// TODO: tweak the first byte if and when sig type is not sr25519
+				const sig = u8aConcat(SIG_TYPE_SR25519, hexToU8a('0x' + signedData));
+				signedData = u8aToHex(sig, -1, false); // the false doesn't add 0x
+			}
+
+			this.setState({ signedData });
+		} catch (e) {
+			console.log(e);
```
hanwencheng: as comment in `MessageDetails.js`, better pass that error message back.

src/stores/ScannerStore.js
```diff
@@ -406,6 +409,59 @@ export default class ScannerStore extends Container<ScannerState> {
...
 	}
 
+	async signDataBiometric(legacy = false) {
+		const { dataToSign, isHash, sender } = this.state;
+
+		const isEthereum =
+			NETWORK_LIST[sender.networkKey].protocol === NetworkProtocols.ETHEREUM;
+
+		let signable;
+		if (isEthereum) {
```
hanwencheng: duplicate code here.

src/stores/ScannerStore.js
```diff
@@ -406,6 +409,59 @@ export default class ScannerStore extends Container<ScannerState> {
...
+			} else if (isAscii(dataToSign)) {
+				signable = hexStripPrefix(asciiToHex(dataToSign));
+			}
+		}
+
+		let signedData;
+		try {
+			if (isEthereum) {
+				signedData = await secureEthkeySign(
+					APP_ID,
```
hanwencheng: better put `APP_ID` into app scope `native.js`

src/stores/ScannerStore.js
```diff
@@ -406,6 +409,59 @@ export default class ScannerStore extends Container<ScannerState> {
...
+					sender.pinKey,
+					signable,
+					sender.encryptedSeed
+				);
+			} else {
```
hanwencheng:
```suggestion
			const signingFunction = isEthereum ? secureEthkeySign : secureSubstrateSign;
			signedData = signingFunction(APP_ID,
				sender.pinKey,
				signable,
				sender.encryptedSeed,
				legacy
				);
```

hanwencheng:
Thanks for the effort, especially for writing the test code and wrapping signing into Rust.

Leave some comments so far as I guess integration with v4 keypairs is still in progress.

Apart from that:
1. I see lots of rust styling changes, do rust has any official lint method? styles like trailing comma are really abitary according to different editors. Not neccesary in this commit, but it would be nice to have lint for Rust code in the future.
2. I see native test code in both android and ios, how could we run these test code?
3. In terms of e2e test with `SecureNativeTest.js` file, we still need to create a spec file like `e2e/native.spec.js` in this [PR](https://github.com/paritytech/parity-signer/pull/353/files), so that we could run `yarn e2e:ios` or `yarn e2e:android` for testing them.

android/app/src/androidTest/java/io/parity/signer/EthkeyBridgeTest.java
hanwencheng: I mean there are many duplcated function name just because the arguments are different, but arguments are not used, by using `Object... x`, we will ignore them and only have one `reject` function

rust/signer/src/lib.rs
hanwencheng: I see, I forget what it is in the previous `suri.js`, but this one seems like the regex from `parseDerivationPath`, which is not the `parseSuri` one.

sjeohp: Yes, the regex above is from `parseDerivationPath`, which is here nested as `parse_derivation_path` inside `parse_suri`.  The other one below should be identical to the regex from `parseSURI` in suri.js (before your recent changes).  The overall logic should be the same.

hanwencheng: ah, yeah, I missed the nested function.

rust/signer/src/lib.rs
sjeohp: Yes, it gets destroyed when it leaves scope (when the function returns)

rust/signer/src/util.rs
sjeohp: It's not as straightforward as you might think, so I won't do it now.  Over time though I will absorb the Return/Arg traits into the `secure-native` lib, which is tidier, and remove them from this file.

src/screens/MessageDetails.js
sjeohp: The only likely biometric error here will be an invalidated fingerprint, in which case it should fallback to non-biometric pin entry so the user can still sign.  The boolean return is the simplest way of doing that.  But I agree it should tell the user something.

package.json
sjeohp: iPhone 8 is the earliest simulator that comes with the latest Xcode tooling.

sjeohp:
>     1. I see lots of rust styling changes, do rust has any official lint method? styles like trailing comma are really abitary according to different editors. Not neccesary in this commit, but it would be nice to have lint for Rust code in the future.
>

Let's talk about this offline.. 

>     2. I see native test code in both android and ios, how could we run these test code?
> 

I run them in AndroidStudio or Xcode.  In theory you should be able to use `xcodebuild` from the command line.  The android unit tests don't work though, I just left them in for reference, as discussed.

>     3. In terms of e2e test with `SecureNativeTest.js` file, we still need to create a spec file like `e2e/native.spec.js` in this [PR](https://github.com/paritytech/parity-signer/pull/353/files), so that we could run `yarn e2e:ios` or `yarn e2e:android` for testing them.
> 

Having issues getting e2e working with the native lib at all, will discuss.

>     4. Just wondering, I see the dependency of the rust code library from your github repo, how do you think the pros and cons for migrating them here?(I mean keep the repos of course, but instead of "depend on it",  moving the code here, as in single codebase. My opinion it is easy to manager if codes are related)

It should be separate, so that it can be easily used in other projects in the future.  We can publish it as a crate at some stage so we don't have to depend on a github repo.

hanwencheng:
> I run them in AndroidStudio or Xcode. In theory you should be able to use xcodebuild from the command line.

Could you please put the command into `package.json`, and use a script like `yarn unit:ios` for example.

> Having issues getting e2e working with the native lib at all, will discuss.

Is it still the problem with import? does `import {NativeModule} from 'react-native'` works ?

src/screens/MessageDetails.js
hanwencheng: I see you have already wrapped them into a promise.

The setErrorMsg I think it only display a warning message on the screen, in the fingerprint authentication case, we could show, `failed, please try again`. if so then they could be grouped together into the `try` block, which may easy the flow. 

But I am not sure on the fingerprint authentication case, maybe that would be annoying since the sensor is checking all the time (e.g. 10 times per second). Since I have not try the feature on real device.

sjeohp:
> > I run them in AndroidStudio or Xcode. In theory you should be able to use xcodebuild from the command line.
> 
> Could you please put the command into `package.json`, and use a script like `yarn unit:ios` for example.

Yep, done

> > Having issues getting e2e working with the native lib at all, will discuss.
> 
> does `import {NativeModule} from 'react-native'` works ?

That breaks jest and still doesn't let me run the detox test.  Working e2e tests for the native modules might have to be a separate PR.

src/screens/IdentityManagement.js
```diff
@@ -39,17 +42,44 @@ function IdentityManagement({ accounts, navigation }) {
...
+
+	const onOptionSelect = async value => {
 		if (value === 'PathDelete') {
 			alertDeleteIdentity(async () => {
-				await unlockSeedPhrase(navigation);
-				const deleteSucceed = await accounts.deleteCurrentIdentity();
-				if (deleteSucceed) {
-					navigateToLandingPage(navigation, true);
+				if (currentIdentity.biometricEnabled) {
+					await unlockIdentitySeedWithBiometric(currentIdentity)
```
hanwencheng: `await` and `promise.then()` do not occurs at same time, please use `await` instead

src/screens/PathDetails.js
```diff
@@ -49,19 +51,88 @@ export function PathDetailsView({ accounts, navigation, path, networkKey }) {
...
+	}
+
+	async function noBiometric(value) {
+		try {
+			if (value === 'PathDelete') {
+				alertDeleteAccount('this key pairs', async () => {
+					await unlockSeedPhrase(navigation);
+					await onDelete();
+				});
+			} else if (false) {
```
hanwencheng: why use `if(false)` here?

src/screens/PathDetails.js
```diff
@@ -49,19 +51,88 @@ export function PathDetailsView({ accounts, navigation, path, networkKey }) {
...
+		try {
+			if (value === 'PathDelete') {
+				alertDeleteAccount('this key pairs', async () => {
+					await unlockIdentitySeedWithBiometric(currentIdentity)
+						.then(onDelete)
+						.catch(() => {
+							unlockSeedPhrase(navigation).then(onDelete);
+						});
+				});
+			} else if (false) {
```
hanwencheng: also here unreachable code

src/screens/AccountDetails.js
```diff
@@ -98,7 +143,7 @@ function AccountDetails({ accounts, navigation }) {
 				</View>
 			</View>
 			<AccountCard
-				address={account.address}
+				accountId={account.address}
```
hanwencheng:
```suggestion
				address={account.address}
```

src/screens/IdentityBackup.js
```diff
@@ -60,7 +73,7 @@ function IdentityBackup({ navigation, accounts }) {
 		return () => {
 			setSeedPhrase('');
 		};
-	}, [isNew, navigation]);
+	}, [accounts, currentIdentity, isNew, navigation]);
```
hanwencheng:
```suggestion
	}, [currentIdentity, isNew, navigation]);
```

Issue #402, opened 2019-10-02
Show the "Substrate dev" network in Dev mode


Tbaut:
We used it in the past and it was super useful to do end to end test with a Substrate-like network (e.g using the dev-node genesish hash).

Basically the task is: un-commenting what has been commented here: https://github.com/paritytech/parity-signer/pull/384/files and only allow showing the "Substrate Dev" in the network chooser when in dev mode (see https://github.com/paritytech/parity-signer/pull/365 also for reference how to allow things in dev)

PR #417, opened 2019-10-14
chore: Enable "Substrate Dev" netwok in Dev mode

Kriyszig:
Issue in focus: #402 
* Removed in #384, this brings back the Substrate Dev network when `__DEV__` is active

In case any changes are required, please let me know and I'll address them ASAP!


Kriyszig:
[clabot:check]

Tbaut:
Thanks for tackling this. It needs a little more than this if we want things to work.
There is also the need for adding the `static-substrate.js` because it's been deleted here https://github.com/paritytech/parity-signer/pull/384/files

Also as explained in the issue, the "Substrate Dev" should be shown in the Network chooser when we are in Dev.

Finally, the "Substrate Dev" Network should be uncommented in `constants.js`

Kriyszig:
@Tbaut Made the necessary changes in the latest commit. Out if the commented lines I have replaced `const isSubstrateDev = this.props.prefix === SUBSTRATE_NETWORK_LIST[SubstrateNetworkKeys.SUBSTRATE_DEV].prefix;` with just `__DEV__` to detect the Dev environment, should I revert that?

src/components/PayloadDetailsCard.js
```diff
@@ -61,15 +60,14 @@ export default class PayloadDetailsCard extends React.PureComponent {
 				decimals: SUBSTRATE_NETWORK_LIST[SubstrateNetworkKeys.KUSAMA].decimals,
 				unit: SUBSTRATE_NETWORK_LIST[SubstrateNetworkKeys.KUSAMA].unit
 			});
+		} else if (__DEV__) {
```
Tbaut: We should not chose those metadata by default. So `isSubstrateDev` should definitely be checked
```suggestion
		} else if (__DEV__ && isSubstrateDev) {
```

src/components/PayloadDetailsCard.js
```diff
@@ -51,7 +51,6 @@ export default class PayloadDetailsCard extends React.PureComponent {
 		const isKusama =
 			this.props.prefix ===
 			SUBSTRATE_NETWORK_LIST[SubstrateNetworkKeys.KUSAMA].prefix;
-		// const isSubstrateDev = this.props.prefix === SUBSTRATE_NETWORK_LIST[SubstrateNetworkKeys.SUBSTRATE_DEV].prefix;
```
Tbaut: This should be uncommented

src/constants.js
```diff
@@ -25,8 +25,9 @@ export const EthereumNetworkKeys = Object.freeze({
 
 // genesisHash is used as Network key for Substrate networks
 export const SubstrateNetworkKeys = Object.freeze({
-	KUSAMA: '0xe3777fa922cafbff200cadeaea1a76bd7898ad5b89f7848999058b50e715f636' // https://polkascan.io/pre/kusama-cc2/block/0
-	// SUBSTRATE_DEV: '0x4393a679e1830a487e8ae92733f089a80f3e24ba515b08dd8adb40fc6cedee8d', // substrate --dev commit ac6a2a783f0e1f4a814cf2add40275730cd41be1 hosted on wss://dev-node.substrate.dev .
+	KUSAMA: '0xe3777fa922cafbff200cadeaea1a76bd7898ad5b89f7848999058b50e715f636', // https://polkascan.io/pre/kusama-cc2/block/0
+	SUBSTRATE_DEV:
+		'0x4393a679e1830a487e8ae92733f089a80f3e24ba515b08dd8adb40fc6cedee8d' // substrate --dev commit ac6a2a783f0e1f4a814cf2add40275730cd41be1 hosted on wss://dev-node.substrate.dev .
```
Tbaut: Having it on one line would be better.

src/components/PayloadDetailsCard.js
Kriyszig: Addressed in 5c04b93

src/components/PayloadDetailsCard.js
Kriyszig: Addressed in 5c04b93

src/constants.js
Kriyszig: That previously gave linting error so I used `yarn run lint --fix`

PR #423, opened 2019-10-16
chore: lock the version of eslint-babel

hanwencheng:
### TLDR:

ESLint does not works as expected, this fix the `no-unused-var` error in "for.. in" of Javascript, and revert previous workaround.

### Long version:

Eslint evaluate `no-unused-var` for the keys in "for.. in" loop with eslint-babel: 10.0.1, which discussed [here](https://github.com/eslint/eslint/issues/12117).

So we have changed the dependency version, but this is still used in @react-native-community/eslint-config, and @react-native-community/eslint-config is rarely updated.

Though I have changed the code in `yarn.lock`, but every time we add new dependencies, remove dependencies, the lock file will be rewrite. It will further cause the original issue, it is just not the correct way to lock dependency's version.

So the correct way to do that is with yarn's [selective-version-resolutions](https://yarnpkg.com/lang/en/docs/selective-version-resolutions/) with a tiny modification on `package.json` file. That's what this PR about. **Also I suggest remove `package-lock.json` file if there is any**.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,5 +77,8 @@
     "prettier": "1.18.2",
     "react-test-renderer": "16.9.0",
     "reactotron-react-native": "^3.6.5"
+  },
+  "resolutions": {
+    "@react-native-community/eslint-config/babel-eslint": "10.0.3"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -1110,7 +1110,7 @@
   dependencies:
     "@typescript-eslint/eslint-plugin" "^1.5.0"
     "@typescript-eslint/parser" "^1.5.0"
-    babel-eslint "10.0.3"
+    babel-eslint "10.0.1"
     eslint-plugin-eslint-comments "^3.1.1"
     eslint-plugin-flowtype "2.50.3"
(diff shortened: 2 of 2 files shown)


yjkimjunior:
Hey @hanwencheng the rationale and the discussion you [linked](https://github.com/eslint/eslint/issues/12117) both sound reasonable, but I couldn't reproduce on  master branch with 
```
for (let i in [0, 1, 2]) {
    console.log(i);
}
```
and eslint doesn't complain.

hanwencheng:
Ah, this is because the workaround I did before. 

> Though I have changed the code in yarn.lock, but every time we add new dependencies, remove dependencies, the lock file will be rewrite. It will further cause the original issue, it is just not the correct way to lock dependency's version.

Once you do `yarn add xxx` again, this [line](https://github.com/paritytech/parity-signer/pull/423/files#diff-8ee2343978836a779dc9f8d6b794c3b2L1113) of `yarn.lock` file will changed, so we need to manually fix this line again to prevent further reinstall of `node_modules`.

yjkimjunior:
ah okay 👍

Issue #326, opened 2019-08-22
Add e2e Testing with Cavy to CI


yjkimjunior:
We currently have to have separate unit tests for the JS side and iOS/Android native ends of the bridge, which is a barrier to having a properly comprehensive automated QA. 

A bit of research points to https://cavy.app/ being the holy grail of such testing needs.

hanwencheng:
Integration test is important also for testing the bridges and native modules, I am also looking at the Cavy, find it is promising. I will submit a PR for set up soon.

yjkimjunior:
will log smaller issues for each thing that should be e2e tested

PR #424, opened 2019-10-17
test: Add integration(e2e) test

hanwencheng:
This is the configuration part of #326 which enable us to run CI locally.

For test it locally you need install `detox-cli` as global npm dependency. it has conflict with local `detox` dependency so that we could not add it locally and start with `npx`

```
npm install -g detox-cli
```

After testing with Cavy and Detox, I find Detox is a better choice for us, since we want to develop the test and the origin code base at the same time. And the only complex part of Detox is the configuration part, to use it we do not to use High Order Component, but just with an `testID` on the component. It is more powerful and It is in active development by Wix team.

There are some comparison articles like [Detox vs Appium](https://medium.com/reactive-hub/detox-vs-appium-ui-tests-in-react-native-2d07bf1e244f) or [Detox vs Cavy](https://github.com/pixielabs/cavy/wiki/Detox-comparison)

c37c6e4 : fixes a recursive path problem caused by react-native-camera when build test on iOS, discussion [here](https://github.com/react-native-community/react-native-camera/issues/1407#issuecomment-435702383)

### Test Demo
| iOS       | Android  |
|out|out_android|

### Further steps
After a lot of attempts trying to add the integration test on Travis CI, I keep get the failing test because the build process failed without variant configurations like Rust environment or Android NKD. This will need helps from other people who is experienced with setting Travis CI test environment, and I would like to suggest **now we do the CI test locally, and work on Travis CI integration in another PR**, to prevent blocking current development work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ node_js: 8
 
 cache:
   directories:
-    - node_modules
+  - node_modules
 
 install:
   - yarn
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,44 @@
 ---
-language: node_js
-node_js:
-  - 8
 
-cache:
-  directories:
-  - node_modules
+matrix:
+  include:
+  - language: objective-c
+    osx_image: xcode8.3
 
-install:
-  - yarn
+    branches:
+      only:
+        - master
 
-script:
-  - yarn run lint
-  - yarn run test
+    env:
+      global:
+        - NODE_VERSION=stable
+
+    install:
+      - brew tap wix/brew
+      - brew install applesimutils
+      - curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.2/install.sh | bash
+      - export NVM_DIR="$HOME/.nvm" && [ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh"
+      - nvm install $NODE_VERSION
+      - nvm use $NODE_VERSION
+      - nvm alias default $NODE_VERSION
+
+      - npm install -g react-native-cli
+      - npm install -g detox-cli
+
+    script:
+      - detox build --configuration ios.sim.release
+      - detox test --configuration ios.sim.release --cleanup
+
+  - language: node_js
+    node_js: 8
+
+    cache:
+      directories:
+        - node_modules
+
+    install:
+      - yarn
+
+    script:
+      - yarn run lint
+      - yarn run test
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,23 +11,24 @@ matrix:
 
     env:
       global:
-        - NODE_VERSION=stable
+        - NODE_VERSION = 8
 
     install:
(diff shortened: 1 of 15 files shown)


android/app/build.gradle
```diff
@@ -107,7 +108,7 @@ android {
 
     defaultConfig {
         applicationId "io.parity.signer"
-        minSdkVersion 16
+        minSdkVersion 18
```
hanwencheng: minimal version for Detox

android/build.gradle
```diff
@@ -47,4 +51,12 @@ subprojects {
             }
         }
     }
-}
\ No newline at end of file
+    afterEvaluate {project ->
+        if (project.hasProperty("android")) {
```
hanwencheng: fix the dependency version problem, e.g. in `react-native-secure-store` the compiledSdkVersion is 27, which will leads a compiling error.

.travis.yml
```diff
@@ -13,3 +12,85 @@ install:
...
+#      - gem install xcpretty
+#      - yarn
+#
+#    script:
+#      - detox build -c ios.sim.release
+#      - detox test -c ios.sim.release
+#
+#  - language: android
+##    env:
+##      - REACT_NATIVE_VERSION=0.59.4
```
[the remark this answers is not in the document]
hanwencheng: Ah, yeah, good catch.

hanwencheng:
I have tested it on Emulator, it is also possible to use real device, but need some configuration, basically we have to override device settings with device name, I will make an update on it soon.

> The device should be listed in the output of adb devices command under provided name. Use this type to connect to Genymotion emulator.

hanwencheng:
Tried on real device, but do not has the luck to run it success on device. It seems to be a problem of Detox with support of RN 0.60.x, will track the problem on this issue https://github.com/wix/Detox/issues/1689

yjkimjunior:
@hanwencheng not sure if this is an Xcode version mismatch error, but my binary path is at: `./ios/build/Build/Products/Debug-iphonesimulator/NativeSigner.app`, which led to the error I messaged you about earlier

sjeohp:
> I can't review this PR further since I do not have any emulator installed.
> @yjkimjunior or @sjeohp can you have a look?

Detox build gives me a linker error for ios: `library not found for -lRNSecureStorage`.  Just that lib for some reason, the others are fine.

android/app/build.gradle
```diff
@@ -146,6 +149,8 @@ android {
         release {
             minifyEnabled enableProguardInReleaseBuilds
             proguardFiles getDefaultProguardFile("proguard-android.txt"), "proguard-rules.pro"
+            // Detox-specific additions to pro-guard
+            // proguardFile "${rootProject.projectDir}/../node_modules/detox/android/detox/proguard-rules-app.pro"
```
Tbaut: Should we remove this stray comment?

hanwencheng: These code is needed once ProGuard is Enabled, I will add one line comment above.

PR #428, opened 2019-10-21
fix: #426 (Payload Details When Signing Hash), #427 (Multipart Missed Frames)

### LINKED ISSUE #426: Enhance Multipart scanning UX
When I scan a big multipart QR code my phone might miss some of the codes.
If I scan say a 10 parts code, I miss the 10th:
- I have no idea that I missed it if I don't carefully keep track of the QR codes changes and the number being iterated
- I have no idea which one I missed, and I'll need to remain steady for another full round of QR hoping not to miss it another time
- I have to wait (at least) one other full round (this is on the code generation part)

TL;DR
I'd be nice to know if we missed a QR code, and get some visualization of where we are in the scanning process represented by boxes white (not yet scanned), green (scanned successfully) or red (missed) e.g:
. . :white_check_mark: :white_check_mark: :stop_sign: :white_check_mark: . . .

### LINKED ISSUE #427: setCode is showed as a Message signing
When scanning a multipart QR code for e.g `system>setCode` in apps, Signer shows to the user a hash to sign and doesn't give any hint regarding the method that we call.

We should instead show the usual Tx details view, but with a truncated payload e.g:

You will sign the transaction calling the method **setCode** with the following arguments: 
code: "0x87EF...BEEF"

yjkimjunior:
-- on Signer
* [x] - show the payload prehash image so user knows what's being signed (fixes #427)
* [x] - detect when a frame is missed and alert user what frame number it is. (fixes #426)
-- on @polkadot/react-qr
*  on the generation end, make it possible to skip to a certain frame
https://github.com/polkadot-js/ui/issues/233


yjkimjunior:
hey @Tbaut , could you please review again? I can't reproduce the second issue, but I suspect it might have been a side effect of the off by one error, which is fixed now

yjkimjunior:
ezgif-6-31b82f34c100

yjkimjunior:
signing hash should work now, @Tbaut

src/util/decoders.spec.js
```diff
@@ -97,7 +104,11 @@ const SIGN_TX_TEST = u8aConcat(
 	new GenericExtrinsicPayload(SIGNER_PAYLOAD_TEST, { version: 4 }).toU8a()
 );
 
-describe.skip('sanity check', () => {
+describe.only('sanity check', () => {
```
hanwencheng: this will be removed before merging, right?

src/util/decoders.spec.js
```diff
@@ -97,7 +104,11 @@ const SIGN_TX_TEST = u8aConcat(
...
+	it.only('blake2 hasher works', () => {
```
hanwencheng: and `only` here

src/stores/ScannerStore.js
```diff
@@ -390,16 +450,22 @@ export default class ScannerStore extends Container<ScannerState> {
 	}
 
 	cleanup() {
-		this.setState(defaultState);
+		return new Promise(resolve => {
```
hanwencheng: Why this is wrapped into `Promise`? there are all synchronous functions I think.

yjkimjunior: this is a workaround to an edge case Thibaut discovered in the second part of this: https://github.com/paritytech/parity-signer/pull/428#pullrequestreview-305117130

It is to make sure the state is set before the next frame is scanned. In this case the setState is asynchronous.

yjkimjunior: actually in all cases `setState` is asynchronous, it just usually doesn't matter.

hanwencheng: Ah, yeah, set state return a promise here, you can do `await this.setStatte`, which is the same I think.

hanwencheng:
IMG_0040

Tested functionality with 4 random appeared frame on iOS with sudo `system.setRemark`

yjkimjunior:
Thanks for the review, I will log those as UX/UI enhancement requests, to be added on top of this so that this PR can remain in bulk about the core issues.

yjkimjunior:
@Tbaut I really could not reproduce the "wrong pin" issue, I'm able to sign and send on Kusama and Substrate Dev both no problem. Did you build the Rust code again after cleaning?

src/stores/ScannerStore.js
```diff
@@ -286,6 +282,10 @@ export default class ScannerStore extends Container<ScannerState> {
 				multipartData: nextDataState
 			});
 		}
+
```
hanwencheng: could be moved together with line 281. :p

yjkimjunior: that's in an else block, I want to update completedFramesCount every frame

Issue #407, opened 2019-10-07
Manage metadata screen


Tbaut:
This page should be available from the top level of the app.
It should:

- show the default networks metadata (Kusama)
- show the added networks metadata with the ability to delete them (no edit for now)
- show the metadata version
- show a "add metadata" button similar to the "Scan" button (bringing to the scanner)

yjkimjunior:
this answers a question of how to make clear the intention to scan metadata as opposed to any other random message clear to the scanner.

on the reading end:
* [ ] - save the current metadata for Kusama, Substrate in `db`
* [ ] - load the metadata from `db`, should be able to query it along with the rest of the `networkSpec`s
* [ ] - button to scan metadata will pass that as a flag in props so that ScannerStore will know not to handle it as a regular message (in which case it would just hash it and try to sign it
* [ ] - construct that Metadata from the concatenated blob
* [ ] - register types
* [ ] - PayloadDetails use the appropriate type registry depending on the `genesisHash` provided in the payload.
  * currently we have a single type registry instance that is injected only once. There could be conflicts and overrwrites now with multiple networks, so need make sure to reset that each time a new payload is scanned.

On the generation end:
* [ ] - UI for easily getting a well formatted QR for Signer to be able to decode it. Input: ws endpoint. Output: QR. In between would call state_getMetadata, strip everything but calls from it, hex encode, then feed the result to @polkadot/react-qr.
* [ ] - UI for browsing supported QRs.

Tbaut:
One thing that is missing here and I'm not sure it's easy to add securely: 
We want to sign the metadata with a "Parity" private key on the QR generation side, and verify this signature on Signer side with a trusted public key.

Fetching metadata and feeding the blob to @polkadot/react-qr doesn't have this step and is therefore quite unsecure if a malicious endpoint is provided or if the node is attacked. 

Your solution has the great advantage of being dynamic, we could eventually modify it to add a signing part that we could do manually each time there's a new network/metadata.

Slesarew:
Not a screen, but this functionality is added.

PR #429, opened 2019-10-22
feat (#407): Dynamic network specs + metadata management

yjkimjunior:
In NetworkSettings, user should be able to view networks for which he/she has already the full networkSpec.

A network spec will remain the same as it currently is in `constants.js`

The calls-only metadata blob will be persisted with AsyncStorage. It will be keyed via networkKey, just like the network spec. So the networkKey, i.e. genesisHash is essentially just a primary key.

The separation is ideal as the metadata (I'd guesstimate most chains would have something averaging around 32kb) should only be queried at the point of completing the scanning and validation of a QR payload. Otherwise, it'd be very very slow.

This should also allow for easily building the HDKD refactor on top of without any big conflicts.

I will be up to whomever supplies the Network Spec QR payload to make sure it is well formatted. Which is where https://github.com/yjkimjunior/Substrate-Metadata-Qrs aims to mitigate friction. Particularly, the `identiconFn` supplied needs to meet the same format as https://github.com/polkadot-js/ui/blob/f2f36e2db07f5faec14ee43cf4295f5e8a6f3cfa/packages/reactnative-identicon/src/icons/Polkadot.tsx#L37

on the reading end:
* [x] - save the current metadata for Kusama, Substrate in `db`
* [x] - load the metadata from `db`, should be able to query it along with the rest of the `networkSpec`s
* [x] - button to scan metadata will pass that as a flag in props so that ScannerStore will know not to handle it as a regular message (in which case it would just hash it and try to sign it
* [x] - construct that Metadata from the concatenated blob
* [x] - register types
* [x] - PayloadDetails use the appropriate type registry depending on the `genesisHash` provided in the payload.
* [x] - Essentially all the Substrate stuff currently in `constants.js` will have to be moved to `db.js`. 
* currently we have a single type registry instance that is injected only once. There could be conflicts and overrwrites now with multiple networks, so need make sure to reset that each time a new payload is scanned.

On the generation end: https://metadata-qrs.herokuapp.com
* [x] - UI for easily getting a well formatted QR for Signer to be able to decode it. Input: ws endpoint. Output: QR. In between would call state_getMetadata, strip everything but calls from it, hex encode, then feed the result to @polkadot/react-qr.
* [x] - UI for browsing supported QRs.


yjkimjunior:
main functionality works, need to rebase on HDKD branch

hanwencheng:
would be happy to review it if conflicts solved.

yjkimjunior:
i'll wrap up this weekend

hanwencheng:
closed as included in #681

PR #431, opened 2019-10-23
major: HDKD feature

### LINKED ISSUE #311: Disable instant pin check, and add maximum input time.
It would be securer to disable the instant check of the pin code, and add a limit for trying to input the pin. For example, once user click more than 5 times with the wrong pin, then the app will reset the `secure store` with the private keys.

It is convenient without a final click, but I do not think it is secure enough.

out mp4

### LINKED ISSUE #315: Replace StackNavigator with SwitchNavigator
### Problem:
It seems before the v2.1 bump that we have been incorrectly using a StackNavigator for navigation, which was fine but with the latest version of `react-navigation` we had to work around it by setting key to undefined when resetting the route.
```
StackActions.reset({
                index: 1,
                key: undefined, // FIXME
                actions: [
                  NavigationActions.navigate({ routeName: 'AccountList' }),
                  NavigationActions.navigate({ routeName: next })
                ]
              });
```

### Solution
The correct solution is to refactor our navigator to use a SwitchNavigator instead. https://github.com/react-navigation/react-navigation/issues/1127#issuecomment-295841343

IIRC I used the workaround as the refactor would be potentially quite hairy and that particular bump PR was already big and hairy enough.

### LINKED ISSUE #324: UI tells me pin is wrong before I finish entering it
1. Create an account with a more than 6 digit pin
2. Change pin
3. 'Wrong pin, please try again' after 6 digits entered, whether correct or not

The app shouldn't assume a pin length.  Instead there should be a button for the user to press once they have finished entering the pin.

### LINKED ISSUE #404: Feat: Move to root phrase HDKD model of accounts

### LINKED ISSUE #432: 'use strict' everywhere
minor annoyance, it may require a tiny bit of refactoring. Same with using `flow` everywhere. If we're going to have it we should use it everywhere or nowhere, IMHO (personal preference to using it everywhere).

Of course, all these can be addressed broadly by migrating to Typescript #418.

### LINKED ISSUE #443: E2E - Accounts Management
This is a fairly easy flow (excluding the derivations for now) that doesn't change all that often so should be a good one to start with.

Below are the main cases, missing many edge cases to fill in as we go:

1. Expect account list to be empty
2. Go to create (for each network)
  a. create an account
  b. recover an account
3. Go back to accounts list
  a. click on an account
  b. expect qr to render
  c. expect public key to actually be valid
  d. edit works
  e. delete works

### LINKED ISSUE #448: Fix Lint warnings
At the time of writing:
- parity-signer/src/components/PayloadDetailsCard.js
  159:39  warning  'methodArgs' is already declared in the upper scope  no-shadow

- parity-signer/src/screens/AccountNew.js
  151:22  warning  'name' is already declared in the upper scope                   no-shadow
  158:10  warning  'derivationPassword' is already declared in the upper scope     no-shadow
  159:10  warning  'derivationPath' is already declared in the upper scope         no-shadow
  160:10  warning  'isDerivationPathValid' is already declared in the upper scope  no-shadow

- parity-signer/src/screens/AccountRecover.js
  214:21  warning  'name' is already declared in the upper scope                   no-shadow
  232:21  warning  'seedPhrase' is already declared in the upper scope             no-shadow
  245:9   warning  'derivationPassword' is already declared in the upper scope     no-shadow
  246:9   warning  'derivationPath' is already declared in the upper scope         no-shadow
  247:9   warning  'isDerivationPathValid' is already declared in the upper scope  no-shadow

- parity-signer/src/stores/ScannerStore.js
  260:11  warning  'missedFrames' is already declared in the upper scope  no-shadow

- parity-signer/src/util/decoders.js
  297:25  warning  Unexpected use of '>>'  no-bitwise
  297:37  warning  Unexpected use of '&'   no-bitwise

- parity-signer/src/util/decoders.spec.js
  100:1  warning  Skipped test suite  jest/no-disabled-tests

- parity-signer/src/util/strings.js
  23:9  warning  Unexpected control character(s) in regular expression: \x00  no-control-regex

hanwencheng:
So I am thinking how to collaborate together with @goldsteinsveta on this issue, we may work and track on this branch and record the progress here.

### Design

References to design: https://www.figma.com/file/4odbGhaJ3OTN2ZQb52uvIy/Signer_UI-2.0?node-id=1760%3A148

### Data Schemas

The accountId with Ethereum saved into storage is now with Uppercase. But we still recognize QR code with lower case account.

```javascript
const networks = [
  {
    name: 'kusama',
    genesisHash: "0xb28bd355a4cd1df46bc56a2949bce8feb84eebc864f1c3f1f77668bd3b6559b3",
    prefix: 2,
    decimals: 12,
    pathID: 'kusama',
    title: 'Kusama CC2',
    unit: 'KSM',
    color: '#e6007a',
    protocol: 'substrate'
  },
  {
    color: '#64A2F4',
    ethereumChainId: '1',
    secondaryColor: colors.card_bgSolid,
    title: 'Ethereum Frontier',
    protocol: 'ethereum'
  }
];


const identities = [
  {
    encryptedSeed: '{"cipher":"aes-128-ctr","cipherparams":{"iv":e872394210984322432423,"cipherText":"a352dfg3g4245..."}}',
    //NOTICE: meta is a Map Object
    meta: {
      '//kusma//funding/1': {
        name: 'funding account1',
        accountId: "substrate:DwGa4E65nbGafaGirjMH5kXbtT7jHuAcSXHLLkTePnqw95p:0xe3777fa922cafbff200cadeaea1a76bd7898ad5b89f7848999058b50e715f636",
        createdAt: 1571068850409,
        updatedAt: 1571078850509,
      },
      '//kusma//funding/2': {
        accountId: "substrate:EoXsgmP36WXkyCamoC9iHy7Gr2tZpEykz8rmFw3bAS1nArV:0xe3777fa922cafbff200cadeaea1a76bd7898ad5b89f7848999058b50e715f636",
        name: 'funding account2',
        createdAt: 1571068850409,
        updatedAt: 1571078850509,
      },
      '1': {
        name: 'ethereum frontier account',
        accountId: "ethereum:0x06Fc73F6797Ed62f46A4E6A13dD9366C5f2725D7@42",
        createdAt: 1571068850409,
        updatedAt: 1571078850509,
      }
    },
    //NOTICE: accountIds is a Map Object
    accountIds: {
      'substrate:DwGa4E65nbGafaGirjMH5kXbtT7jHuAcSXHLLkTePnqw95p:0xe3777fa922cafbff200cadeaea1a76bd7898ad5b89f7848999058b50e715f636': '//kusama//funding/1',
      'substrate:EoXsgmP36WXkyCamoC9iHy7Gr2tZpEykz8rmFw3bAS1nArV:0xe3777fa922cafbff200cadeaea1a76bd7898ad5b89f7848999058b50e715f636': '//kusama//funding/2',
      'ethereum:0x06Fc73F6797Ed62f46A4E6A13dD9366C5f2725D7@42': '1',
    }
  },
  {
    encryptedSeed: '{"cipher":"aes-128-ctr","cipherparams":{"iv":f872394210984322432423,"cipherText":"sad12343fdcdxa..."}}',
    //NOTICE: meta is a Map Object
    meta: {
      '//kusma//funding/1': {
        name: 'funding account1',
        accountId: "substrate:FJBQ3tv6w3qhYkhKWnxfdjPwhvCd879oUeQKt6u8bHyhqzX:0xe3777fa922cafbff200cadeaea1a76bd7898ad5b89f7848999058b50e715f636",
        createdAt: 1571068850409,
        updatedAt: 1571078850509,
      },
    },
    //NOTICE: accountIds is a Map Object
    accountIds: {
      "substrate:FJBQ3tv6w3qhYkhKWnxfdjPwhvCd879oUeQKt6u8bHyhqzX:0xe3777fa922cafbff200cadeaea1a76bd7898ad5b89f7848999058b50e715f636": '//kusama//funding/1',
    }
  }
];
```

### Seed Unlock Refractor

Also include the refactoring of unlocking logic, the setPin and unLock will become a function to call the Pin input screen like the following:

```javascript
//create pin or recover pin
<Button
	title="Next"
	onPress={async () => {
		const pin = await setPin(navigation);
		//then navigate to other screen
	}}
/>

//sign with pin
<Button
	title="Sign"
	onPress={async () => {
		const seed = await unLock(navigation, data);
		//use the seed to sign data
	}}
/>
```

With new pin code input screen, the pin code input mode will be changed.
Also closes #311, closes #324, closes #448, closes #315 , closes #432 , closes #443, closes #404  

#### Manual

* Sign extrinsics with Legacy Substrate account 
* Sign transaction with Legacy Ethereum account
* Sign messages with Legacy Ethereum account
* Sign extrinsics with New Substrate account 
* Sign transaction with New Ethereum account
* Sign messages with New Ethereum account
* Identity Creation
* Ethereum keypairs creation
* Path Derivation
* Path deletion
* Ethereum keypairs deletion
* Renaming

#### Automat
* E2E Android: `yarn e2e:android`
* E2E iOS: `yarn e2e:ios`
* Unit Jest: `yarn test`
* Unit Rust `yarn test-rust`

### Not 100% Sure Thing
dcd7477274bb7c63388491ba2c4c2c1b60869079 : Reset `preHash` in `cleanup()` function in ScannerStore. 

### Problem Tracking

| component/screen Name | type| creation | styling | data-binding |   |
|-----------------------|--|--------|---------|--------------|---|
| LegacyAccountList  | Screen | :white_check_mark:   | :white_check_mark:  | :white_check_mark:   |  
| LegacyAccountBackupScreen  | Screen | :white_check_mark: | :white_check_mark:  | :white_check_mark:  | 
| IdenititesSwitch    | Component |  :white_check_mark:    |   :white_check_mark:   | :white_check_mark: |  
| IdentityNew (Include Recover) | Screen | :white_check_mark: | :white_check_mark:  | :white_check_mark:  | 
| IdentityBackup  | Screen | :white_check_mark: | :white_check_mark:  | :white_check_mark:  | 
| IdentityPin  | Screen | :white_check_mark: | :white_check_mark:  | :white_check_mark:  | 
| IdentityManagement  | Screen | :white_check_mark: | :white_check_mark:  | :white_check_mark:  |
| PathDerivation  | Screen | :white_check_mark: | :white_check_mark:  | :white_check_mark:  |  
| PathsList  | Screen | :white_check_mark: | :white_check_mark:  | :white_check_mark:  | 
| PathDetails  | Screen | :white_check_mark: | :white_check_mark:  | :white_check_mark:  | 
| PathManagement  | Screen | :white_check_mark: | :white_check_mark:  | :white_check_mark:  | 
| PathCard  | Component | :white_check_mark: | :white_check_mark:  | :white_check_mark:  |


src/stores/AccountsStore.js
```diff
@@ -41,7 +42,22 @@ export type Account = {
...
+type AccountMeta = {
+	address: string,
+	createdAt: number,
+	name: string,
+	updatedAt: number
+};
+
+type Identity = {
+	encryptedSeedPhrase: string,
+	derivationPassword: string,
```
yjkimjunior: derivation password is out of scope for v4, no?

yjkimjunior: cool, just double checking.

src/components/SecurityHeader.js
```diff
@@ -48,30 +46,23 @@ class SecurityHeader extends React.Component {
...
+					paddingRight: 14
+				}}
+			>
+				<IdentitiesSwitch />
+				<ButtonIcon
+					onPress={() => this.props.navigation.navigate('Security')}
+					iconName="security"
+					iconColor={colors.bg_alert}
+				/>
+				<ButtonIcon iconType="ionicon" iconName="md-more" />
```
hanwencheng: if here use `onPress={()=> {}}`, then maybe we can keep the `props: onPress` as `isRequired`, so later we wont't forget to add any onPress button for the ButtonIcon, I suppose it will always come with an `onPress` function?

goldsteinsveta: sure, done in [205fbdc](https://github.com/paritytech/parity-signer/pull/431/commits/205fbdcef81c5df61941a34702bf10e2ccbb5bf5)

burdges:
Just fyi, we'll eventually tweak the HDKD interfaces to fix https://github.com/paritytech/substrate/issues/3396 see https://github.com/w3f/schnorrkel/issues/45 but no need to worry about it right now

.eslintrc.js
```diff
@@ -14,6 +14,7 @@ module.exports = {
     },
   },
   rules: {
+  	"no-bitwise": "off",
```
yjkimjunior: nit: indentation

src/stores/AccountsStore.js
```diff
@@ -220,4 +234,132 @@ export default class AccountsStore extends Container<AccountsState> {
...
+		await this.setState({ currentIdentity: identity });
+	}
+
+	updateNewIdentity(identityUpdate) {
+		this.setState({
+			newIdentity: { ...this.state.newIdentity, ...identityUpdate }
+		});
+	}
+
+	async updateCurrentToIdentities() {
```
yjkimjunior: To enforce better developer experience for all of us and for the future contributors, let's all aim to be more thorough with inline comments and JSDoc. 

This function name is not particularly clear at a glance what it does. Could you please add a/some comments and maybe make the function name more specific?

src/screens/PathsList.js
```diff
@@ -0,0 +1,77 @@
...
+// Parity is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+import React from 'react';
+import { ScrollView, Text, View } from 'react-native';
```
yjkimjunior: nit: new line between absolute and relative imports

src/screens/PathManagement.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { withAccountStore } from '../util/HOC';
```
yjkimjunior: nit: new line between absolute and relative imports

src/screens/PathDerivation.js
```diff
@@ -0,0 +1,92 @@
...
+// Parity is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+import React, { useState } from 'react';
+import { withNavigation } from 'react-navigation';
```
yjkimjunior: nit: new line between absolute and relative imports

src/screens/IdentityNew.js
```diff
@@ -0,0 +1,140 @@
...
+} from '../util/navigationHelpers';
+import { alertIdentityCreationError } from '../util/alertUtils';
+
+function IdentityNew({ accounts, navigation }) {
+	const [isRecover, setIsRecover] = useState(false);
+	const [seedPhrase, setSeedPhrase] = useState('');
+
+	useEffect(() => {
+		accounts.updateNewIdentity(emptyIdentity());
+		return function() {
```
yjkimjunior: an arrow function would also work here

src/screens/IdentityNew.js
```diff
@@ -0,0 +1,140 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Parity.  If not, see <http://www.gnu.org/licenses/>.
+
+import React, { useEffect, useState } from 'react';
+import { StyleSheet, View, Text } from 'react-native';
+import { withNavigation } from 'react-navigation';
```
yjkimjunior: nit: new line between absolute and relative imports

src/screens/IdentityManagement.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
```
yjkimjunior: nit: new line between absolute and relative imports

src/screens/AccountNew.js
```diff
@@ -16,191 +16,175 @@
...
-
-export default class AccountNew extends React.Component {
-	static navigationOptions = {
-		headerBackTitle: 'Back',
-		title: 'New Account'
+import AccountCard from '../components/AccountCard';
+import { withAccountStore } from '../util/HOC';
+
+/**
+ * @return {null}
```
yjkimjunior: I'm assuming this is a placeholder for now :)

src/screens/AccountNetworkChooser.js
```diff
@@ -18,115 +18,84 @@
 
 import React from 'react';
 import { ScrollView, StyleSheet, Text } from 'react-native';
-import { Subscribe } from 'unstated';
+import { withNavigation } from 'react-navigation';
```
yjkimjunior: nit: new line between absolute and relative imports

src/components/IdentitiesSwitch.js
```diff
@@ -0,0 +1,223 @@
...
+
+	const renderCurrentIdentityCard = () => {
+		if (!currentIdentity) return;
+		const currentIdentityTitle = getIdentityName(currentIdentity, identities);
+
+		return (
+			<>
+				<ButtonIcon
+					title={currentIdentityTitle}
+					onPress={async () => {
```
yjkimjunior: please extract to a method

src/components/AccountIcon.js
```diff
@@ -49,17 +50,32 @@ export default function AccountIcon(props) {
 			.catch(console.error);
 	};
 
+	if (address === '') {
```
yjkimjunior: It's probably better to change the address proptype to be potentially undefined and do a null check. but that's just bike shedding.

.eslintrc.js
hanwencheng: Will use `lint:fix` to auto fix these erros by merging the master brach

src/stores/AccountsStore.js
hanwencheng: Agree, actually I prefer to use explicit function name with minimal comments.

src/screens/PathsList.js
hanwencheng: I am thinking to use [eslint-plugin-import](https://github.com/benmosher/eslint-plugin-import) for doing that, not pretty sure if it will work

src/components/AccountIcon.js
hanwencheng: In places like `emptyAccount()` we have the default value of address to empty string. And In React if we render address in `<Text>`, an undefined error would appear. This is the main concern to use empty string like this situation as the default value for the string props.

hanwencheng:
@YJ, except the last two points, the others are addressed.

I am thinking we will implement the bottom bar in the future PR. So make it work as soon as possible.

And according to the discussion yesterday, I will using AccountNetworkChooser as the main screen for now (with the first identity in the account store).

src/stores/ScannerStore.js
```diff
@@ -451,11 +455,9 @@ export default class ScannerStore extends Container<ScannerState> {
 
 	cleanup() {
 		return new Promise(resolve => {
-			const prehash = this.state.prehash;
 			this.setState(
 				{
-					...DEFAULT_STATE,
-					prehash
```
hanwencheng: not sure why `prehash` is here, delete here, would like to get more context

src/screens/AccountNetworkChooser.js
```diff
@@ -16,118 +16,241 @@
...
+import Separator from '../components/Separator';
+import fontStyles from '../fontStyles';
 
-export default class AccountNetworkChooser extends React.PureComponent {
-	static navigationOptions = {
-		headerBackTitle: 'Back',
-		title: 'Choose a network'
-	};
-	render() {
+function AccountNetworkChooser({ navigation, accounts }) {
```
yjkimjunior: on start, with clean build, getting a undefined error:

Screenshot 2019-11-18 at 11 03 16

src/screens/AccountNew.js
```diff
@@ -16,191 +16,174 @@
...
+					</Text>
+					<Button
+						buttonStyles={styles.nextStep}
+						title="Next Step"
+						disabled={
+							!validateSeed(seed, validBip39Seed).valid ||
+							!isDerivationPathValid
+						}
+						onPress={() => {
+							validateSeed(seed, validBip39Seed).valid &&
```
yjkimjunior: if `validateSeed()` returns false, the button will be disabled anyway, so could remove this check.

src/screens/IdentityManagement.js
```diff
@@ -0,0 +1,109 @@
...
+						}
+					]}
+				/>
+			</View>
+			<TextInput
+				label="Display Name"
+				onChangeText={name => accounts.updateIdentityName(name)}
+				value={currentIdentity.name}
+				placeholder="Enter a new identity name"
+				focus={true}
```
yjkimjunior:
```suggestion
				focus
```

src/screens/IdentityPin.js
```diff
@@ -0,0 +1,215 @@
...
+	const submit = async () => {
+		const isIdentityCreation = navigation.getParam('isNew');
+		const { pin, confirmation } = state;
+		if (pin.length >= 6 && pin === confirmation) {
+			if (isIdentityCreation) {
+				const resolve = navigation.getParam('resolve');
+				setState(initialState);
+				resolve(pin);
+			} else {
+				// await accounts.save(accounts.getSelectedKey(), account, pin);
```
yjkimjunior: can we remove this else block?

src/screens/IdentityPin.js
```diff
@@ -0,0 +1,215 @@
...
+		} else {
+			updateState({ pinTooShort: true });
+		}
+	};
+
+	const showHintOrError = () => {
+		if (state.pinTooShort) {
+			return ' Your pin must be at least 6 digits long!';
+		} else if (state.pinMismatch) {
+			return " Pin codes don't match!";
```
yjkimjunior: nit single quotes

src/screens/IdentityPin.js
```diff
@@ -0,0 +1,215 @@
...
+		}
+		return ' Choose a PIN code with 6 or more digits';
+	};
+
+	const onPinInputChange = (stateName, pinInput) => {
+		if (/^\d+$|^$/.test(pinInput)) {
```
yjkimjunior: small suggestion: the `/^\d+$|^$/.test` could be put as an export from `util/numbers`

src/screens/LegacyAccountBackup.js
```diff
@@ -0,0 +1,160 @@
...
+import TouchableItem from '../components/TouchableItem';
+import DerivationPasswordVerify from '../components/DerivationPasswordVerify';
+import AccountsStore from '../stores/AccountsStore';
+import { NetworkProtocols, NETWORK_LIST } from '../constants';
+import { alertBackupDone, alertCopyBackupPhrase } from '../util/alertUtils';
+
+export default class LegacyAccountBackup extends React.PureComponent {
+	render() {
+		return (
+			<Subscribe to={[AccountsStore]}>
```
yjkimjunior: why not use the HOC?

src/screens/LegacyAccountBackup.js
```diff
@@ -0,0 +1,160 @@
...
+function LegacyAccountBackupView(props) {
+	useEffect(() => {
+		const handleAppStateChange = nextAppState => {
+			if (nextAppState === 'inactive') {
+				props.navigation.goBack();
+			}
+		};
+
+		AppState.addEventListener('change', handleAppStateChange);
+		return function() {
```
yjkimjunior:
```suggestion
		return () => {
```

src/screens/LegacyAccountBackup.js
```diff
@@ -0,0 +1,160 @@
...
+	const {
+		address,
+		derivationPassword,
+		derivationPath,
+		name,
+		networkKey,
+		seed,
+		seedPhrase
+	} = isNew ? accounts.getNew() : accounts.getSelected();
+	const protocol =
```
yjkimjunior: hmm this could be bad if protocol gets set to empty string. Could we throw an error and `<Alert />` for that case?

src/screens/LegacyAccountList.js
```diff
@@ -0,0 +1,107 @@
...
+import AccountCard from '../components/AccountCard';
+import Background from '../components/Background';
+import AccountsStore from '../stores/AccountsStore';
+import testIDs from '../../e2e/testIDs';
+import ButtonMainAction from '../components/ButtonMainAction';
+
+export default class LegacyAccountList extends React.PureComponent {
+	render() {
+		return (
+			<Subscribe to={[AccountsStore]}>
```
yjkimjunior: why not use the HOC?

src/screens/AccountNetworkChooser.js
hanwencheng: Do you have installed the app in this branch before?

src/screens/AccountNew.js
hanwencheng: resolved

src/screens/IdentityManagement.js
hanwencheng: resolved

src/screens/IdentityPin.js
hanwencheng: resolved

src/screens/IdentityPin.js
hanwencheng: There is single quote inside, change to ```

src/screens/IdentityPin.js
hanwencheng: create a `utils/regex.js` file to order the regexs

src/screens/LegacyAccountBackup.js
hanwencheng: Resolved, I have not check that, HOC is not 100% covered now.

src/screens/LegacyAccountBackup.js
hanwencheng: The reason to use arrow function is to avoid `this`, but I have not replace all the `function` syntax with arrow function. In this case they are the same,  is there any specific reasons for changing to arrow functions here?

src/screens/LegacyAccountBackup.js
hanwencheng: Good catch! fallback to `NetworkProtocol.UNKNOW`

src/screens/LegacyAccountList.js
hanwencheng: resolved

src/screens/LegacyAccountBackup.js
hanwencheng: If we us HOC, most times `withNavigation` is used together. In this case need to make sure that the the wrapped component is not a Class but a function (`withNavigation` need that). Hence more work and test is needed, for 100% coverage and use HOC everywhere I suggest we could do that in the future PR.

hanwencheng:
> When I'm asked my pin to unlock my identity and enter a wrong pin it says "password don't match"

Good Catch, resolved.

> Sending a Tx on (on any network, Kusama cc2/ Substrate dev), the sender card says "no name"
although I do have a name for this account
Could you please be more specific, does the account have name before, because if you do not set a name for it, it will display a dummy name which is the combination of its path. I tried test with both legacy account and account generated by seed, and both tx signing and message signing, it works. Extrinsic I use is `sudo: sudo, system: setRemarks`;

> Prob. for @goldsteinsveta, I find the method of the extrinsic (e.g balance.transfer) to be lost in the middle of the text. Although it's super important to make it stand out, as a user I should see very quickly what I will do if I sign this extrinsic). It used to be bold, I'd go as far as put in bold, bigger, another color... It really needs to stand out.

I did the styling on the method part, and I agree that the method is hard to see,  now I will put it in the top of the content list, and let @goldsteinsveta decide the styling.

My commits added some dropdown to the identity switch, but it's gone for some reason :( this allowed ppl to click on a little icon on the right of an identity to manage/backup identity. I'll add it back.
> It is disabled by `dropdown={true}` on this [line](https://github.com/paritytech/parity-signer/blob/sveta-hanwen-hdkd/src/components/IdentitiesSwitch.js#L152)

src/screens/LegacyAccountBackup.js
yjkimjunior: yeah in this case it doesn't make a difference, I don't mind either way.

src/screens/LegacyAccountBackup.js
yjkimjunior: sounds good to me

hanwencheng:
In `TxDetails` `SignedTx` `MessageDetails` `SignedMessage` please use `CompatibleCard` instead of `AccountCard`, for backward compatibility with legacy account.

specs/util/identitiesUtils.spec.js
```diff
@@ -0,0 +1,179 @@
...
+				paths: [paths[0]],
+				title: '//default'
+			},
+			{
+				paths: [paths[2]],
+				title: '/softKey1'
+			},
+			{
+				paths: [paths[4]],
+				title: '//stacking'
```
yjkimjunior:
```suggestion
				title: '//staking'
```

specs/util/identitiesUtils.spec.js
```diff
@@ -0,0 +1,179 @@
...
+		]);
+	});
+
+	it('get the path name', () => {
+		const expectNames = [
+			'default',
+			'funding account1',
+			'softKey1',
+			'funding2',
+			'stacking1',
```
yjkimjunior:
```suggestion
			'staking1',
```

src/components/AccountIcon.js
```diff
@@ -49,17 +50,45 @@ export default function AccountIcon(props) {
...
+						backgroundColor: colors.bg_text_sec,
+						borderRadius: 40,
+						height: 40,
+						width: 40
+					}}
+				/>
+			</View>
+		);
+	}
+	if (address === 'new') {
```
yjkimjunior: what is the reasoning for setting meta-info about an address in the address itself (L74 as well)? I guess this typechecks but it's quite confusing. A separate prop would be appropriate here IMHO.

src/components/DerivationPathField.js
```diff
@@ -36,7 +38,7 @@ export default function DerivationPathField(props) {
 		<>
 			<TouchableOpacity
 				onPress={toggleShowAdvancedField}
-				style={{ diplay: 'flex' }}
+				//style={{ diplay: 'flex' }}
```
yjkimjunior:
```suggestion
				style={{ display: 'flex' }}
```

src/components/IdentitiesSwitch.js
```diff
@@ -0,0 +1,311 @@
...
+		await accounts.selectIdentity(identity);
+		navigation.navigate('AccountNetworkChooser');
+	};
+
+	const renderIdentityOptions = identity => {
+		return (
+			<>
+				<ButtonIcon
+					title="Manage Identity"
+					onPress={async () => {
```
yjkimjunior: please pull this out into its own handler function

src/components/IdentitiesSwitch.js
```diff
@@ -0,0 +1,311 @@
...
+					iconType="antdesign"
+					iconName="arrowright"
+					iconSize={18}
+					testID={testIDs.IdentitiesSwitch.manageIdentityButton}
+					textStyle={fontStyles.t_regular}
+					style={styles.i_arrowStyle}
+				/>
+				<ButtonIcon
+					title="Show Recovery Phrase"
+					onPress={async () => {
```
yjkimjunior: same as above

src/components/IdentitiesSwitch.js
```diff
@@ -0,0 +1,311 @@
...
+		);
+	};
+
+	const renderSettings = () => {
+		return (
+			<>
+				<ButtonIcon
+					title="About"
+					onPress={() => {
+						setVisible(false);
```
yjkimjunior: same as above

src/components/IdentitiesSwitch.js
```diff
@@ -0,0 +1,311 @@
...
+						navigation.navigate('About');
+						// go to Settings;
```
yjkimjunior: stray comment? :)
comment says go to Settings but navigate says to About?

src/components/IdentitiesSwitch.js
```diff
@@ -0,0 +1,311 @@
...
+					iconBgStyle={styles.i_arrowBg}
+					iconType="antdesign"
+					iconName="arrowright"
+					iconSize={18}
+					textStyle={fontStyles.t_regular}
+					style={styles.i_arrowStyle}
+				/>
+				<ButtonIcon
+					title="Privacy Policy"
+					onPress={() => {
```
yjkimjunior: same as above

src/components/IdentitiesSwitch.js
```diff
@@ -0,0 +1,311 @@
...
+										title="Legacy Accounts"
+										onPress={() => {
+											setVisible(false);
+											navigation.navigate('LegacyAccountList');
+										}}
+										iconName="solution1"
+										iconType="antdesign"
+										iconSize={24}
+										textStyle={fontStyles.t_big}
+										style={{ paddingLeft: 8 * 4 }}
```
yjkimjunior: = 32?

src/screens/AccountEdit.js
```diff
@@ -45,21 +40,17 @@ export default class AccountEdit extends React.PureComponent {
...
 								title={selected.name}
-								address={selected.address}
+								accountId={selected.address}
 								networkKey={selected.networkKey}
 							/>
-							<Text style={styles.title}>ACCOUNT NAME</Text>
 							<TextInput
+								label="Account Name"
 								style={{ marginBottom: 40 }}
 								onChangeText={async name => {
```
yjkimjunior: please bring this out into its own handler funciton

src/screens/AccountNetworkChooser.js
```diff
@@ -16,118 +16,242 @@
...
+			{renderScreenHeading()}
+			<ScrollView>
+				{NetworkList.map(([networkKey, networkParams], index) => (
+					<AccountCard
+						isNetworkCard={true}
+						accountId={''}
+						key={networkKey}
+						testID={testIDs.AccountNetworkChooser.networkButton + index}
+						networkKey={networkKey}
+						onPress={async () => {
```
yjkimjunior: same as above

src/screens/AccountNew.js
yjkimjunior: bump

src/screens/IdentityPin.js
```diff
@@ -0,0 +1,206 @@
...
+			extraHeight={120}
+			testID={testIDs.IdentityPin.scrollScreen}
+		>
+			<Background />
+			{renderPinInput()}
+		</KeyboardScrollView>
+	);
+}
+
+class PinInput extends Component {
```
yjkimjunior: suggestion: make this a PureComponent or functional component

.eslintrc.js
```diff
@@ -22,10 +22,12 @@ module.exports = {
...
     "comma-dangle": ["error", "never"],
     "object-curly-spacing": ["error", "always"],
     "quotes": ["error", "single",  { "avoidEscape": true }],
     "no-unused-vars": ["error", { "args": "none" }],
+    "react-hooks/exhaustive-deps": "off",
```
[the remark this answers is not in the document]
hanwencheng: Disable this link because most of time we use it as componentDidMounted and cleanup when componentWillUnmounted, so I do 
```javascript
useEffect(()=> {
  //do something
}, []) //empty array here!
```

but yarn will always give an error says I need to add dependent variable here.
And if you turn it on and run `yarn lint:fix`, you will see lint will add all the related variables into dependent array, which is not the case we want to use it.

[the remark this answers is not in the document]
yjkimjunior: I don't understand what you mean. In most cases you really shouldn't just ignore your dependent variables for convenience. Eslint is there for a reason. Could you point to me a specific example that justifies this?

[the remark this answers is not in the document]
hanwencheng: I look back and through the link you posted, I think I really previously have some misunderstanding about `useEffect`. And this [commnet](https://github.com/facebook/react/issues/14920#issuecomment-470945051) is especially useful. 
Enable the lint again.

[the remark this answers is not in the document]
yjkimjunior: cool cool

src/components/DerivationPathField.js
hanwencheng: I think `display:flex` does not work in react native, so just delete it.

src/components/IdentitiesSwitch.js
hanwencheng: you are so right 😀

e2e/e2eUtils.js
```diff
@@ -0,0 +1,62 @@
...
+
+export const testNotVisible = async componentId =>
+	await expect(element(by.id(componentId))).toBeNotVisible();
+
+export const tapBack = async () =>
+	await element(by.id(testIDs.Header.headerBackButton))
+		.atIndex(0)
+		.tap();
+
+// export const timeout = m => new Promise(r => setTimeout(r, m));
```
Tbaut: (nit) stray comments
```suggestion
```

e2e/e2eUtils.js
```diff
@@ -0,0 +1,62 @@
...
+
+export const testInput = async (inputId, inputText) => {
+	await element(by.id(inputId)).typeText(inputText);
+	await element(by.id(inputId)).tapReturnKey();
+	// if (device.getPlatform() !== 'ios') {
+	// 	await device.pressBack();
+	// }
```
Tbaut: (nit) stray comments

src/components/IdentitiesSwitch.js
```diff
@@ -0,0 +1,310 @@
...
+			: identities;
+
+		return (
+			<>
+				<ScrollView style={{ maxHeight: 180 }}>
+					<FlatList
+						data={identitiesToShow}
+						renderItem={renderNonSelectedIdentity}
+						keyExtractor={item => item.encryptedSeed}
+						style={{ paddingVertical: identities.length > 5 ? 8 : 0 }}
```
Tbaut: This (and the same check `>5` 2 lines down) looks sketchy :eyes:
Why `5` ?

src/components/IdentitiesSwitch.js
```diff
@@ -0,0 +1,310 @@
...
+										iconSize={24}
+										textStyle={fontStyles.t_big}
+										style={styles.indentedButton}
+									/>
+									<Separator />
+								</>
+							)}
+
+							{__DEV__ && (
+								<ButtonIcon
```
Tbaut: Why for DEV only? is that an experimental feature for now?

hanwencheng: Thanks a lot! the dev only should be the `add legacy account` button but not this one

hanwencheng:
I have add the current data schema in the description, hope that helps!

And I also agree with group styling, for example, there are lots of uniform styling in `TxDetails` `SignedTx` `MessageDetails` `SignedMessage`. It would be make reading and editing easy if we extract them out into one `commonStyles.js` file. In project scope that would need a lot of work and I agree to put them into future PR.

src/components/IdentitiesSwitch.js
hanwencheng: Just talked with Sveta, 8 is the Height of the scroll container.
If there are more than 5 identities, the list will become scrollable.
there will be shadow to indicate that there's more and some other details.

 So I think it is decide by the smallest screen, just checked on iPhone SE. Also tried on my big Android screen, there are more margin, but I think it is OK for now. We could programmatically change that as well.

hanwencheng: It would be better to discuss it separately, I think that would need discussion multiple possibilities together with Sveta, I will open an issue later this afternoon.

hanwencheng: IMG_0044

src/components/SecurityHeader.js
```diff
@@ -14,64 +14,48 @@
...
+
+function SecurityHeader({ navigation }) {
+	const [isConnected, setIsConnected] = useState(false);
+
+	useEffect(() => {
+		const subscribe = () =>
+			NetInfo.addEventListener(state => {
+				setIsConnected(state.isConnected);
+			});
+		return subscribe();
```
yjkimjunior: seems this is missing an unsubscribe() on component unmount?

src/components/Separator.js
```diff
@@ -0,0 +1,61 @@
...
+						backgroundColor: 'black',
+						height: 1,
+						marginVertical: 8
+					},
+					style
+				]}
+			>
+				{shadow && (
+					<Image
+						source={require('../../res/img/card_shadow.png')}
```
yjkimjunior: this require should be brought out of the component so it doens't happen on every render.

src/screens/AccountEdit.js
```diff
@@ -17,24 +17,24 @@
...
-	static navigationOptions = {
-		title: 'Edit Account'
-	};
-
 	constructor(props) {
 		super(props);
 	}
 
 	render() {
+		const onNameInput = async (accounts, name) => {
```
yjkimjunior: this should be brought out of the  `render()`

src/screens/AccountNetworkChooser.js
```diff
@@ -16,118 +16,241 @@
...
+			}
+			navigation.navigate('PathsList', { networkKey });
+		}
+	};
+
+	if (!loaded) return <ScrollView style={styles.body} />;
+
+	if (identities.length === 0) return showOnboardingMessage();
+
+	const NetworkList = Object.entries(NETWORK_LIST).filter(getNetworkKeys);
```
yjkimjunior: was there a special reason why this is not camelCase?

src/components/SecurityHeader.js
Tbaut: I noticed this, but it's "just" a typo imho. It should be named **un*subscribed, and this will be called when the component unmounts (because it's what the `useEffect` returns)

It was named `this.unsubscribe` previously

hanwencheng: It is kind of strange because this `subscribe` directly return an unsubscribe function. Probably name it as `unsubscribe`. 
sample usage from [NetInfo](https://github.com/react-native-community/react-native-netinfo)
```javascript
// Subscribe
const unsubscribe = NetInfo.addEventListener(state => {
  console.log("Connection type", state.type);
  console.log("Is connected?", state.isConnected);
});

// Unsubscribe
unsubscribe();
```

src/screens/AccountEdit.js
hanwencheng: Good point, I will also check if there is same issue in other component.

src/screens/AccountNetworkChooser.js
hanwencheng: typo here

src/components/SecurityHeader.js
Tbaut: This is how I'm used to see subscriptions. Polkadot-js api does the same. So to do things right we can rename it yes, but I think the logic is correct.

yjkimjunior: yeah, let's just refactor to save "future us" some stress

hanwencheng: now it is even cleaner:
```javascript
	useEffect(
		() =>
			NetInfo.addEventListener(state => {
				setIsConnected(state.isConnected);
			}),
		[]
	);
```

hanwencheng:
YJ also approved the PR with today's chat

yjkimjunior:
indeed

PR #435, opened 2019-10-25
chore: upgrade to RN 0.61.2

Tbaut:
I couldn't perform a successful e2e test on substrate development using `apps`. Because of the recent changes in Kusama I believe. I was able to perform an e2e test (transaction) on an older version of `apps` using the `dev-node`.
Ethereum Tx and signing were successfully tested as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "node-libs-react-native": "^1.0.3",
     "prop-types": "^15.6.1",
     "react": "^16.9.0",
-    "react-native": "0.60.5",
+    "react-native": "0.61.2",
     "react-native-camera": "^3.4.0",
     "react-native-gesture-handler": "^1.4.1",
     "react-native-keyboard-aware-scroll-view": "^0.9.1",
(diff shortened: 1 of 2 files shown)


Tbaut:
the following test fails:
>FAIL  src/util/decoders.spec.js
  ● Test suite failed to run
    TypeError: Cannot read property 'EthkeyBridge' of undefined
      20 | import { checksummedAddress } from './checksum';
      21 | 
    > 22 | const { EthkeyBridge } = NativeModules;

I don't get it, there's no `EthkeyBridge` in `src/util/decoders.spec.js` at line 22..

hanwencheng:
Ah, in jest they do not support bridging to the NativeModules, in jest we need to mock the `blake2s` function in the test, and for testing native module, we need to move them to e2e test like [here](https://github.com/paritytech/parity-signer/commit/4ad0e1fc96386d3d0fe9bdc47d7c4fd94df483c7)

I do not why there is not such an error in the previous version.

yjkimjunior:
i believe it was in a `skip` block before, might have been accidentally uncommented here.

Tbaut:
closing this for now. There's no urgency to upgrade.

PR #439, opened 2019-10-29
fix: Switch native blake to blake2b

maciejhirsz:


yjkimjunior:
It is included here https://github.com/paritytech/parity-signer/commit/ed696ac26321006c66e60fdf47632066799df20b, will add in the test too

PR #445, opened 2019-10-30
chore: clean build when do `yarn clean`

hanwencheng:
self explained.


yjkimjunior:
also maybe add `"clean-rust": cd rust/signer && cargo clean`

PR #454, opened 2019-11-05
feat: Rust<>JS bindings for Schnorrkel signature verification

yjkimjunior:


hanwencheng:
LGTM, though I do not have much knowledge about the Rust code part.

Questions are:
1. What is the usage of testing context part?
2. What is the difference of signing data compared with transactions and extrinsic, they are all string here, if there is a difference, is it possible to set up a test for some trasactions bytes or extrinsic bytes?

maciejhirsz:
Don't forget the swift part for iOS :)

@hanwencheng: not sure what you mean by context part. As for signing data vs extrinsics - there is no difference. I don't know if we have any concept of signing messages that are distinct from extrinsics/transactions in Substrate, if so we should at least deploy a similar protection Ethereum is using there (adding some salt to messages, so you can't craft a message that is actually a transaction).

yjkimjunior:
1. The context is like a consistent magic number used to sign along with the payload, in this case `Substrate` to indicate we are signing for Substrate

2. Messages, Transactions, Extrinsics can all be represented as byte slices and hexadecimal strings - the signature will always be created from the byte slice. The specifics of the Ristretto signing algorithm goes over my head but I know that it must use the byte slice as part of the signature requires setting the high bit of byte 31 to distinguish between sr25519 and ed25519 signature.  But we take as input the hexadecimal representation from JS. The conversion from hex to u8a in this case happens in `lib.rs`.

Issue #412, opened 2019-10-10
"Delete Account" does not work on accounts created with prev version; accounts restored after re-install


laboon:
DESCRIPTION
If you delete an account, that account is not really deleted.  If you uninstall and re-install the application, the deleted accounts are available once again.

As a side note, upon reinstall, any name changes you made will be reverted back to the original name.

TO REPRODUCE
1. Delete account TESTACCOUNT 
2. Uninstall Parity Signer
3. Re-install Parity Signer
4. Open Parity Signer

EXPECTED BEHAVIOR
TESTACCOUNT is not in list of accounts

OBSERVED BEHAVIOR
TESTACCOUNT _is_ in list of accounts

NOTES
This was on an iPhone 6 SE (Model Number MP7T2LL/A), software version 12.4.1.

laboon:
Looking through my accounts, it looked like it _did_ delete my Kusama accounts, just not my Ethereum ones.

laboon:
I just realized that I also generated these on a different version of Parity Signer, which may be relevant.

Tbaut:
Right, this is a very tricky case.. a proper Signer usage would prevent that, because you should not update Signer by upgrading the app with having your phone online...

Now I'm less surprised by this. Although we did implement a way to migrate older accounts, you most probably have a borked DB. I could not reproduce with a freshly installed Signer, with Kusama or Ethereum accounts.

laboon:
Yes, earlier I had assumed that I could use Parity Signer, delete the application, and simply keep the generated seed on paper.  But I did not realize that deleting the app did not delete its data.

laboon:
So I guess this not as big an issue as originally thought as it will not happen to users using the app correctly.

Tbaut:
>  But I did not realize that deleting the app did not delete its data

If you delete the app without having deleted the account, an iphone will keep that info (encrypted). Android would flush everything.

On iphone you'de have to delete things manually before uninstalling the app. Our instructions are much more radical though, you should factory reset the phone before setting it online again to download a new Signer version.

hanwencheng:
> If you delete an account, that account is not really deleted. If you uninstall and re-install the application, the deleted accounts are available once again.

The seed data on iOS is stored in the keychain. In this case after uninstalling and reinstalling the app, iOS do not have ways to prevent accessing data from same device. 

> As a side note, upon reinstall, any name changes you made will be reverted back to the original name.

[...]

PR #460, opened 2019-11-18
fix: loading error

hanwencheng:
Not sure when this part of loading code is changed.
Description:
If there is substrate code in the keychain, after reinstall, will have accountID generate error. It is because that migration ignored substrate account here.
related to #412 

Reproduce bug (only on iOS): 
Install the app, create one substrate account on KUSAMA, delete app, install again, now the first screen will see the error screen.


hanwencheng:
Indeed, let me check it again.

hanwencheng:
Closed, It seems on my devices there are multiple substrate account created in v2, deleted them and all looks good now.

PR #467, opened 2019-11-22
fix: seed recovery validation and unlock fix

### LINKED ISSUE #461: No feedback when entering the recovery phrase
On V3 we used to have a:
- green background when the recovery phrase was a valid BIP39
- warning message when the recovering a brain wallet (e.g a non valid BIP39)

I think this is very valuable to prevent users from freaking out if they recover from a phrase with typos and don't realize it.

example, this is an invalid BIP39, but I'll never know.
image

hanwencheng:
closes #461 .

tested with brain wallet seed with extra space:

1. '  ' (two spaces) Checked signing and account recovery.
2. 'a  ' (trailing spaces)
3. '  h' (heading spaces)
4. 'a  h' (two spaces in middle)


src/screens/IdentityNew.js
```diff
@@ -52,6 +59,18 @@ function IdentityNew({ accounts, navigation }) {
...
 
+	const onSeedTextInput = inputSeedPhrase => {
+		setSeedPhrase(inputSeedPhrase);
+		const addressGeneration = () =>
+			brainWalletAddress(inputSeedPhrase)
+				.then(({ bip39 }) => {
+					setIsSeedValid(validateSeed(inputSeedPhrase, bip39));
+				})
+				.catch(() => setIsSeedValid(inputSeedPhrase, false));
+		const debouncedAddressGeneration = debounce(addressGeneration, 200);
```
Tbaut: This debounce isn't working. I remember struggling a lot with it in the past (in `screens/AccountRecover.js`).
To verify what I said, add a `console.log` in  `addressGeneration`, you'll see that it's printed at every key stroke. I couldn't manage to make it work in a reasonable time in this component though :(.

hanwencheng: true, let me have a try.

hanwencheng: updated, in `debounce.js` we create new instance of `timeout` at each call of `debounce`.

PR #468, opened 2019-11-22
chore: update build version in ios

hanwencheng:


ios/NativeSigner/Info.plist
```diff
@@ -21,7 +21,7 @@
 	<key>CFBundleSignature</key>
 	<string>????</string>
 	<key>CFBundleVersion</key>
-	<string>3011</string>
+	<string>4000</string>
```
Tbaut: should actually be 400 if possible to keep android and ios aligned
```suggestion
	<string>400</string>
```

Tbaut:
we need to bump to 401 since we already pushed a 400 to apple connect. doing it now

PR #475, opened 2019-11-23
fix: prevent signing with unknown sender

### LINKED ISSUE #474: Scanning a transaction from an unknown sender does not report error.
Scanning a transaction with an unknown keyPairs still can be processed, it is wrongly recognized as the account in currentIdentity.

On `TxDetails` or `MessageDetails` pages it shows the account with currentIdentity, and signing process does not prevent it.

hanwencheng:
fix #474

### DIFF
--- a/src/screens/MessageDetails.js
+++ b/src/screens/MessageDetails.js
@@ -52,6 +52,8 @@ export default class MessageDetails extends React.PureComponent {
 				sender,
 				accountsStore.state.identities
 			);
+			if (!senderIdentity)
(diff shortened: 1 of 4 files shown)


hanwencheng:
Yeah, I finally removed that check. That situation actually will never happen. The `sender` in this case is from `getById`, and here we already [have](https://github.com/paritytech/parity-signer/pull/475/files#diff-5a05bc3a334f8afb74c93afef94f139bL375-L377) the check if sender is find, so that sender is never be null if it is not legacy account.

PR #477, opened 2019-11-26
chore: renaming to seedPhrase when phrase is used

hanwencheng:
Naming is the most difficult problem.

Renaming `seed` to `seedPhrase` if phrase is used.

In schema we use `encryptedSeed` since password will also used to secure that in the future version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,7 @@ build/
 .gradle
 local.properties
 *.iml
+gen/
 
 # node.js
 #
--- a/src/stores/AccountsStore.js
+++ b/src/stores/AccountsStore.js
@@ -296,13 +296,9 @@ export default class AccountsStore extends Container<AccountsStoreState> {
 
 	async saveNewIdentity(seedPhrase, pin) {
 		const updatedIdentity = deepCopyIdentity(this.state.newIdentity);
-		//TODO integrate password
-		const seed = constructSURI({
-			derivePath: '',
-			password: '',
-			phrase: seedPhrase
-		});
-		updatedIdentity.encryptedSeed = await encryptData(seed, pin);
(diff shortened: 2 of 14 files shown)


src/stores/AccountsStore.js
```diff
@@ -296,6 +296,8 @@ export default class AccountsStore extends Container<AccountsStoreState> {
 
 	async saveNewIdentity(seedPhrase, pin) {
 		const updatedIdentity = deepCopyIdentity(this.state.newIdentity);
+		//TODO encrypt seedPhrase with password in the future version,
+		// current encryption with only seedPhrase is compatible.
 		updatedIdentity.encryptedSeed = await encryptData(seedPhrase, pin);
```
hanwencheng: Currently we only encrypt `seedPhrase` in `encryptedSeed`, but will include `password` in the future

src/stores/types.js
```diff
@@ -38,7 +38,8 @@ type AccountMeta = {
 };
 
 export type Identity = {
-	encryptedSeedPhrase: string,
+	// encrypted seed include seedPhrase and password
```
hanwencheng: currently only seedPhrase

.gitignore
```diff
@@ -45,6 +45,7 @@ build/
 .gradle
 local.properties
 *.iml
+gen/
```
yjkimjunior: just curious, what is generating this?

hanwencheng: I don't know either, after build android with `gradle assemble`, it created there files: `buildConfig.java`, `Manifest.java` and `R.java`. On each of them there is comments said: 

```
/*___Generated_by_IDEA___*/
```

So I just add them there.

PR #494, opened 2019-12-05
feat: enable create root account under a network

### LINKED ISSUE #484: Create an account under root of the network
In order to support soft derivation from the networks root key pairs, under each network a root key pairs / account (e.g. with path `//kusama`) need to be created as default. 

Also when user choose the new network, now it is a path derivation screen, but it should now lead to pin input and create an aforementioned root keypiar / account.

hanwencheng:
closes #484 

Enable user to use root path for transferring the public address to a hot machine and creating soft derivations.

out mp4

Now there is two way to create a root account, one is to click the top account bar, and the other one is to use derive path function with empty string.

455290f : This commit changes the default created account to root account, for example, when you click add new network and then click Kusama, the app create path `//kusama` instead of `//kusama//default`.

The display of it is kind of implicit, I use this way because the root account is used differently from other account, it probably will be used frequently for soft key derivation, and not for signing. But we should not guess what the user do with the key pairs,  discussion are welcome and UI could be changed further.


hanwencheng:
Merge to Master and will need review again. 

I mistaken add the refactor of "`Add New Network` button to the bottom" in this PR, as it is not related to the topic, just removed, and will do it in other PR.

PR #497, opened 2019-12-09
feat: Enable create custom path

### LINKED ISSUE #488: Enable v3 user to recover their legacy substrate account
If user have created substrate account before v4 version. Now user could still use it in legacy accounts list, but if they lose their phone/reset it, they are not being able to recover a legacy account. Since all the substrate account need to couple with a network path like `//kusama`.

So it would be useful to make backward compatible with v3 substrate root accounts. The function could existed without affect the main UI. For example: in the recover identity page, the confirm button could has a dropdown, the second option is that to create a legacy account, and when user clicks it, we prompt a warning that this is only for such v3 user lost recovering case.

hanwencheng:
closes #488

This PR enables create account with custom path and new display of root account.

### Custom Path

| Unknown Network        | KnownNetwork           | 
|custom_unknown|custom_known|

Now the address is generated with Polkadot Canary Prefix which is `2`. In the future we should support user to switch between different networks as Sveta comments here: https://github.com/paritytech/parity-signer/issues/488#issuecomment-565509817.

### Root Account

The root account should be access from the landing screen which is network select view, after this PR all user will defaulty get root account, for backward-compatibility, old user will generate the account when clicking on the identity head.

| Pre v4 User get Root Account  | Current v4 User has  Root by default  | 
|out|out|

### Signing test
| Unknown Network Account Signing  | Root Account Signing  | 
|unknown_sign|out mp4|

* Naming of the custom account in the records may be mixed with identity name, fixed it in the latest commit

Warnning text are changed to 

Screen Shot 2019-12-18 at 18 18 17

Rephrase suggestions are welcome!

Root account could also be generated with the path of empty string "".

### How to test

1. Generate a custom network with random path. It should be recognize as a substrate account on Polkadot.js, also extrinsic signing is possible with the account.

2. Root account created should also be possible for signing.

3. All account generate should be possible to be deleted and regeneragted.


src/components/UnknownAccountWarning.js
```diff
@@ -21,19 +21,31 @@ import { StyleSheet, Text, View } from 'react-native';
...
-				{'\n'}
-				To be able to use this account you need to:{'\n'}- write down its
-				recovery phrase{'\n'}- delete it{'\n'}- recover/derive it{'\n'}
-			</Text>
+			{isPath ? (
+				<Text style={styles.warningText}>
+					The account is now using Polkadot Canary Address Format.
+					{'\n'}
+					In the future user may be able specify different address formats like
+					Polkadot Live, which may lead to a different address.
```
Tbaut: This will not age well, "in the future" doesn't tell much to a user that will not want to update.
I'd go for explaining the current situation that's it. Also Canary network: Kusama.
suggestion:
> This account is not bond to a specific network. The address currently displayed is using Kusama format, however, this account can be used for any network. The address may be different.

hanwencheng: > however, this account can be used for any network. The address may be different.

`used` maybe misleading, so I suggest rephrase it to.

> however, this account can be encoded by another format, with a different address.

hanwencheng:
* What about renaming "unknown network" to "any network" ?
"Any network" IMHO is not a good item name in the network list. With the warning I can understanding the meaning. 

* We should show the icon of the account here as well.
Good point, changed in the new commits.

* Signing bug:
I did not check the signing in Ethereum, thanks for checking that out. It is the problem where the undefined attributes of `account` is ignored. At this point, type checker is definitely would easy the development. Wrapped all of them in typescript would be very helpful.

Some further thought:
The `unknown network` naming and icon display is highly related to the future feature, which is enable choose the network of the `unknown account` as Sveta comment and sketches. https://github.com/paritytech/parity-signer/issues/488#issuecomment-565509817.

To that point we I think probably we can rename it further to `custom network`, and identicon could be changed to `polkadot` user has select `polkadot` in the path detail view.

Tbaut:
> To that point we I think probably we can rename it further to custom network, and identicon could be changed to polkadot user has select polkadot in the path detail view.

These accounts are for substrate account anyways since Ethereum account don't have derivation. The identicon is the same for accounts on Kusama/dev/Substrate/anything because it's linked to the public key, not to the address. I really think it'd make sense to show the icon there. 

As mentioned before, this is a security consideration since the address can look different, at least the identicon will be the same.

hanwencheng:
The commit name is misleading, icon display is already include in the commit e80a4fe .

Tbaut:
I was able to reproduce. So it's a Tx originating from a root account (in my case `//kusama_dev`).
Either it throws (like above), or it actually give you the QR code, but the Tx errors out, because I imagine the signing is wrong.

here is the tx that either errors out, or throws:
image

PR #498, opened 2019-12-09
fix: error with new recipient account

### LINKED ISSUE #496: Can't sign a Tx with Sparta
How to reproduce:

- create an identity with "this is sparta"
- select Ethereum Kovan

get:
image

hanwencheng:
closes #496 .

It is not a problem with "this is sparta" account, it is a problem if recipent is not a recognized account in Ethereum.

### Reproduce problem:

Create any ethereum account, scan any transaction QR with an unrecognized destination address, error happens.

The bug affect so widely and is so obvious, I am feeling guilty for creating it, we will need to release a fix version soon.

### DIFF
--- a/src/stores/ScannerStore.js
+++ b/src/stores/ScannerStore.js
@@ -384,7 +384,7 @@ export default class ScannerStore extends Container<ScannerState> {
 			(await accountsStore.getById({
 				address: recipientAddress,
 				networkKey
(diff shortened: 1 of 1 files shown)


Tbaut:
> The bug affect so widely and is so obvious, I am feeling guilty for creating it, we will need to release a fix version soon.

The responsibility is shared with the reviewers and testers :)

Testing now.

PR #499, opened 2019-12-10
style: move scan button to top bar

### LINKED ISSUE #486: Move scan button to the top bar
Now the scan button is displayed with absolute position which will also cover the other UI element. 

For example:

4860ad50-f494-44dd-bb3c-eb164e0343b3

It would be better to move it to the top bar with a small icon which is the same as identitiesSwitch.

hanwencheng:
closes #486.

Screen Shot 2019-12-10 at 15 48 10

Now the scan button is moved to the top right bar of the application.


hanwencheng:
Just think of one UI thing, maybe in the above network choosing screen, we may keep the `add Network Account` button to the bottom of the screen.

PR #503, opened 2019-12-17
WIP: feat: Generate Root account for all the user.

### LINKED ISSUE #488: Enable v3 user to recover their legacy substrate account

hanwencheng:
closes #488


hanwencheng:
merged into #497

PR #508, opened 2019-12-27
fix: custom path derivation with known network

### LINKED ISSUE #507: Custom path derivation with a known network add an extra path
How to reproduce:

Select any identity -> Add network account -> Create Custom Path -> on path input use "//kusama" as prefix

Then the path created will be prefixed with "//kusama//kusama"

|incorrect behavior| expected correct behavior|
|---|---|
|out|out|

hanwencheng:
closes #507 .

It works as same as expected behaviour in the issue description.

### DIFF
--- a/src/screens/PathDerivation.js
+++ b/src/screens/PathDerivation.js
@@ -30,7 +30,7 @@ import {
 	navigateToPathsList,
 	unlockSeedPhrase
 } from '../util/navigationHelpers';
-import { NETWORK_LIST, UnknownNetworkKeys } from '../constants';
+import { NETWORK_LIST } from '../constants';
 import { alertPathDerivationError } from '../util/alertUtils';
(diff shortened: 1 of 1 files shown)


src/screens/PathDerivation.js
```diff
@@ -43,22 +43,20 @@ function PathDerivation({ accounts, navigation }) {
...
+	const completePath = isCustomPath
+		? derivationPath
+		: `${currentNetworkPath}${derivationPath}`;
 
 	const onPathDerivation = async () => {
 		if (!validateDerivedPath(derivationPath)) {
 			return setIsPathValid(false);
 		}
 		const seedPhrase = await unlockSeedPhrase(navigation);
+		console.log('complete path is', completePath, 'networkKey:', networkKey);
```
Tbaut: stray log

PR #515, opened 2020-01-03
feat: add support for 12 mnemonic option

### LINKED ISSUE #405: Feat: add advanced option to allow 12 word recovery phrase

hanwencheng:
closes #405 

Behavior on iOS and Android as following:

out

@goldsteinsveta  do you have any suggestions about the two new buttons?

### DIFF
--- a/src/screens/IdentityBackup.js
+++ b/src/screens/IdentityBackup.js
@@ -50,8 +50,8 @@ function IdentityBackup({ navigation, accounts }) {
 	const renderTextButton = buttonWordsNumber => {
 		const textStyles =
 			wordsNumber === buttonWordsNumber
-				? fontStyles.t_codeS
-				: { ...fontStyles.t_codeS, color: colors.label_text };
+				? { ...fontStyles.t_codeS, color: colors.label_text }
+				: fontStyles.t_codeS;
 		return (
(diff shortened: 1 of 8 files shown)


src/screens/IdentityBackup.js
```diff
@@ -34,9 +34,11 @@ import { withAccountStore } from '../util/HOC';
 import testIDs from '../../e2e/testIDs';
 import ScreenHeading from '../components/ScreenHeading';
 import { alertBackupDone, alertCopyBackupPhrase } from '../util/alertUtils';
+import Button from '../components/Button';
 
 function IdentityBackup({ navigation, accounts }) {
 	const [seedPhrase, setSeedPhrase] = useState('');
+	const [wordsNumber, setWordsNumber] = useState(12);
```
yjkimjunior: I hate to be so nitty but I think the default should be 24 words, and power users can opt to use just a slightly less safe 12 word mnemonic. Just my opinion, not sure what Gav asked for here.

hanwencheng: true, the original discussion was using 12 words as an option.

hanwencheng:
fixed with 24 words as default

PR #516, opened 2020-01-07
feat: enable path derivation from existed path card

hanwencheng:
A small feature which ease the path creation process.

How to test:
Select Kusama network -> select any path -> tap top right menu button -> tap "Derive Account" button -> path derivation card should be a pre-filled path, and path creation should be correct.

At the same time, other path derivation should keep the same as before, which is:
1. network path derivation: Network selection screen tap "Kusama" -> tap "Create New Derivation" button -> behavior same as before
2. custom path derivation: Network selection screen tap "Add Network Account" -> Scroll to bottom and choose "Create Custom Path" -> behavior same as before

out

### DIFF
--- a/specs/util/identitiesUtils.spec.js
+++ b/specs/util/identitiesUtils.spec.js
@@ -19,6 +19,7 @@
 import {
 	deserializeIdentities,
 	getExistedNetworkKeys,
+	getNetworkKeyByPath,
 	getPathName,
 	groupPaths,
 	serializeIdentities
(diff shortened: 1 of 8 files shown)


hanwencheng:
Also I am also thinking about the renaming of the button from `Derive Account` to `Derive Child Account`, what do you suggest @Tbaut ?

Tbaut:
Regarding the naming (add "child"), I don't mind, not sure it clarifies anything though. One thing for sure, we need to unify the naming. We have "derive account", then on the button "derive address".

Regarding this PR, just looking at the presentation, I think it's confusing that the workflow is different (not a button but a menu entry) but in the end the behaviour is the same as the "derive account" button. I feel that for the same behaviour, we should have the same flow. Any thought on that?

hanwencheng:
> Regarding the naming (add "child"), I don't mind, not sure it clarifies anything though. One thing for sure, we need to unify the naming. We have "derive account", then on the button "derive address".

True, will change that.

> but in the end the behaviour is the same as the "derive account" button. I feel that for the same behaviour, we should have the same flow. Any thought on that?

The behavoir is different, take a look at the path, this derive path start with the path of the current select account, in the above example,  it is `//kusama//controller//nominating` and user give `/1` will derive account of `//kusama//controller//nominating/1`. It is like a syntax sugar for user with different and complex account.

Do not have any feature request from users, just personally think that would be useful.

hanwencheng:
Change the related text.

From the comments I realize that the derive base path is hard to distinguish. So in the commit: 
7a5dbd3 I refactored the subtitle in the path derivation screen.

So now the subtitle always shows the base path of the derivation.

For example, in the flow:
* select identity -> select kusama network -> select path "//kusama//default" -> "derive account" button -> show base path "//kusama//default"
* select identity -> select kusama network -> tap "create new derivation" -> show base path "//kusama"
* select identity -> tap "addn new network" -> tap "create custom path" -> show base path empty string ""

Issue #519, opened 2020-01-07
Android - The submit key on the keyboard doesn't do anything


Tbaut:
From the same active user, on Android with a stock keyboard:
> in the PIN entry dialog somehow only the on-screen button "Done" allows for submitting the pin. The "Submit" button on my digit keyboard (the one with a checkbox) does nothing — and this is confusing. I would pretty much prefer that button to actually submit the PIN, since I don't see any reason not to

hanwencheng:
closed by #520.

PR #520, opened 2020-01-08
fix: pin input text and return button

### LINKED ISSUE #518: Misleading PIN helper text
From this same active user:
> when I type in the PIN (which I need to do individually for every single derivation, which is OK), I'm shown the same dialog as for setting up the PIN? The "Choose a PIN ode with 6 or more digits" subtitle in that dialog is really misleading, since I'm not setting up a PIN, but retyping existing one.

We should probably say "wrong pin" in the case where we are checking the pin, and not even talk about the 6 digits limit.

hanwencheng:
closes #518 , #519

## Unlock pin Input Text
4a30916 : is for #518, now the text is changed, and all the text in `IdentityNew` is ordered into a object/map called `t`.

Screen Shot 2020-01-08 at 20 09 49

## Return button in text input component

The other three commits are for #519 , the main fix is in dc8d65e , which add `onSubmitEditing` for TextInput Component. Other two commits is mainly for e2e tests.
The unfunctioned button is not only in Android but also in iOS, now I have address all the return button in both platforms:
1. in path derivation process, path input and account name input.
2. in pin unlock screen, pin input
3. in pin creation screen, pin input
4. in path recovery screen, seed input

## How to test:
for above 1 and 2, try create any path from a Kusama network:
 
out

for above 3 and 4, try recovery an identity with any seed:

|android|ios|
|---|---|
|out_android|out|

### DIFF
--- a/e2e/e2eUtils.js
+++ b/e2e/e2eUtils.js
@@ -43,6 +43,17 @@ export const testInput = async (inputId, inputText) => {
 	await element(by.id(inputId)).tapReturnKey();
 };
 
+export const testInputWithDone = async (inputId, inputText) => {
+	await element(by.id(inputId)).typeText(inputText);
+	if (device.getPlatform() === 'ios') {
+		await element(by.label('Done'))
+			.atIndex(0)
+			.tap();
+	} else {
+		await element(by.id(inputId)).tapReturnKey();
+	}
+};
+
 export const testScrollAndTap = async (buttonId, screenId) => {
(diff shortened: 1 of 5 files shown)


hanwencheng:
Does the slowness also happens before this PR? what about the seed input in recovery page?

Tbaut:
it's really random, you can disregard that comment, I didn't have the problem any more, on master or this branch..

PR #521, opened 2020-01-09
refactor:  migrating to typescript

### LINKED ISSUE #418: Migrate to typescript
There are many places where stricter guarantees would be useful.

hanwencheng:
closes #418.

Include update to React Navigation v4.

This PR I avoid to make changes to all the functionalities and UI features.

But may need to test all the signing issues.

Other changes includes: 
* reorder the file structure
* use alias for import all `components`, `screens`, `styles`, `constants`, `utils`, `stores`, `types`
* update requirement of node to `>=10` because of dev dependency `fsevents`


jest.config.js
```diff
@@ -14,14 +14,30 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
-const { defaults } = require('jest-config');
+// const { defaults } = require('jest-config');
```
Tbaut: stray comment

hanwencheng: good catch! also deleted other `console.log`

sjeohp:
Testing on iOS.

- Adding legacy account doesn't look right.

Screenshot 2020-03-02 at 11 46 49 AM

- `<` button on `About` screen is sometimes invisible. Don't know how to reproduce reliably, it just happens sometimes.

IMG_9E6793CB433C-1

- For an Ethereum account: 
1. Choose `Delete`
2. Alert (`Do you really want to delete key pairs for this account`) sometimes appears, sometimes does not.
3. `Unlock Identity` pin entry fails (`You pin must be at least 6 digits long!`) even when the pin is correct.
4. `<` back button takes me to a blank screen.
IMG_FA72A5713C1B-1
5. `<` from the blank screen takes me to the accounts list and the Ethereum account has been deleted.

sjeohp:
- Account icons sometimes disappear, possibly after the buggy Ethereum deletion.
IMG_6C25F77A1BC2-1

hanwencheng:
Re @sjeohp , thanks for checking it!

1. Add legacy account screen is only used for developing, it will never shows in production app, so I just use it to check backward-compatibility, I will leave it for now.

2. Ethereum deletion problem is a vital bug, I have fixed it, also noticed that the alert text with "key pairs" is not clear to the user, so delete all the occurence of "key pairs" and use "account" instead.

3. With checking back button issue, I tried hard but still can't reproduce it on either Andorid and iOS, probably it is related to ehtereum account deletion problem. I guess it is caused by react navigation, so I updated  it to latest version. 

4. When trying reproduce `3`, found a new bug, "after screen transition, only after the small delay, buttons on new screen will work."  it is fixed by upgrading react navigation. 

5. Network Icon missing problem, it is a bug by misusage of react `useEffect` hook, with async function in `useEffect`. The bug is updating UI (in `useEffect` hook) after the screen is destroyed. So I introduces `eslinit-plugin-react-hooks` into Eslint, and fixes the bug in both `AccoutIcon` and `QrView`.

PR #523, opened 2020-01-13
fix: display soft derivation custom path

### LINKED ISSUE #522: Custom path start with "/" can not be displayed
The problem happened in both iOS and Android.

How to reproduce:
Select any identity -> tap "Add Network Account" -> "Create Custom Path" -> in path input any thing start with `/`, for example `/soft`.

The account created then will not be displayed on any screen. But it could be used for sign the transaction.

hanwencheng:
close #522 

The bug is caused by the incorrect sorting function in `identitiesUtils.js`

Fix the bug and refactor the test suit in `identitiesUtils.spec.js`, which make it easy to add more cases.

Also now the unknown paths will be grouped by the first catch group of path.

In the following example  `/soft/path`, `/soft/222`, `/soft/333` will be grouped together under `/soft`

Signing tested successfully on iOS.

out

### DIFF
--- a/src/util/identitiesUtils.js
+++ b/src/util/identitiesUtils.js
@@ -206,12 +206,11 @@ export const getPathName = (path, lookUpIdentity) => {
 };
 
 /**
- * This function grouped paths list into a list with different groups, for a better display.
- * The group will based :
- * on their second subpath if it is a known network, for example, '//staking' of '//kusama//staking/0'
- * on their first subpath if it is an unknown network, for example, '/random' of '/random//derivation/1'
- * examples please refer to the path group test suit in identitiesUtils.spec.js
- */
+ * This function decides how to group the list of derivation paths in the display based on the following rules.
+ * If the network is unknown: group by the first subpath, e.g. '/random' of '/random//derivation/1'
+ * If the network is known: group by the second subpath, e.g. '//staking' of '//kusama//staking/0'
+ * Please refer to identitiesUtils.spec.js for more examples.
+ **/
 export const groupPaths = paths => {
 	const insertPathIntoGroup = (matchingPath, fullPath, pathGroup) => {
 		const groupName = matchingPath.match(pathsRegex.firstPath)[0];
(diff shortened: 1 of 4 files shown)


src/util/identitiesUtils.js
```diff
@@ -205,33 +206,38 @@ export const getPathName = (path, lookUpIdentity) => {
 };
 
 export const groupPaths = paths => {
```
sjeohp: Can you write some comments documenting what this function is doing?

src/util/identitiesUtils.js
```diff
@@ -37,12 +37,13 @@ const extractPathId = path => {
...
 	const pathFragments = path.match(pathsRegex.allPath);
-	if (!pathFragments || pathFragments.length <= 1) return '';
+	if (!pathFragments || pathFragments.length === 0) return '';
+	if (pathFragments.length === 1) return removeSlash(pathFragments[0]);
 	return removeSlash(pathFragments.slice(1).join(''));
 };
 
 export const isSubstratePath = path =>
-	path.split('//')[1] !== undefined || path === '';
+	path.match(pathsRegex.allPath) != null || path === '';
```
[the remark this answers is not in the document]
hanwencheng: my mistake, changed.

src/util/identitiesUtils.js
```diff
@@ -205,6 +205,13 @@ export const getPathName = (path, lookUpIdentity) => {
 	return extractSubPathName(path);
 };
 
+/**
+ * This function grouped paths list into a list with different groups, for a better display.
+ * The group will based :
+ * on their second subpath if it is a known network, for example, '//staking' of '//kusama//staking/0'
+ * on their first subpath if it is an unknown network, for example, '/random' of '/random//derivation/1'
+ * examples please refer to the path group test suit in identitiesUtils.spec.js
+ */
```
yjkimjunior: quick grammar cleanup:

```
/**
* This function decides how to group the list of derivation paths in the display based on the following rules.
* If the network is unknown: group by the first subpath, e.g. '/random' of '/random//derivation/1'
* If the network is known: group by the second subpath, e.g. '//staking' of '//kusama//staking/0'
* Please refer to identitiesUtils.spec.js for more examples.
**/
```

Tbaut:
Applied the grammar correction. Merging so that I can test master asap for a release.

Issue #525, opened 2020-01-16
Kusama --dev can't be decoded correctly


Tbaut:
Metadata for Kusama are outdated, and this results in unwanted behaviors:
How to reproduce:

- connect to kusama --dev (v0.7.17 iirc) on `wss://dev-node.substrate.dev` and update the genesisHash for kusama dev to `0x5e9679182f658e148f33d3f760f11179977398bb3da8d1f0bf7b267fe6b3ebb0`
- submit a pre-image (democracy) with system.setRemarks 0x01 (and copy the hash)
- signing the pre-image says it can't decode it (shouldn't be the case). Sign it still
- submit a proposal and paste the pre-image previously copied

this results in a wrongly decoded extrinsic (setCouncilMember ?):
![image](https://user-images.githubusercontent.com/33178835/72521602-09c0d300-385c-11ea-9309-5f7f657c02ec.png)

edit: this only happens to --dev, the real Kusama is actually able to decode the `democracy.notePreimage`

Tbaut:
pretty sure it's linked to https://github.com/paritytech/parity-signer/issues/438

yjkimjunior:
that's really really really weird that it only happens on --dev (even before with your Substrate node).

It's a longshot but Is it possible that the dev node is the one giving back the wrong metadata?

jacogr:
`--dev` runs the Polkadot chainspec, not Kusama.

Tbaut:
Thanks for the heads-up, I wasn't aware :(. Closing

PR #527, opened 2020-01-16
chore: Bump Polkadot api

Tbaut:
In an attempt to solve #525 still in vain.
Bumping polkadot packages can't hurt still.

I was able to sign a tx, please make sure all tests pass (I still can't run them).


package.json
```diff
@@ -38,11 +38,11 @@
   "dependencies": {
     "@plugnet/util": "^1.1.100",
     "@plugnet/wasm-crypto-js": "^0.11.102",
-    "@polkadot/api": "^0.98.0-beta.0",
-    "@polkadot/reactnative-identicon": "^0.46.1",
-    "@polkadot/types": "^0.98.0-beta.1",
-    "@polkadot/util": "^1.7.1",
-    "@polkadot/util-crypto": "1.7.1",
+    "@polkadot/api": "^0.100.1",
```
yjkimjunior: i think it's `1.0.0-beta.7`

it seems `yarn upgrade-interactive --latest` ignores the beta tags

package.json
```diff
@@ -38,11 +38,11 @@
...
+    "@polkadot/reactnative-identicon": "^0.47.1",
+    "@polkadot/types": "^0.100.1",
+    "@polkadot/util": "^1.8.1",
+    "@polkadot/util-crypto": "^1.8.1",
```
yjkimjunior: same here `2.0.0-beta.4` https://github.com/polkadot-js/common/tree/master/packages

package.json
```diff
@@ -38,11 +38,11 @@
...
     "@plugnet/wasm-crypto-js": "^0.11.102",
-    "@polkadot/api": "^0.98.0-beta.0",
-    "@polkadot/reactnative-identicon": "^0.46.1",
-    "@polkadot/types": "^0.98.0-beta.1",
-    "@polkadot/util": "^1.7.1",
-    "@polkadot/util-crypto": "1.7.1",
+    "@polkadot/api": "^0.100.1",
+    "@polkadot/reactnative-identicon": "^0.47.1",
+    "@polkadot/types": "^0.100.1",
+    "@polkadot/util": "^1.8.1",
```
yjkimjunior: same here 2.0.0-beta.4 https://github.com/polkadot-js/common/tree/master/packages

package.json
yjkimjunior: https://github.com/polkadot-js/api/blob/be258a1f7c997ad8e31e8b515833c943672bfb76/package.json#L2

Tbaut:
Do we want the beta though? I did a `yarn upgrade --latest` on those specific packages. We used to have the beta only for `/api` and `/types`, not sure what `beta` would bring or break.

Tbaut:
Chatted with Jaco, `beta` is needed if we want to support everything indeed.

Tbaut:
I was able to do a tx with beta as well with `kusama --dev`.

PR #528, opened 2020-01-16
chore: bump Signer to v4.0.6

Tbaut:


Tbaut:
Merging as insubstantial so that I can build the apk.

PR #531, opened 2020-01-17
fix: add whitespace after tapping the suggestion text

### LINKED ISSUE #529: Add a white space when clicking on a proposed mnemonic word
When entering a mnemonic:
- a white space should be added automatically after the word
- the mnemonic should be trimmed from ending white space before being checked

hanwencheng:
close #529 .

Now after tapping the suggestion text, the trailing space will be automatically added.

And for a 24 words recovery phrase, it will not add the whitespace to the end.

But for a 12, 15, 18, or 21 words (They should all be possible to be recovered), the trailing whitespace is still added, but after tapping the "Recover Identity" button and input PIN, it will automatically be trimmed.

Test cases:
1. tapping any suggestion word, a whitespace should follow.
2. move the cursor to middle, and add another suggestion word, no extra trailing whitespace in the end should be added.
3. After input bip39 mnemonic phrase, the box border should not be red though there is a trailing whitespace.
4. after  input bip39 mnemonic phrase and recovered the identity, the recover phrase should not include any trailing whitespace, they should be removed during recovery process.
5. After input 24 bip39 words, there should be no more traling whitespaces added.

### DIFF
--- a/src/components/AccountSeed.js
+++ b/src/components/AccountSeed.js
@@ -112,7 +112,7 @@ export default class AccountSeed extends Component {
 						<TouchableItem
 							key={i}
 							onPress={e => {
-								let phrase = left.concat(suggestion, right).join(' ');
+								let phrase = left.concat(suggestion, right).join(' ').trimEnd();
 								const is24words = phrase.split(' ').length === 24;
 								if (!is24words) phrase += ' ';
 								this.props.onChangeText(phrase);
--- a/src/components/AccountSeed.js
+++ b/src/components/AccountSeed.js
@@ -112,7 +112,10 @@ export default class AccountSeed extends Component {
 						<TouchableItem
 							key={i}
 							onPress={e => {
-								let phrase = left.concat(suggestion, right).join(' ').trimEnd();
+								let phrase = left
+									.concat(suggestion, right)
+									.join(' ')
+									.trimEnd();
 								const is24words = phrase.split(' ').length === 24;
 								if (!is24words) phrase += ' ';
 								this.props.onChangeText(phrase);
--- a/src/components/AccountSeed.js
+++ b/src/components/AccountSeed.js
@@ -112,8 +112,9 @@ export default class AccountSeed extends Component {
 						<TouchableItem
 							key={i}
 							onPress={e => {
-								const phrase = left.concat(suggestion, right).join(' ');
-
+								let phrase = left.concat(suggestion, right).join(' ');
+								const is24words = phrase.split(' ').length === 24;
+								if (!is24words) phrase += ' ';
 								this.props.onChangeText(phrase);
 							}}
 						>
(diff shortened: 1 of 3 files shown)


src/util/account.js
```diff
@@ -72,17 +72,19 @@ export function validateSeed(seed, validBip39Seed) {
...
 		return {
 			accountRecoveryAllowed: false,
+			bip39: false,
 			reason: 'A seed phrase is required.',
 			valid: false
 		};
 	}
 
-	const words = seed.split(' ');
+	const words = validBip39Seed ? seed.trimEnd().split('') : seed.split(' ');
```
sjeohp:
```suggestion
	const words = validBip39Seed ? seed.trim().split(' ') : seed.split(' ');
```

src/screens/IdentityNew.js
```diff
@@ -62,7 +62,7 @@ function IdentityNew({ accounts, navigation }) {
 	const onSeedTextInput = inputSeedPhrase => {
 		setSeedPhrase(inputSeedPhrase);
 		const addressGeneration = () =>
-			brainWalletAddress(inputSeedPhrase)
+			brainWalletAddress(inputSeedPhrase.trimEnd())
```
sjeohp: Can `trimEnd` be removed here and below if `trim` is used in `onPress` above? If so it would be better to do it in just the one place.

src/screens/IdentityNew.js
```diff
@@ -74,7 +74,11 @@ function IdentityNew({ accounts, navigation }) {
 	const onRecoverIdentity = async () => {
 		const pin = await setPin(navigation);
 		try {
-			await accounts.saveNewIdentity(seedPhrase, pin);
+			if (isSeedValid.bip39) {
+				await accounts.saveNewIdentity(seedPhrase.trimEnd(), pin);
```
sjeohp: Same as above

src/components/AccountSeed.js
```diff
@@ -112,8 +112,9 @@ export default class AccountSeed extends Component {
 						<TouchableItem
 							key={i}
 							onPress={e => {
-								const phrase = left.concat(suggestion, right).join(' ');
-
+								let phrase = left.concat(suggestion, right).join(' ');
```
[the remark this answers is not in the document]
hanwencheng: User could still use whitespace with brain wallet, for example, user could use `  sparta arrow`, but when user taping the `arrow`, the heading whitespace will be trimed, we should avoid to make any changes to the `left` part here.

src/util/account.js
hanwencheng: If it is a validBip39Seed, it should not has already be validated by [line 65](https://github.com/paritytech/parity-signer/pull/531/files#diff-d1d34e6b6d649b5a460b1bf20a221d1eR65). So there is no heading whitespaces for seed if it is validBip39Seed.

hanwencheng: second one is true, I missed the whitespace in `split` function

sjeohp: I think `.split(' ')` is the more important change here

src/components/AccountSeed.js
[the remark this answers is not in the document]
sjeohp: What about `trimEnd` then instead?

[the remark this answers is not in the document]
hanwencheng: It indeed prevent the extra space when tapping the word in the middle like in the following situation. Is that the one you are concern? 
out

PR #532, opened 2020-01-20
refactor: identity root account and network root account

hanwencheng:
Extract Identity root account and  network root account from the clickable heading.

### UI display

Now Identity root account is in the custom network account list, and network root account is in the correspond network account list.

|identity root account | network root account | network chooser list
|---|---|---|
|Screen Shot 2020-01-20 at 13 20 50|Screen Shot 2020-01-20 at 13 20 07|Screen Shot 2020-01-20 at 13 19 59|

### Account Generation

* The network root account will be automatically generated when user choose to add the certain substrate network.
* The identity root account will be generated when user explicitly derive a custom path with empty string.

| identity root account | network root account |
|---|---|
|out|out|

### DIFF
--- a/src/stores/AccountsStore.js
+++ b/src/stores/AccountsStore.js
@@ -361,7 +361,6 @@ export default class AccountsStore extends Container<AccountsStoreState> {
 		// current encryption with only seedPhrase is compatible.
 		updatedIdentity.encryptedSeed = await encryptData(seedPhrase, pin);
 		//TODO now hard coded to polkadot canary prefix which is 2, future enable user to change that.
-		await this._addPathToIdentity('', seedPhrase, updatedIdentity, 'Root', 2);
 		const newIdentities = this.state.identities.concat(updatedIdentity);
 		this.setState({
 			currentIdentity: updatedIdentity,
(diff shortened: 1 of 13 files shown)


src/stores/AccountsStore.js
```diff
@@ -361,7 +361,6 @@ export default class AccountsStore extends Container<AccountsStoreState> {
 		// current encryption with only seedPhrase is compatible.
 		updatedIdentity.encryptedSeed = await encryptData(seedPhrase, pin);
 		//TODO now hard coded to polkadot canary prefix which is 2, future enable user to change that.
-		await this._addPathToIdentity('', seedPhrase, updatedIdentity, 'Root', 2);
```
sjeohp: can you explain this change?  does it address the TODO and if so, should the TODO be deleted as well?

hanwencheng: In the previous version we automatically create a root account without network affiliation if user create a new identity, this line of code does that.

In the current version we only allow user to explicitly create a root account with a network key chosen.

And yes, the TODO need to be deleted.

PR #536, opened 2020-01-22
feat: enable network selection for custom path

### LINKED ISSUE #533: Enable network selection for accounts with custom path
Currently all the accounts with custom path falls back to Kusama Network.

We would enable a user for selecting other network for custom paths during creation, which will need to add network affiliation info in to meta for the accounts with custom path.

would be same as the bottom part of the following screenshot.
Screen Shot 2020-01-20 at 23 24 15

hanwencheng:
closes #533 .

This PR includes the changes from #532 , it should be merged after #532,.

Now user could select the network when creating an account with custom path.

#### Creating Account with Custom Path

The network selector is enabled if we can not get the indicated network from the path. 

|derive a custom netwrok| select network| derive a known network|
|---|---|---|
|Simulator Screen Shot - iPhone SE - 2020-01-30 at 06 52 42|Simulator Screen Shot - iPhone SE - 2020-01-30 at 06 52 47|Simulator Screen Shot - iPhone SE - 2020-01-30 at 06 53 11|

#### View the account with cutom path

Previous Version shows account with custom path in a V3 account card without displaying the path. Now it is all changed to v4 path card.

| now | previous |
|----|----|
|Simulator Screen Shot - iPhone SE - 2020-01-30 at 07 15 22|Simulator Screen Shot - iPhone SE - 2020-01-30 at 07 26 50|

4ab41c5 : To support the further network update feature, it would be better to use `pathId` instead of `networkKey` to index the custom path account's network. 

Take Kusama network for example, it would be more frequent for a network to update its genesisHash rather than its pathId. 

### For account with custom path, meta data before: 

```javascript
      '//customNetwork//funding/1': {
        name: 'funding account1',
        address: "DwGa4E65nbGafaGirjMH5kXbtT7jHuAcSXHLLkTePnqw95p",
        createdAt: 1571068850409,
        updatedAt: 1571078850509,
      },
``` 

now:

```javascript
      '//customNetwork//funding/1': {
        name: 'funding account1',
        address: "DwGa4E65nbGafaGirjMH5kXbtT7jHuAcSXHLLkTePnqw95p",
        createdAt: 1571068850409,
        updatedAt: 1571078850509,
        networkPathId: "westend"
      },
```
 
If the pathId is not defined, we show the warning message as following:

IMG_0090


hanwencheng:
According to suggestion from @sjeohp :

8567d7ca5cbdd71058a75a839f4be1cd22ebd555 enables create a custom path after create a new identity.

IMG_BAFB4D5766BE-1

PR #538, opened 2020-01-30
chore: add name into account qr code

### LINKED ISSUE #534: Add name information into Account QR code
When adding an account from parity signer into polkadot.js, user has to input the name again to polkadot.js.

Currently the QR code for account is accountId which is the combination of:
```javascript
`${protocol}:${address}:${genesisHash}`
```

We could append the name to the end, so that the hot machine could get the name when importing the account, which would be like.

```javascript
`${protocol}:${address}:${genesisHash}:${name}`
```

This feature also needs the update on Polkadot.js and Mycrypto.

hanwencheng:
close #534 

please merge into master after #536 ,  #532

Self explained.

### DIFF
--- a/src/screens/AccountDetails.js
+++ b/src/screens/AccountDetails.js
@@ -110,7 +110,9 @@ function AccountDetails({ accounts, navigation }) {
 			/>
 			<View>
 				{protocol !== NetworkProtocols.UNKNOWN ? (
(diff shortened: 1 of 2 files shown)


hanwencheng:
Thanks for checking it out! @jacogr

PR #541, opened 2020-02-05
feat: Add westend network support

### LINKED ISSUE #466: refactor: GenericCall cleanup
https://github.com/polkadot-js/api/pull/1592/files#diff-8f9c1338eaf62f816d813b24392771cbR77

TLDR;
`const { meta, method, section } = GenericCall.findFunction(extrinsic.callIndex);`
becomes
```
import { TypeRegistry } from '@polkadot/types`;
const registry = new TypeRegistry();
const { meta, method, section } = registry.findMetaCall(extrinsic.callIndex);
```

### LINKED ISSUE #540: Add westend network support
Which include add westend metadata, network spec.

And also test if signing request can be correct preformed.

hanwencheng:
closes #540 closes #466

This PR has essential update for the release of Polkadot Network.

### Add Westend Support

Signing extrinsics have been test on iOS with Polkadot.js.

### Upgrade @polkadot.js related dependencies and React Native

`__dirname` is used across @polkadot packages, to prevent the version conflict of client side app and browser extension, here I have a custom babel plugin to rewrite all the `__dirname` into static path string value as a walkaround.

### How to test:

1. Create new network with `Westend` network, and create any account under it.
2. Signing extrinsics works for any account under `Westend` network.
3. At the same time, signing for kusama account should not be broken.


hanwencheng:
metadata of Kusama and Substrate_Dev would also need to updated, I would do them in another separate PR, since the polkadot.js APIs need to be updated at the same time.

PR #549, opened 2020-02-25
test: add e2e signing and ethereum tests

hanwencheng:
This PR improve current e2e tests.

### 1. Add more tests for Substrate signing 

Now we have two e2e suites: one is for key management and path derivation, the other is for signing test.

`e2e/signing.spec.ts`
 |--Single QR Extrinsic (Sign TX)
   |---SetRemark
   |---Transfer
 |--Multi Part QR Extrinsic(Sign Message)
   |---SetRemark

### 2. Add the Ethereum account tests: generation, deletion, and signing

`e2e/signing.spec.ts`
-Ethereum
 |--Sign Transaction
 |--Sign Message

`e2e/identityManipulation.spec.ts`
-Ethereum
 |--Account Creation
 |--Account Deletion

### 3. Reformat project layout for units test and e2e test

previously:
```
-src
-e2e
-specs
-jest.config.js
...
```

currently:
```
-src
-test
 |--e2e
   |---specs
   |---jest.config.js
   |---...
 |--unit
   |---specs
   |---jest.config.js
   |---...
 |--commonConfig.js
```

Since `Detox` and Unit test both use `Jest` as test runner, and they share similar parts of the configs, now move them together into the `test` folder, which brings a clear structure layout and configuration.

### 4. How to test

* `yarn` to upgrade dependencies
* Start react native server by `yarn start --reset-cache`
* Run `yarn e2e:ios` to test on ios, and `yarn e2e:android` for the android test suits. *notice: android e2e may be not stable with pin code input, sometime input is missing, just try again will fix the error*
* At the same time existed unit test should not be broken `yarn unit` should succeed.


hanwencheng:
TODO: Add ethereum account deletion test

PR #554, opened 2020-03-03
ci: use github actions

hanwencheng:
test gitlabs actions with @General-Beck


hanwencheng:
Thanks for your work! @General-Beck

PR #565, opened 2020-03-19
feat: add passworded identity

hanwencheng:
This PR add support for passworded account inside Identity.

### About password account:

This is an option for certainn account, if user add password when derive a new account, this password will be always needed when signing with this account. (even when we have enabled seed reference feature, user do not need to input pin, but he/she will still need to input password).

When user signing with a passworded account, UI will not give feedback if the password is correct or not. Only when user use it with the hot machine, hot machine will give the feedback if it is a wrong signature, which indicate that the password is wrong. For instance, when interact with Polkadot.js, when you give wrong password and correct pin, you will be able to generate the signature, but when Polkadot.js scan it, it will gives the error "badProof". By this, it will be harder for other people try to compromise the password by hand input.

### This pr inlcude:

* adding password
* refactor the project layout with use modules to group file for the certain functionality.
* use `try..catch..` to handling the errors during signing and account generation, and gives feedback to the UI.
* add e2e test for passworded account derivation and signing.

### Screenshots

|overview |  generation| signing|
|---|---|---|
|Simulator Screen Shot - iPhone 11 - 2020-04-04 at 20 45 10|Simulator Screen Shot - iPhone 11 - 2020-04-04 at 20 45 40|Simulator Screen Shot - iPhone 11 - 2020-04-04 at 20 46 00|


hanwencheng:
Github diff here is not correct, should not be so many changed files.

sjeohp:
- why only for derived accounts, why not root accounts?
- I can't hit `Next` after entering a password, see screenshot.

IMG_0B7F5B735D19-1

hanwencheng:
> * why only for derived accounts, why not root accounts?
> * I can't hit `Next` after entering a password, see screenshot.

It is possible, here you can't derive that because the path is not correct, it should start with `//` or `/`

sjeohp:
> * I can't hit `Next` after entering a password, see screenshot.

I just noticed the error text in the screenshot is black! virtually invisible.  

Perhaps it could be made more obvious.

sjeohp:
LGTM apart from the `Invalid Path` error message above and the conflict.

hanwencheng:
Fortunately you find this black text! otherwise we can't release it forever.

Now I use `TextInput` read bottom line to indicator the error, it looks like followings

|invalid path| valid path|
|---|---|
|Screen Shot 2020-04-20 at 21 43 13|Screen Shot 2020-04-20 at 21 43 27|

PR #570, opened 2020-03-25
fix: kusma extrinsic interpretion and signing error

### LINKED ISSUE #567: Wrong value displayed at Sign Transaction step
When signing a simple transfer of 1KSM, "value: 333.662μ KSM" is displayed.
IMG-5832

### LINKED ISSUE #569: Simple transfer on Kusama results in "Could not decode method with available metadata"
A simple transfer of 1KSM is now giving me a different error from the other issue #567 

I have tried sending a few accounts just 1KSM, and cannot verify the data on my device.

IMG_5833

hanwencheng:
closes #569 , closes #567

### Reason
1. In Kusama, the type definition of `Address` and `AccountId` is swaped, which brokes the previous extrinsic interpretion result.
2. React Native Camera need to be updated with latest React Native.

### Solution 
1. Use Override from `@polkadot.js/api` to override the [chain-specific type defintions](https://polkadot.js.org/api/start/types.extend.html#node-and-chain-specific-types)
2. Update React Native Version.

### Refractor included
Previously, each type we scan a extrinsic, we build the registry for interpret the extrinsic, which is not eco and takes extra time. Now all different kind of registry is stored into `RegistriesStore`, and only with the first scan for the certain network it is generated.

### Problem Remains
In the QR code, we get the extrinsic and destination address (in bytes). For get the related SS58 address we need to know which network this extrinsic belongs to, to get the network we need to first get genesisHash from the extrinsic. But to inteprete the extrinsic we need to know the network type.

It is not possible with current QR code content, we will need to included the network genesisHash in the QR code. For now, we just use a dumb registry to interprete the extrinsic.  But in the future, it may not be safe.

*So this PR would be a quick fix for now, in the future we need add at least `genesisHash` in the QR code. To interprete Extrinsic details, we still need `version` of ExtrinsicPayloads and `specVersion` of chain.*

Current solution.

```javascript
const registry = new TypeRegistry();
const extrinsicPayload = registry.createType('ExtrinsicPayload', rawPayload, {
    version: 4
});
const network = findNetwork(extrinsicPayload.genesisHash.toHex())
const accountID = encodeAddress(
    publicKeyAsBytes,
    network.prefix
); 
```

Display the extrinsic details is a nice-to-have feature,. But due to the fact Substrate Network is easy to upgrade, which will lead to the active upgrade of metadata, specVersion of the chain, types changes. They make Signer a active update App, we should plan to disable extrinsic information display in the coming version. Also it will simplify the flow.

### Commits

main work is in commit 424fd66 .

### How to test

Kusama transfer should work, information display should be correct.
Other networks extrinsic signing and interpretion should not be broken.


src/components/PayloadDetailsCard.tsx
```diff
@@ -45,31 +250,30 @@ interface Props {
...
+			fallback: !isKnownNetworkKey
 		};
 	}
 
 	render(): React.ReactElement {
 		const { fallback } = this.state;
 		const { description, payload, networkKey, signature, style } = this.props;
-		const prefix = SUBSTRATE_NETWORK_LIST[networkKey].prefix;
+
+		console.log('payload is', payload);
```
sjeohp:
```suggestion
```

hanwencheng:
Merged as it blocked all transactions now onn Kusama, need release this week.

PR #572, opened 2020-03-27
docs: update docs for 32-bit apple architecture

hanwencheng:
### Update docs for rust version

From 1.42.0 rust [dropped 32-bit apple target support](https://blog.rust-lang.org/2020/01/03/reducing-support-for-32-bit-apple-targets.html)), the latest version for building ios target is 1.41.1.

Update [building documenatation](https://github.com/paritytech/parity-signer/wiki/Building)

In order to manager the documentation in one place, remove the doc in READM.md and add links there.

### Remove i386 architecture target from rust.

As of Xcode 10 (10A255) - September 17, 2018, i386 architecture is no longer supported.

Also the version which use only the i386 target is iPhone 3GS from [here](https://developer.apple.com/library/archive/documentation/DeviceInformation/Reference/iOSDeviceCompatibility/DeviceCompatibilityMatrix/DeviceCompatibilityMatrix.html). iPhone 3GS only support up to iOS 6.1.6, so we should not consider it.


sjeohp:
It isn't just i386.  armv7 and armv7s (aka arm64) are also now tier 3 (essentially unsupported).  That will affect devices older than iPhone 5s, which could be a problem since the point of signer is to run on old devices.

hanwencheng:
The main reason we have to deprecate i386 is because Xcode 10 drop the support, even though we have i386 target compiled, it will not bundle into the app.

In terms of dropping armv7 and armv7s is a another discussion, so I refreshed my brain again here:

* Advantage: we can use latest rust toolchains, always safe to check on the new features of rust or docs. 
* Disadavantage, we will drop support for iPhone 4, 4s, 5c, 5. and always use rust toolchain 1.41.

I think it would be good if user could still use their old iPhone 4 or 5 (IMO these models especially 4 and 4s are very poplar I guess many people has the old one). And always using 1.41 should not be a big problem so far.

One day apple decide drop support armv7 and armv7s in Xcode, we will definitely drop support for them as well.

PR #574, opened 2020-03-30
fix: add backward compatibility for kusama account with different genesisHash

### LINKED ISSUE #573: Unable to parse transaction: No private key found for account ... Edgeware edge case
Please note this is an edge case and should not happen for most users.

I generated a private key for Edgeware mainnet via Parity Signer, however this was before the "Edgeware" chain was an option in Signer. So I generated a key in the Kusama chain with a derivation path of `//kusama/edg`

Now, when I attempt to access the funds I moved to that address, I am hit with the error: "Unable to parse transaction: No private key found for account .... found in your signer key storage for the Edgeware chain"

I am able to "log in" via the QR code and my balance shows with the correct address on-chain, however I cannot access the private key for this account. I am only looking for a workaround, as this is an edge case and should not occur for any other user due to the way I generated the account.

hanwencheng:
As a quick fix, it closes #573 

Add fault tolerance for the previous created account.

For example:
1. the account created with previous pathId `//kusama_CC2` now want to be used in `Kusama CC3` 
2. account supposed to be used for other unsupported network, but was created inside Kusama as temperary solution


hanwencheng:
closes as included in #575

PR #575, opened 2020-03-30
refactor: regroup the paths into their own network.

hanwencheng:
This PR refactor the way to categorize accounts under their network, is the extend of #574, it address #573 , more specifically:

1. The network created by custom path should be categorized into their own network.
2. Now we support user created any path with any network, though it is not encouraged. For example, user should be able to create account under Westend network with path `//kusama//any_custom_path` and explicitly pointing that using westend network.

also rename "Custom Network" back to "Unknown Network", since now only account created in the previous version without appointing any network is listed here. (They currently use Kusama, but the hint text ask them to recreate it)

Previously:

| Kusama| Custom Network |
|----|----|
|Only contain account path prefixed with `//kusama`|contains all the custom path|
|IMG_0097|IMG_0098|

After this fix:
| Kusama | Unknown Network |
|---|---|
|Contain account with kusama network appointed and account prefixed with `//kusama`| contain only legacy account created without network appointed| 
|IMG_0096|IMG_0099|

### DIFF
--- a/src/components/PathCard.tsx
+++ b/src/components/PathCard.tsx
@@ -24,9 +24,12 @@ import TouchableItem from './TouchableItem';
 import AccountPrefixedTitle from './AccountPrefixedTitle';
 
 import {
+	isSubstrateNetworkParams,
+	isUnknownNetworkParams
+} from 'types/networkSpecsTypes';
+import {
 	defaultNetworkKey,
 	NETWORK_LIST,
-	NetworkProtocols,
 	UnknownNetworkKeys
 } from 'constants/networkSpecs';
 import colors from 'styles/colors';
@@ -45,6 +48,7 @@ export default function PathCard({
 	identity,
 	path,
 	name,
+	networkKey,
 	testID,
 	titlePrefix
 }: {
@@ -52,6 +56,7 @@ export default function PathCard({
 	identity: Identity;
 	path: string;
 	name?: string;
+	networkKey?: string;
 	testID?: string;
 	titlePrefix?: string;
 }): React.ReactElement {
@@ -60,11 +65,12 @@ export default function PathCard({
 	const address = getAddressWithPath(path, identity);
 	const isUnknownAddress = address === '';
 
-	const networkKey = getNetworkKeyByPath(path);
-	const network =
-		networkKey === UnknownNetworkKeys.UNKNOWN && !isUnknownAddress
+	const computedNetworkKey =
+		networkKey || getNetworkKeyByPath(path, identity.meta.get(path)!);
+	const networkParams =
+		computedNetworkKey === UnknownNetworkKeys.UNKNOWN && !isUnknownAddress
 			? NETWORK_LIST[defaultNetworkKey]
-			: NETWORK_LIST[networkKey];
+			: NETWORK_LIST[computedNetworkKey];
 
 	const nonSubstrateCard = (
 		<View testID={testID}>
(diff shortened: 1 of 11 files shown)


android/app/build.gradle
```diff
@@ -165,9 +165,9 @@ android {
         release {
             // Caution! In production, you need to generate your own keystore file.
             // see https://facebook.github.io/react-native/docs/signed-apk-android.
-            // signingConfig signingConfigs.debug
-            // minifyEnabled enableProguardInReleaseBuilds
-            // proguardFiles getDefaultProguardFile("proguard-android.txt"), "proguard-rules.pro"
+//             signingConfig signingConfigs.debug
+             minifyEnabled enableProguardInReleaseBuilds
```
hanwencheng: These line helps to reduce the android package size

android/gradle.properties
```diff
@@ -21,5 +21,6 @@ MYAPP_RELEASE_STORE_FILE=apk.keystore
 MYAPP_RELEASE_KEY_ALIAS=testing
 MYAPP_RELEASE_STORE_PASSWORD=testing
 MYAPP_RELEASE_KEY_PASSWORD=testing
+android.enableR8=false
```
hanwencheng: use Proguard need to disable R8

src/types/networkSpecsTypes.ts
```diff
@@ -47,9 +47,9 @@ export function isSubstrateNetworkParams(
 		| UnknownNetworkParams
 		| EthereumNetworkParams
 ): networkParams is SubstrateNetworkParams {
+	const { protocol, pathId } = networkParams as SubstrateNetworkParams;
 	return (
-		(networkParams as SubstrateNetworkParams).protocol ===
-		NetworkProtocols.SUBSTRATE
+		protocol === NetworkProtocols.SUBSTRATE && pathId !== unknownNetworkPathId
```
sjeohp: This will also be true for `isUnknownNetworkParams` below.

src/types/networkSpecsTypes.ts
```diff
@@ -71,8 +71,10 @@ export function isUnknownNetworkParams(
 		| UnknownNetworkParams
 		| EthereumNetworkParams
 ): networkParams is UnknownNetworkParams {
+	const { protocol, pathId } = networkParams as SubstrateNetworkParams;
 	return (
-		(networkParams as UnknownNetworkParams).protocol ===
-		NetworkProtocols.UNKNOWN
+		(protocol === NetworkProtocols.SUBSTRATE &&
+			pathId !== unknownNetworkPathId) ||
```
sjeohp: This matches the same case as above.

sjeohp:
```suggestion
		(isSubstrateNetworkParams(networkParams) ||
```

sjeohp: The function name should be changed to `isSubstrateOrUnknownNetworkParams`, or else it should only be true for unknown.

src/components/PathCard.tsx
```diff
@@ -133,8 +143,8 @@ export default function PathCard({
 		</View>
 	);
 
-	return network.protocol === NetworkProtocols.SUBSTRATE ||
-		network.protocol === NetworkProtocols.UNKNOWN
+	return isSubstrateNetworkParams(networkParams) ||
+		isUnknownNetworkParams(networkParams)
```
sjeohp: This is equivalent to `return isUnknownNetworkParams(networkParams)`, see below.

src/screens/AccountNetworkChooser.tsx
```diff
@@ -240,9 +241,15 @@ function AccountNetworkChooser({
 			}
 		} else {
 			const paths = Array.from(currentIdentity!.meta.keys());
-			if (isSubstrateNetworkParams(networkParams)) {
-				const listedPaths = getPathsWithSubstrateNetworkKey(paths, networkKey);
-				if (listedPaths.length === 0)
+			if (
+				isSubstrateNetworkParams(networkParams) ||
+				isUnknownNetworkParams(networkParams)
```
sjeohp: This is equivalent to `if (isUnknownNetworkParams(networkParams)`, see below.

src/types/networkSpecsTypes.ts
```diff
@@ -71,8 +71,10 @@ export function isUnknownNetworkParams(
...
+		protocol === NetworkProtocols.UNKNOWN
```
sjeohp:
```suggestion
		(protocol === NetworkProtocols.UNKNOWN
```

src/types/networkSpecsTypes.ts
hanwencheng: It is still unknown network, but need to change `pathId !== unknownNetworkPathId` to `pathId === unknownNetworkPathId`

PR #578, opened 2020-03-30
build: use proguard and remove internet permission in android

### LINKED ISSUE #577: Remove internet permission on Android
As a app supposed to be use offline, the following permission in `AndroidManifest.xml` should be removed:

```xml
<uses-permission android:name="android.permission.INTERNET" />
```

hanwencheng:
closes #577 

This PR:
1. enable proguard for android to reduce package size 
2. remove internet permission on production app on Android, on Debug we stil need it to connect to React Native server.


hanwencheng:
Merged as not affect main code base

Issue #489, opened 2019-12-04
Use pointer to get seed from rust.


hanwencheng:
Currently (in v4 version), user need to input Pin to unlock seed each time seed phrase is needed, it avoids to store the clearText of seed phrase into app's state. Though we shorten the time that clear text value stored in memory, the data is still garbage collected automatically by Javascript engine and we can not force it. Furthermore, Such an unlocking process blocks some UI features like showing the address of the key pairs with each keyboard stroke, and bring more pin input prompt which raises controversial user experiences.

With biometric input from #416 , we still face the frequent authentication problem. In order to solve it and have the benefit of biometric data input in pin input, (i.e. minimal the exposure of seed phrase in Javascript runtime), we could keep the current seed phrase unlock process for the first time in the app's lifecycle, in addition, we return a pointer to the seed, without the seed itself crossing the native modules bridge.

The second time the user needs the seed phrase, he/she just use pointer to get the seed and do the signing function.

```javascript
  //first time use seed, navigate to pin input screen
  const seedPointer = await unlockSeedPhrase(navigation, currentIdentity);

  // second time use seed in app lifecycle
  const signedData = await signDataWithPointer(dataToSign, seedPointer);
```

Workflow would be like:

![UnlockWithPointer](https://user-images.githubusercontent.com/6014309/70170885-e687fd00-16cd-11ea-8663-2979ae852a12.png)

More details could be find from this [gist](https://github.com/hanwencheng/gists/blob/master/KeyPointer/KeyPointer.md)

PR #580, opened 2020-04-01
feat: sign with seed reference

sjeohp:
#489 

- `decryptDataRef` returns a reference instead of the seed itself
- `brainWalletSignWithRef` & `substrateSignWithRef` take a seed reference and sign
- `destroyDataRef` cleans up memory allocated by `decryptDataRef`

This PR adds the functions above, a future PR will make use of them.


hanwencheng:
Thanks for the PR! it seems you need to update eslitnt related dependencies with `yarn install` and then `yarn lint:fix` to fix the lint errors

rust/signer/src/lib.rs
```diff
@@ -229,6 +229,48 @@ export! {
...
+    }
+
+    @Java_io_parity_signer_EthkeyBridge_ethkeyDestroyDataRef
+    fn destroy_data_ref(data_ref: i64) -> () {
+        unsafe { Box::from_raw(data_ref as *mut String) };
+    }
+
+    @Java_io_parity_signer_EthkeyBridge_ethkeyBrainwalletSignWithRef
+    fn ethkey_brainwallet_sign_with_ref(seed_ref: i64, message: &str) -> Option<String> {
+        let seed = unsafe { Box::from_raw(seed_ref as *mut String) };
```
hanwencheng, line 249: If the seed did not got by the pointer, will be a panic here? 

Is there any way to catch the error here?

sjeohp: > If the seed did not got by the pointer, will be a panic here?
Yeah this is unsafe. If we destroy the pointer or pass the wrong number this behavior will be undefined.
> Is there any way to catch the error here?
No it will most likely just segfault. We have to prevent it on the typescript side with a safe abstraction.

PR #582, opened 2020-04-01
test: signing with seed reference

hanwencheng:
add unit test for #580

### DIFF
--- a/rust/signer/src/lib.rs
+++ b/rust/signer/src/lib.rs
@@ -14,9 +14,6 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
-//#![feature(trace_macros)]
-//trace_macros!(true);
-
 use bip39::{Language, Mnemonic, MnemonicType};
 use blake2_rfc::blake2b::blake2b;
 use ethsign::{keyfile::Crypto, Protected};
@@ -338,7 +335,7 @@ mod tests {
     #[test]
     fn test_substrate_sign() {
         let msg: String = b"Build The Future".to_hex();
-        let signature = dbg!(substrate_brainwallet_sign(SURI, &msg).unwrap());
+        let signature = substrate_brainwallet_sign(SURI, &msg).unwrap();
 
         let is_valid = schnorrkel_verify(SURI, &msg, &signature).unwrap();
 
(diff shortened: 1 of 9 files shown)


rust/signer/src/lib.rs
```diff
@@ -341,4 +342,31 @@ mod tests {
...
+    fn test_substrate_sign_with_ref() {
+        let msg: String = b"Build The Future".to_hex();
+        let data_pointer = decrypt_data_ref(ENCRYPTED_SEED, String::from(PIN)).unwrap();
+        let signature_by_ref = substrate_brainwallet_sign_with_ref(data_pointer, &msg).unwrap();
+        let is_valid = schnorrkel_verify(SURI, &msg, &signature_by_ref).unwrap();
+        assert!(is_valid);
+    }
+
+    #[test]
+    #[should_panic]
```
hanwencheng, line 356: This flag does not prevent `malloc` error on test, still make the test failed.

rust/signer/src/lib.rs
```diff
@@ -341,4 +342,31 @@ mod tests {
...
+    fn test_substrate_sign_with_ref_destroyed() {
+        let msg: String = b"Build The Future".to_hex();
+        let data_pointer = decrypt_data_ref(ENCRYPTED_SEED, String::from(PIN)).unwrap();
+        destroy_data_ref(data_pointer);
+        let signature_by_ref = substrate_brainwallet_sign_with_ref(data_pointer, &msg).unwrap();
+        let is_valid = schnorrkel_verify(SURI, &msg, &signature_by_ref).unwrap();
+        assert!(!is_valid);
```
hanwencheng, line 363: And tried by multiple times, it does not always get `malloc` error.

sjeohp: The behavior will be undefined, since it is accessing dropped memory, but it is not a guaranteed fault.

rust/signer/src/lib.rs
```diff
@@ -341,4 +342,31 @@ mod tests {
 
         assert!(is_valid);
     }
+
+    #[test]
+    fn test_substrate_sign_with_ref() {
+        let msg: String = b"Build The Future".to_hex();
+        let data_pointer = decrypt_data_ref(ENCRYPTED_SEED, String::from(PIN)).unwrap();
+        let signature_by_ref = substrate_brainwallet_sign_with_ref(data_pointer, &msg).unwrap();
```
sjeohp, line 350:
```suggestion
        let signature_by_ref = substrate_brainwallet_sign_with_ref(data_pointer, &msg).unwrap();
        destroy_data_ref(data_pointer);
```

rust/signer/src/lib.rs
sjeohp: I don't think we should test this.  It is testing a pattern that should never occur, ie. accessing dropped memory.

rust/signer/src/lib.rs
hanwencheng: put it to the end now

PR #583, opened 2020-04-02
fix: seed reference wrapper

sjeohp:
- add a wrapper object to safely manage signing with seed ref
- add react hooks to interface with UI


src/utils/native.ts
```diff
@@ -158,31 +158,57 @@ export function schnorrkelVerify(
...
-// Use a reference returned by decryptDataRef to sign a message
-export function brainWalletSignWithRef(
-	seedRef: number,
-	message: string
-): Promise<string> {
-	return EthkeyBridge.brainWalletSignWithRef(seedRef, message);
-}
+	// Decrypt a seed and store the reference. Must be called before signing.
+	tryCreate(encryptedSeed: string, password: string): Promise<void> {
+		if (this.valid) {
```
hanwencheng, line 172:
```suggestion
		if (!this.valid) {
```

hanwencheng:
I am not sure put it into a class is a good idea, becasue:
* it need to use `seed = new SeedRef()` everywhere, and they do not share the same instance. At least a Class Singleton should used.
* As it is not in store or hooks, the UI will not get update when the dataRef or valid is changed. For example, if I have "destroy poitner" toggle button on the UI, then I clicked the button, the pointer is destoryed, , the button should also gone. But actually here the UI will not notice that, and it will still shown.

hanwencheng:
add fix suggestion in #585

src/utils/native.ts
sjeohp: The line is correct as it is.  `tryCreate` should *fail* if the reference *is* valid.  If the reference is valid and we create another one, the old one will leak.

On the other hand, `tryDestroy` and signing functions should *fail* if the reference is *not* valid.

sjeohp:
I used a class because we want the data ref and signing functions to be as tightly coupled as possible.  I think your suggestion good.  I've merged #585 and made `SeedRefHooks` an interface for `SeedRef`.

This way we can avoid exporting the signing functions from `utils/native.ts` and instead export `SeedRef`, which always checks validity before calling them, making them much harder to use incorrectly (much harder to sign with an invalid reference).  In the UI we will always access them via `SeedRefHooks`.  Let me know what you think.

src/utils/native.ts
hanwencheng: yeah, my mistake

src/utils/seedRefHooks.ts
```diff
@@ -0,0 +1,43 @@
...
+
+type IsValidFunc = () => boolean;
+type TryCreateFunc = (encryptedSeed: string, password: string) => Promise<void>;
+type TryDestroyFunc = () => Promise<void>;
+type TrySignFunc = (message: string) => Promise<string>;
+
+type SeedRefHooks = [IsValidFunc, TryCreateFunc, TryDestroyFunc, TrySignFunc];
+
+export function useSeedRef(): SeedRefHooks {
+	const [seedRef] = useState<SeedRef>(new SeedRef());
```
hanwencheng, line 13: if the `setSeedRef` is not defined here and it is not in the function closure, I think `seedRef` will not be updated when the object changed. That might be a problem.

hanwencheng:
LGTM with latest changes, I think just need some changes to add `setSeedPhrase` there.

src/utils/seedRefHooks.ts
```diff
@@ -0,0 +1,43 @@
...
+	// Use the seed reference to sign a message. Will throw an error if
+	// `tryDestroy` has already been called or if `tryCreate` failed.
+	const brainWalletSign: TrySignFunc = function () {
+		return seedRef.tryBrainWalletSign(message);
+	};
+
+	// Use the seed reference to sign a message. Will throw an error if
+	// `tryDestroy` has already been called or if `tryCreate` failed.
+	const substrateSign: TrySignFunc = function () {
+		return seedRef.trySubstrateSign(message);
```
hanwencheng, line 39: message need to be passed here

src/utils/seedRefHooks.ts
```diff
@@ -0,0 +1,43 @@
...
+	// Destroy the decrypted seed. Must be called before this leaves scope or
+	// memory will leak.
+	const destroy: TryDestroyFunc = function () {
+		return seedRef.tryDestroy();
+	};
+
+	// Use the seed reference to sign a message. Will throw an error if
+	// `tryDestroy` has already been called or if `tryCreate` failed.
+	const brainWalletSign: TrySignFunc = function () {
+		return seedRef.tryBrainWalletSign(message);
```
hanwencheng, line 33: message need to be passed here

src/utils/seedRefHooks.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export function useSeedRef(): SeedRefHooks {
+	const [seedRef] = useState<SeedRef>(new SeedRef());
+
+	const isValid: IsValidFunc = function () {
+		return seedRef.isValid();
+	};
+
+	// Decrypt a seed and store the reference. Must be called before signing.
+	const create: TryCreateFunc = function (encryptedSeed, password) {
```
hanwencheng, line 20: types need to be specified...

src/utils/seedRefHooks.ts
```diff
@@ -0,0 +1,43 @@
+import { useState } from 'react';
+
+import { SeedRef } from 'utils/native';
+
+type IsValidFunc = () => boolean;
+type TryCreateFunc = (encryptedSeed: string, password: string) => Promise<void>;
+type TryDestroyFunc = () => Promise<void>;
+type TrySignFunc = (message: string) => Promise<string>;
+
+type SeedRefHooks = [IsValidFunc, TryCreateFunc, TryDestroyFunc, TrySignFunc];
```
hanwencheng, line 10: BTW here, what about changing them to a object/enum, because now to import sign function we need do 
```
const [isvalid, , , , substrateSign] = useSeedRef();
if(isvalid) {
 substrateSign(message)
}
```

with enum could be 
```
const [isValid, {substrateSign}] = useSeedRef();
if(isvalid) {
 substrateSign(message)
}
```

which could be more clear, BTW with the latter one you have to specify the types of each function, I just tried yesterdat...

PR #585, opened 2020-04-02
fix: wrap seed reference into react hooks

hanwencheng:
By using hooks: UI will notify and check if `valid` and `dataRef` existed.

In React component just get seedRef by 

```typescript
const [valid, seedRef, create, destory] = useSeedRef();
if(!valid){
  create(seedPhrase, password)
}
```


src/utils/seedRefHooks.ts
```diff
@@ -0,0 +1,31 @@
+import { useState } from 'react';
+
+import { decryptDataRef, destroyDataRef } from 'utils/native';
+
+type TryCreateFunc = (encryptedSeed: string, password: string) => Promise<void>;
+type TryDestroyFunc = () => Promise<void>;
+
+type SeedRefHooks = [number, boolean, TryCreateFunc, TryDestroyFunc];
```
sjeohp, line 8:
```suggestion
type SeedRefHooks = [number, boolean, TryCreateFunc, TryDestroyFunc, TryBrainWalletSign, TrySubstrateSign];
```

src/utils/seedRefHooks.ts
```diff
@@ -0,0 +1,31 @@
+import { useState } from 'react';
+
+import { decryptDataRef, destroyDataRef } from 'utils/native';
+
+type TryCreateFunc = (encryptedSeed: string, password: string) => Promise<void>;
+type TryDestroyFunc = () => Promise<void>;
+
```
sjeohp, line 7: type TrySign = (message: string) => Promise<string>;

sjeohp, line 7:
```suggestion
type TrySign = (message: string) => Promise<string>;

```

src/utils/seedRefHooks.ts
```diff
@@ -0,0 +1,31 @@
...
+	// Destroy the decrypted seed. Must be called before this leaves scope or
+	// memory will leak.
+	const destroy: TryDestroyFunc = function () {
+		return destroyDataRef(dataRef).then(() => {
+			setValid(false);
+		});
+	};
+
+	return [dataRef, valid, create, destroy];
```
sjeohp, line 30:
```suggestion
    // Use the seed reference to sign a message. Will throw an error if
    // `destroy` has already been called or if `create` failed.
    const brainWalletSign: TrySign = function (message) {
        if (!valid) {
            // Seed reference was never created or was already destroyed.
            throw new Error('cannot sign with an invalid seed reference');
        }
        return EthkeyBridge.brainWalletSignWithRef(dataRef, message);
    }

    // Use a reference returned by decryptDataRef to sign a message
    const substrateSign: TrySign = (message) {
        if (!valid) {
            // Seed reference was never created or was already destroyed.
            throw new Error('cannot sign with an invalid seed reference');
        }
        return EthkeyBridge.substrateSignWithRef(dataRef, message);
    }
    
	return [dataRef, valid, create, destroy, brainWalletSign, substrateSign];
```

sjeohp: Signing functions should be included to make sure `valid` is always checked before calling.

sjeohp:
Looks good.  I've added signing functions since we want them to be as tightly coupled as possible to `create` and `destroy`.

sjeohp:
I have included this in #583 with some additional changes.

hanwencheng:
Closed as included in #583

PR #586, opened 2020-04-02
fix: sync extend types with polkadot.js/api

hanwencheng:
the override types are migrated now from `@polkadot/types/known` to `@polkadot/types-known`.

So we need to adjust with it. Also the previous way to access types might be problematic since it is not supposed to be public, now use `getSpecTypes` function to get it.

Also need to adjust pathId with the chain name used in api.


src/stores/RegistriesStore.ts
```diff
@@ -15,26 +15,64 @@
...
+				specName = networkTypes.alias ?? networkName;
+				chainName = chainAlias ?? pathId;
+			} else {
+				return false;
+			}
+			return true;
+		}
+	);
+	// @ts-ignore
+	return getSpecTypes(registry, chainName, specName, '0xffffffff');
```
hanwencheng, line 68: the `0xffffffff` will be parsed into big number in api, so no problem here.

PR #590, opened 2020-04-08
feat: enable generate address with seed

hanwencheng:
This is a following PR for #580 

Which add support to generate address with seed reference.

Also for the function to sign with seed reference, add suri suffix as argument.  

This is because when we say save the seed. we save only the "seed phrase" of the identity, for signing and account generating, we all need to append the derivation path to the seed phrase and construct a suri. Then in the rust side, we use Suri to signing the transaction.


src/utils/native.ts
```diff
@@ -186,6 +186,30 @@ export class SeedRef {
 		);
 	}
 
+	trySubstrateAddress(suriSuffix: string, prefix: number): Promise<string> {
+		if (!this.valid) {
+			throw new Error('there is no seed reference ');
```
sjeohp, line 191:
```suggestion
			throw new Error('a seed reference has not been created');
```

src/utils/native.ts
```diff
@@ -186,6 +186,30 @@ export class SeedRef {
...
+			suriSuffix,
+			prefix
+		).then((address: string) => {
+			return address;
+		});
+	}
+
+	tryBrainWalletAddress(): Promise<string> {
+		if (!this.valid) {
+			throw new Error('there is no seed reference ');
```
sjeohp, line 204:
```suggestion
			throw new Error('a seed reference has not been created');
```

hanwencheng:
Merged as approved before.

PR #591, opened 2020-04-10
feat: add Centrifuge Mainnet and Centrifuge Testnet Amber

philipstanislaus:
Thanks for your great work on Parity Signer! We would love to use it for our mainnet and testnet ("Amber") as well - would that be possible?


src/stores/RegistriesStore.ts
```diff
@@ -34,7 +34,9 @@ type NetworkTypesMap = {
 const networkTypesMap: NetworkTypesMap = {
 	centrifuge: {
 		alias: 'centrifuge-chain',
-		chains: {}
+		chains: {
+			centrifuge_amber: 'centrifuge-chain-amber'
```
hanwencheng, line 38: It seems this is not record in the [API](https://github.com/polkadot-js/api/tree/master/packages/types-known/src) yet?

hanwencheng:
please also add icons to ios env, which is under `ios/assets/res` folder

src/stores/RegistriesStore.ts
philipstanislaus: If I read the code correctly, it uses `centrifuge` as the lookup in the API's known types. `centrifuge-chain-amber` might be used at a later point as we will use it for breaking changes. Do you think we should add separate types for `centrifuge-chain-amber` already to the API?

philipstanislaus:
Thanks for your review @hanwencheng . Just pushed the logos as well.

src/stores/RegistriesStore.ts
hanwencheng: understood, no need to add separate types for now, the types of `centrifuge-chain` will be included with this structure. If you amber chain has breaking changes then you may add them into API to overwrite the main net's types.

PR #592, opened 2020-04-14
fix: compilation error on android

sjeohp:
Fix android compilation errors.


hanwencheng:
Thanks for fixing that, that was my fault!

PR #594, opened 2020-04-14
fix: identities switch bottom shadow blocks button

### LINKED ISSUE #593: Identity at the very bottom of the identity list cannot be selected
[Most recent released version of Signer; Play Store says 30-03-2020.]

**iOS / iPhone SE**
After creating 5 identities, the one at the very bottom of the identities list isn't fully visible and cannot be selected.

**Android / Pixel 2**
After creating 5 identities, the one at the very bottom of the identities list isn't fully visible but can still be selected.

hanwencheng:
closes #593 

After fix:

out

@sjeohp I have also included a minor fix for the signature of Substrate sign function

### DIFF
--- a/src/components/IdentitiesSwitch.tsx
+++ b/src/components/IdentitiesSwitch.tsx
@@ -236,7 +236,7 @@ function IdentitiesSwitch({
 				{identities.length > 5 && (
 					<Separator
 						shadow={true}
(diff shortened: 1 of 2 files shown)


hanwencheng:
Also tested on Android, it all works as expected.

PR #595, opened 2020-04-16
fix: ios bridge

hanwencheng:
Fix the bridge error, that the method cannot be find in ios.


hanwencheng:
closed as included in #589

PR #598, opened 2020-04-17
chore: add chain metadata

hanwencheng:
Previous added metadata is not added into app logic.

This PR should fix.


hanwencheng:
Merged as a insubstantial fix.

PR #603, opened 2020-04-28
feat: simplify sign flow

hanwencheng:
Jump directly from scanning to the signing results.

Before This PR:

ScanningScreen
 -> TxDetails -> PinInput -> SignedTx
|-> MessageDetails -> PinInput -> Signed Message

Now the signing is 
ScanningScreen  
  -> PinInput -> SignedTx
| -> PinInput -> Signed Message


babel.config.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
+	ignore: ['node_modules/detox'],
```
hanwencheng, line 2: this line is used to prevent the e2e injections introducing detox into babel compiling, with this ignore, there will be an error if any app code rely on detox.

hanwencheng:
Merged for v4.2 release

PR #608, opened 2020-05-06
feat: Tab Navigation

goldsteinsveta:
[Design reference](https://www.figma.com/file/4odbGhaJ3OTN2ZQb52uvIy/Signer?node-id=5006%3A58)


goldsteinsveta:
Hey @hanwencheng I can't make the react-native-tabs to work, could you please take a look at it c91ff97? Please thank you!

Here's the error:

hanwencheng:
should be ok now, the `react-navigation/native` need to be upgraded.

goldsteinsveta:
Hey @hanwencheng, I've never worked with native navigation before. It's quite fun, but I am not sure I don't do anything questionable here.

Could you take a look at how it's nested in [screens.tsx](https://github.com/paritytech/parity-signer/blob/af15ac738f65a0ee1b02821223c40ff39099caa5/src/screens.tsx) and tell me 

1. if anything can go wrong here? and
2. what do I do to now fix this: *(signing testing QR w this is sparta)*



If you think that this is not the optimal navigation nesting, I am happy to know what will work best and/or change design etc. Thanks!

hanwencheng:
Hi, this error is caused by the reset navigation method. if you search `.reset` in `navigationHelpers.ts` you will find all the places reset the route. I guess tab navigation has different screen stacks, so if I reset screens from different tabs together, there will be an error.

You can jump from one tab to another tab with [this function](https://reactnavigation.org/docs/tab-based-navigation/#jumping-between-tabs)

But I think this [reset](https://reactnavigation.org/docs/navigation-actions/#reset) only works for the current tab.

PR #524, opened 2020-01-14
WIP: network offline branch rebase

hanwencheng:
It seems from the old commits from [network offline PR](https://github.com/paritytech/parity-signer/pull/429) is impossible to merge into master since it is outdated.

cherry picked commits in that PR and tried rebase them on master, now it would be easier to merge to master then.

PR #615, opened 2020-05-13
style: bottom QrScannerTab 

goldsteinsveta:
- [x] QrScanner bottom tab
- [x] Removes QR icon from the SecurityHeader 

QrScannerTab

[Design Reference](https://www.figma.com/file/B1MwtTfAXVQI5y9H41EovE/Signer-2.0?node-id=5006%3A58)


goldsteinsveta:
Hey @hanwencheng I decided not to change the whole navigation stack and go with super simple thing here. Let me know what you think?

And the Q: where do I put the definition for the 'react-native-navbar-color'? Thanks!

hanwencheng:
> where do I put the definition for the 'react-native-navbar-color'?

do you mean [here](https://github.com/paritytech/parity-signer/pull/615#diff-46ef2374da6211486f3a408db68b2bcbR43)? I think it would be fine to put it here. Or you need a file with some styling definitions, then may be put under `/src/styles/` folder. Also feel free to move any reusable styles there.

The design LGTM. For the tabs I think create the bottom tab bar is the first step. After the Merge of this PR, I can try to create another tab within network related pages from #524 .

PR #619, opened 2020-05-19
style: revamp / mop up ui

### LINKED ISSUE #624: Touch Target for "Sign Tx" is too small
I constantly miss the touch target of the QR Icon on my ipod touch. Given this is the most important often used button in the app I think it makes sense to make it a bit bigger.

goldsteinsveta:
continues #615  fixes #624 

[design reference](https://www.figma.com/file/B1MwtTfAXVQI5y9H41EovE/Signer-2.0?node-id=5006%3A58)

- [x] semantic colour names
- [x] new icons
- [x] QR > Tab
- [x] Parity pink
- [x] refactor styles and buttons on the way

Frame 3@3x_


goldsteinsveta:
@hanwencheng this is a continuation of the previous (now closed PR). Could you help me with the Detox test before I spend too much time on it?

hanwencheng:
@goldsteinsveta sure, e2e test is a pain, will check it soon.

hanwencheng:
should be fine now, need to add fonts to ios platform

hanwencheng:
on the ios simulator I cannot see the header bar as shown in the following screen.
Screen Shot 2020-05-25 at 16 21 20

hanwencheng:
And I see the bottom bar in the main screen, there is a small between the edge of screen and the QR Scanner bar, is it a expected UI? 

|bottom bar on PathList(do not have space)| bottom bar on Main screen(has space)|
|---|---|
|Screen Shot 2020-05-25 at 16 27 21|Screen Shot 2020-05-25 at 16 27 27|

gavofyork:
image

these two describe the same logical thing - could we merge them in some way?

goldsteinsveta:
@gavofyork yeah, that might be confusing. I've removed the path from header

hanwencheng:
Use `KeyboardAvoidView` is brilliant!

hanwencheng:
On iPhone SE

Screen Shot 2020-05-28 at 16 53 23

hanwencheng:
I find the white color contrast with black background is too sharp, so change the opacity of the font color, lmk what you think @goldsteinsveta , I will update it soon.

hanwencheng:
Play with `KeyboardAvoidView` for almost 1 hour, find has one backward part compared with React-native-keyboard-aware-scroll-view: it cannot dismiss it with click on the screen.

Also we need to make sure in the small screens the height does not covered any critical input related information, for example, when input the seed, need to make the suggestions words display uncovered. This is currently what I think of, will update the commit soon here.

hanwencheng:
Passworded account, the small lock icon is not aligned in vertical center.

Screen Shot 2020-05-28 at 17 17 49

src/components/Button.tsx
```diff
@@ -82,18 +83,19 @@ export default class Button extends React.PureComponent<{
 				disabled={disabled}
 				onPress={onPress}
 				testID={testID}
-				style={[finalButtonStyles, style]}
```
hanwencheng: Why change this, in this case in ios app the bottom buttton won't stick to the bottom. For example in the `PinInput`View.

hanwencheng:
find a nit here: after user signed a message, if it is a small screen, the qr code will not be complete, that means need to scroll down a bit, it would be better for them to directly use the screen for scanner without extra interaction. So I suggest put the QR code to center, message hash would not be a important message in most case, so it could be moved down.

Other part LGTM

Screen Shot 2020-05-29 at 01 56 01

src/components/Button.tsx
goldsteinsveta: Oh! Wow, I thought you've just reduced amount of boxes the there. Should've asked

hanwencheng:
6f84ac5 refactor `SignedTx.tsx`, actually what we need to change is `SignedMessage.tsx`.

goldsteinsveta:
@hanwencheng looking at tests for it now – I'll be able to pass them without changing the layout. The reason for signature QR to be below the extrinsic is a push for double check what is being signed. But I see the niceness of having QR straight there after scan

hanwencheng:
We want make use have least interaction for scan the QR code.

In the following picture user must need to scroll down for scanning. We would better make it as an optional choice if for scrolling. 

Screen Shot 2020-05-29 at 15 20 25

hanwencheng:
Thanks, tested on iOS and Android, looks good!

PR #628, opened 2020-05-26
fix: legacy account list white screen

### LINKED ISSUE #613: Legacy accounts page is broken on 4.2.2
When navigating to the "Legacy accounts" page via the top-right menu. A blank white screen appears.

iOS 13.4
Signer 4.2.2

hanwencheng:
fixes #613 

Unconfirmed if the fix works.

Since it happens on new iOS device,  we also need to update [safe-area-context](https://github.com/th3rdwave/react-native-safe-area-context) dependency.


hanwencheng:
issue fixed by #660, close this one.

PR #637, opened 2020-06-04
build: use prebuild ndk toolchains instead of building standalone one

hanwencheng:
As Android do not support building [standalone NDK](https://developer.android.com/ndk/guides/standalone_toolchain) since r20,  it would be safe and easiler for us to following the latest version and use the prebuild NDK toolchains.

This PR remove the script for building standalone NDK, and the now the current [building process](https://github.com/paritytech/parity-signer/wiki/Building) should also be changed as followings:

1. Download the latest NDK [here](https://developer.android.com/ndk/downloads), and set the environment variable `NDK_HOME` to the folder.
2. Download the SDK API (We use API29), if you want to use other API version, make sure you also change the `API_LEVEL` in `./scripts/variables.sh` file.
3. to set up the building environment, run `./scripts/init.sh` (for both Mac and Linux)


rust/signer/Makefile
```diff
@@ -1,7 +1,9 @@
-NDK_STANDALONE = $(shell pwd)/../../NDK
 ARCHS_IOS = x86_64-apple-ios armv7-apple-ios armv7s-apple-ios aarch64-apple-ios
 ARCHS_ANDROID = aarch64-linux-android armv7-linux-androideabi i686-linux-android
 LIB=libsigner.a
+API_LEVEL=29
```
hanwencheng, line 4: It would be good to get this value from `./scripts/variables.sh` file. AFAIK I can't do it, or probably we could use shell script instead of MakeFile here.

gww-parity: Does it help:

```
$ API_LEVEL="$(grep API_LEVEL scripts/variables.sh |tr -c -d '[:digit:]')"
$ echo $API_LEVEL 
29
```

?

hanwencheng: thank you! It works!

PR #639, opened 2020-06-05
docs: update readme for ndk

hanwencheng:
Self explained.

The `build.sh` file is a file that never used and used as reference for `rust/signer/Makefile`.


README.md
```diff
@@ -25,6 +25,9 @@ Any data transfer from or to the app happens using QR code. By doing so, the mos
 - [Troubleshooting](https://github.com/paritytech/parity-signer/wiki/Troubleshooting)
 - [Publishing](https://github.com/paritytech/parity-signer/wiki/Publishing)
 
+## Changes from 4.3.1
+From [4.3.1](https://github.com/paritytech/parity-signer/commit/ea5786c85661d9b176795b9386af640b3e73aff3) we use the latest prebuild NDK for building rust part library for android, so that need to build the standalone NDK toolchains manually. If you have built or develop Parity Signer before 4.3.1, please download the latest NDK [here](https://developer.android.com/ndk/downloads) and point the `NKD_HOME` environment variable to it.
```
Tbaut, line 29: > so that need to build the standalone NDK toolchains

^ Not sure I understand this part.

Using an example always help, something like `export NDK_HOME=/path/to/latest/ndk`

Also "latest" is something that changes. If I build the app now, I use the latest as of now. In 10 days from now will I need to download it again?

hanwencheng: some additional to the last point: IMO NDK is like the the version of `npm`, after r19 android has this major changes to include prebuilt toolchains, in the future we could all use them, so I think it will be always fine if user are using NDK version > r19.

PR #656, opened 2020-07-14
refactor: Use custom alert

### LINKED ISSUE #644: Use customized alert
React native default system alert have the following limitations:

1. can't dismiss by it self.
2. can't dismiss by click on the outside screen.
3. do not have consistent appearance.

Point 1 is important because we want some popup info only stay 1 second and dispear by itself. 

It would be nice to have a consistent alert UI design with more flexible functionalities.

hanwencheng:
closes #644

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -16,9 +16,8 @@
 
 import '../shim';
 import 'utils/iconLoader';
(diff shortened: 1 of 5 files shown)


hanwencheng:
closed as included in #662

PR #678, opened 2020-08-29
chore: update terminology transaction -> extrinsics

hanwencheng:


hanwencheng:
closed as included in #683 . For the terms and conditions, we would rather keep it more understandable with `transaction`.

PR #682, opened 2020-09-02
fix: add PayloadDetailsCards back

### LINKED ISSUE #680: iOS App doesn't display any info about what you are signing (e.g. Extrinsics)
Hello,

Not sure if this is already fixed and awaiting Apple App Store review but in the screenshots on the [sign a transaction tutorial](https://github.com/paritytech/parity-signer/blob/master/docs/tutorials/Kusama-tutorial.md#4-sign-a-transaction) which were taken on Android this information is displayed.

On the iOS version it is not displayed. Is there a reason for this?

Thanks

hanwencheng:
closes #680 

This PR add back Payload Details Cards to display the info of Extrinsic or Message.

Since Metadata of chains are very big, it is hard to update with QR code. We now only supports the pre-defined network which has its metadata: https://github.com/paritytech/parity-signer/blob/master/src/constants/networkMetadata.ts.

Parity Signer will not try to parse the details of Extrinsics of an imported network.

**NOTICE: This PR should only be merged after the following requirement is meet**

1. If Parity Signer use the old metadata to parsing the latest extrinsic, there is a possibility an error happen, for example, if the type of `AccountId` is changed in new Metadata on chain. To prevent such request, we also need include `specVersion` into the Extrinsic QR code on polkadot.js/apps.

I have prepared the code, the following two commits need to be merged into Polkadot.js **ui** and **apps** one by one.
https://github.com/hanwencheng/ui/pull/1
https://github.com/hanwencheng/apps/pull/1

and this PR need to be merged once the above two commits are merged into Polakdot.js
https://github.com/paritytech/parity-signer/pull/694

Old version parity signer app should not have problems in parsing the new QR code, but need to double-check

2. To always be able to support the latest chain specifications, the product owner of Parity Signer needs to regularly update @polkadot/api to get latest chain-specific types from `@polkadot/known-types` and then generate a TypeRegistry with correct metadata, as current code in `parity-signer/src/stores/RegistriesContext.ts`: 
```typescript
			const newRegistry = new TypeRegistry();
			const overrideTypes = getOverrideTypes(newRegistry, networkParams.pathId);
			newRegistry.register(overrideTypes);
			const metadata = new Metadata(newRegistry, networkMetadataRaw);
			newRegistry.setMetadata(metadata);
```

### Future improvement
* The styling could be update


src/utils/identitiesUtils.ts
```diff
@@ -42,12 +42,11 @@ import {
 import {
 	centrifugeAmberMetadata,
 	centrifugeMetadata,
-	defaultMetaData,
 	edgewareMetadata,
 	kulupuMetadata,
 	kusamaMetadata,
 	polkadotMetaData,
-	substrateDevMetadata,
```
dvdplm, line 50: Is this not supported/useful anymore? Why not?

test/e2e/specs/signing.spec.ts
```diff
@@ -84,13 +84,13 @@ describe('Signing ane exporting test', () => {
...
 		});
 
-		it('should sign transfer request', async () => {
-			await launchWithScanRequest(ScanTestRequest.TransferExtrinsicKusama);
-			await testSignedTx();
-		});
+		// it('should sign transfer request', async () => {
+		// 	await launchWithScanRequest(ScanTestRequest.TransferExtrinsicKusama);
+		// 	await testSignedTx();
+		// });
```
dvdplm, line 93: What is going on here? Should it be removed?

test/e2e/specs/signing.spec.ts
```diff
@@ -123,10 +123,10 @@ describe('Signing ane exporting test', () => {
...
 		});
 
-		it('should sign transfer request', async () => {
-			await launchWithScanRequest(ScanTestRequest.TransferExtrinsicPolkadot);
-			await testSignedTx();
-		});
+		// it('should sign transfer request', async () => {
+		// 	await launchWithScanRequest(ScanTestRequest.TransferExtrinsicPolkadot);
+		// 	await testSignedTx();
+		// });
```
dvdplm, line 129: Same

test/e2e/specs/signing.spec.ts
TarikGul: This is the one test which in the CI was causing a setState issue and would fail. That being said when running the test locally there is no issues, or errors. So for now I have it commented out. But am actively trying to figure it out. It is just incredibly hard to debug when the only time its erroring is in the CI.

src/utils/identitiesUtils.ts
TarikGul: I believe it was just replaced with rococo network instead.

Tbaut: This was used back then with a node running in  `- - dev` (the one hosted on dev-node.substrate.dev). It's still live, but barely updated and used. It's reset every hour, but last I checked it was not even running substrate v2, so you can forget about it and safely remove it.

test/e2e/specs/signing.spec.ts
dvdplm: Would be good to log an issue and add a ref to it here as a comment.

TarikGul:
Added [Issue#729](https://github.com/paritytech/parity-signer/issues/729) in regards to commenting out the two e2e tests that are failing in the CI but not failing locally.

src/modules/sign/hooks.ts
```diff
@@ -0,0 +1,43 @@
...
+					rawPayload,
+					{
+						version: ExtrinsicPayloadLatestVersion
+					}
+				);
+				setPayload(extrinsicPayload);
+				setIsProcessing(false);
+			} catch (e) {
+				//can't generate extrinsic payload, don't display.
+				console.log('error', e);
```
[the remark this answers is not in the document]
TarikGul: No they wont see this :)

test/e2e/specs/signing.spec.ts
Slesarew: This is fixed below now

src/modules/sign/screens/DetailsMessage.tsx
```diff
@@ -0,0 +1,110 @@
+// Copyright 2015-2020 Parity Technologies (UK) Ltd.
```
Slesarew, line 1: This file is not used at all right now, I realized I don't know exactly what to put here so it's not called from anywhere yet, but removing it is a bit painful while I hope to include it in final release (and start using it very soon) - let's leave it for now?

TarikGul: Yea I would leave it for now, I think that is totally fine.

src/modules/sign/hooks.ts
```diff
@@ -0,0 +1,43 @@
...
+	networkKey: string
+): [boolean, GenericExtrinsicPayload | null] {
+	const [payload, setPayload] = useState<GenericExtrinsicPayload | null>(null);
+	const [isProcessing, setIsProcessing] = useState<boolean>(false);
+	const { networks } = useContext(NetworksContext);
+	const { getTypeRegistry } = useContext(RegistriesContext);
+
+	useEffect(() => {
+		setIsProcessing(true);
+		if (getTypeRegistry === null) return;
```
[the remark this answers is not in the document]
TarikGul: The reason for this is because in the `catch` it is returning a null value in `useRegistriesStore` so I am assuming that is there to possibly avoid complaints from the typescript compiler.  but also too, if there is issues with the network's metadata it also returns null so I would assume its a way to sanity check any errors.

src/stores/ScannerContext.ts
```diff
@@ -356,7 +358,7 @@ export function useScannerContext(): ScannerContextState {
 			type: 'transaction'
 		};
 
-		setState(qrInfo);
+		setState({ ...qrInfo, rawPayload: txRequest.data.data });
```
TarikGul, line 361: We have talked about this sentiment before, but I think in the future comments and updated variable names could be good to avoid confusion. I assumed this is using a useReducer, and it is, but by looking at it right away and having the name be `setState` it could be confusing. I think this is just one overall example of a case where better variable naming could increasingly avoid confusion.

src/stores/ScannerContext.ts
```diff
@@ -500,7 +503,8 @@ export function useScannerContext(): ScannerContextState {
 			if (dataToSign instanceof GenericExtrinsicPayload) {
 				signable = u8aToHex(dataToSign.toU8a(true), -1, false);
 			} else if (isHash) {
-				signable = hexStripPrefix(dataToSign);
+				console.log('sign legacy data type is', typeof dataToSign);
```
TarikGul, line 506: Does this need to be in Prod?

TarikGul (approved):
LGTM, some comments about nit picky things, and also how we should comment and potentially do some in house cleanup but overall. Looks good to merge

PR #694, opened 2020-09-16
fix: add specVersion check

hanwencheng:
This PR need to be merged after #682 and the related code on Polkadot.js apps and ui.


Slesarew:
Redundant with #791

Issue #614, opened 2020-05-12
Add ed25519 support


hanwencheng:
As we discussed in #155 We still need to add Ed25519 support.

PR #720, opened 2021-02-08
Issue614

Slesarew:
feat!: add minimal support for multisignature Ed25519 algorithm, as mentioned in issue #614.

BREAKING CHANGE: upgrade is not possible, the app should be installed on a clean phone and identities re-generated as described in original manual

Ecdsa support as placeholder.

Refs #614


maciejhirsz:
Code looks good to me, can try to check the UI later. It would be nice to have some working test vectors in the rust codebase in RN substrate sign, plus the formatting there is a bit off.

What's the blocker from making this not backwards-compatible? Sr25519 is still supported after all. Anyone running signer on a hot phone with auto updates on will get their accounts bricked, which isn't exactly a nice UX, and wiping application data in iOS at least isn't something most users do.

Slesarew:
> Code looks good to me, can try to check the UI later. It would be nice to have some working test vectors in the rust codebase in RN substrate sign, plus the formatting there is a bit off.

Yeah, I really wanted some comments on this if it is going in the right direction first. This is minimally working code, but I would rather refactor the rust signer libs properly so that they are easy to read, with no repetitive code chunks and similar to Substrate reference libs. UI is also just bare minimum, only account details show what's happening - as far as I understood the ideas behind this app, the goal is to make it minimally distractive after all.

> What's the blocker from making this not backwards-compatible? Sr25519 is still supported after all. Anyone running signer on a hot phone with auto updates on will get their accounts bricked, which isn't exactly a nice UX, and wiping application data in iOS at least isn't something most users do.

Actually, just removal of couple trivial lines of code. I decided to leave it in this state after short talk with @kirushik - indeed ideally the app should not be upgraded linearly. But now I realize that the user should probably have freedom to violate this guideline.

Slesarew:
Obsolete; similar functionality already added in backend of staging branch

PR #732, opened 2021-03-05
fix: transaction is previewed before the key is unlocked and invoked

Slesarew:
Slightly more secure preview of the transaction, read it slowly and carefully without signed payload already available to be peeked. Only for transfers so far.


TarikGul (approved):
LGTM

PR #745, opened 2021-04-09
fix: pin validation when delete identity

hanwencheng:
Parity Signer does not validate the pin when deleting the selected identity.

> it happens when hits any pin and no need to confirm once the minimal size is reached.

This PR fixes the issue and add a test for checking the wrong pin code input when deleting the identity


hanwencheng:
e2e test failed due to the connection problem when installing the dependencies, need to be restarted manually.

Slesarew:
> e2e test failed due to the connection problem when installing the dependencies, need to be restarted manually.

restarted; the test script should be modified probably.

Slesarew:
RN removed; closing

PR #754, opened 2021-04-20
fix: Transaction parser

### LINKED ISSUE #739: Bonding transaction not parsed
Regular bonding procedure does not display sufficient number of details

https://wiki.polkadot.network/docs/en/maintain-guides-how-to-nominate-polkadot

Slesarew:
This PR closes #739 and many similar issues, but makes thing slightly uglier.

### DIFF
--- a/src/modules/sign/components/PayloadDetailsCard.tsx
+++ b/src/modules/sign/components/PayloadDetailsCard.tsx
@@ -14,9 +14,15 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
-import { GenericExtrinsicPayload, GenericCall } from '@polkadot/types';
+import { GenericExtrinsicPayload, GenericCall, Struct } from '@polkadot/types';
 import type { Call, ExtrinsicEra } from '@polkadot/types/interfaces';
-import { AnyJson, AnyU8a, IExtrinsicEra, IMethod } from '@polkadot/types/types';
+import {
+	AnyJson,
+	AnyU8a,
+	Codec,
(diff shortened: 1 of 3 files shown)


src/modules/sign/components/PayloadDetailsCard.tsx
```diff
@@ -175,58 +173,16 @@ const ExtrinsicPart = withRegistriesStore<ExtrinsicPartProps>(
...
-								the following arguments:
-							</Text>
-							{paramArgs ? (
-								paramArgs.map(([param, arg]) => (
-									<View key={param} style={styles.callDetails}>
-										<Text style={styles.titleText}>
-											{' { '}
-											{param}:{' '}
-											{arg && arg.length > 50
-												? shortString(arg)
```
Tbaut, line 213: I spend some time achieve the same on Stylo, although to be able to show the whole decoded payload, all I needed was to ditch this `shortString`, which was effectively shortening valuable information.

Tbaut:
Also FYI, in case you want to do something similar at some point, when we're signing an extrinsic whose payload length is > 256, this is considered `oversized`, and we end up only signing the hash of the Tx. Now on Signer, although we do have the info about the payload, we don't show it at all, all we show is the hash of the extrinsic that we are signing. I've removed this limitation in https://github.com/stylo-app/stylo/commit/a2f9f3b87f930b64e3d9c6d15669a79f67f64736 and not only can we show the whole payload (which is the goal of this PR afaiu), but we can also show what we're signing even if it's oversized. And of course, you still only sign the hash of the Tx.

Slesarew:
Let's merge this as is, we'll have another PR branch to make this pretty.

Issue #762, opened 2021-05-03
Polkadot-js support of signer


Slesarew:
Let's list here Signer's features not yet supported by hot wallets.

- [ ] message signing
~- [ ] metadata qr code video generation~

Slesarew:
Actually, I'm not sure about dynamic metadata generation - ideally it should be signed, and it also takes several minutes of time. We should probably just host pre-generated and signed updates and be done with it.

Slesarew:
I suppose, the companions can do it now, that's all we really need.

PR #773, opened 2021-05-20
feat: Support message signing

Tbaut:
addresses partially https://github.com/paritytech/parity-signer/issues/762

Working well with android using https://polkadot.js.org/apps/#/signing

This can be tested with a specific build of the extension https://github.com/polkadot-js/extension/pull/718

### DIFF
--- a/src/utils/decoders.ts
+++ b/src/utils/decoders.ts
@@ -32,6 +32,14 @@ import {
 } from 'types/scannerTypes';
 import { blake2b } from 'utils/native';
 
+// from https://github.com/maciejhirsz/uos#substrate-payload
+const CRYPTO_ED25519 = '00';
+const CRYPTO_SR25519 = '01';
+const CMD_SIGN_MORTAL = '00';
+const CMD_SIGN_HASH = '01';
+const CMD_SIGN_IMMORTAL = '02';
+const CMD_SIGN_MSG = '03';
+
 /*
  * @return strippedData: the rawBytes from react-native-camera, stripped of the ec11 padding to fill the frame size. See: decoders.js
  * N.B. Substrate oversized/multipart payloads will already be hashed at this point.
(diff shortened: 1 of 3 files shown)


src/utils/decoders.ts
```diff
@@ -200,10 +208,11 @@ export async function constructDataFromBytes(
 							); // encode to the prefix;
 
 							break;
-						case '01': // data is a hash
+						case CMD_SIGN_HASH:
+						case CMD_SIGN_MSG:
 							data.action = 'signData';
 							data.oversized = false;
-							data.isHash = true;
+							data.isHash = secondByte === CMD_SIGN_HASH ? true : false;
```
KiChjang, line 215:
```suggestion
							data.isHash = secondByte === CMD_SIGN_HASH;
```

Slesarew:
image

I'm getting this screen on polkadot.js.org

Tbaut:
This must be linked to the network, please check it out on Kusama for instance: https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fkusama-rpc.polkadot.io#/signing

src/utils/decoders.ts
```diff
@@ -140,9 +148,9 @@ export async function constructDataFromBytes(
 					data: {} // for consistency with legacy data format.
 				} as EthereumParsedData;
 				action =
-					firstByte === '00' || firstByte === '01'
+					firstByte === CRYPTO_ED25519 || firstByte === CRYPTO_SR25519
 						? 'signData'
-						: firstByte === '01'
+						: firstByte === CRYPTO_SR25519
```
KiChjang, line 153: How did this condition ever get hit before? If `firstByte` is equal to `'01'`, then it would've gone through the first condition and `action` would be `'signedData'`.

Tbaut: you're right, there's something fishy here. It's def. out of scope for this PR, but we should take a look and see if action being `signTransaction` is ever relevant. I guess it's not.

Slesarew: Exactly; looks like nobody uses Ethereum support in Signer anymore. I think it should be removed completely and soon.

Slesarew:
Redundant with #791

PR #785, opened 2021-07-16
fix: rn build android

Slesarew:
Disable apk minification; this makes apk huge, but it probably works at least.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -32,7 +32,7 @@ android-build-release:
     paths:
       - signer-app-release-signed.apk
   script:
-    - echo "$KEYSTORE_DATA" | base64 -d > /tmp/gitlab-ci.keystore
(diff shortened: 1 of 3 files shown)


Tbaut:
Have you had a major difference? I haven't! I just tried this trick though https://stackoverflow.com/a/68219448 and I did see a difference of ~10Mb

Slesarew:
> Have you had a major difference? I haven't! I just tried this trick though https://stackoverflow.com/a/68219448 and I did see a difference of ~10Mb

I'm not sure, I though I saw it grow more than twice, but then I realized I'm checking against quite old build.

Did you get a working debug artefact from this branch that could be loaded without metro? Can you email it to me? This would save me a lot of time. I just need a demo for designers and now I get all kinds of errors about JS and native incompatibility, with JS being of older version and typical fixes like rebooting console and updating things not working (yet).

I'll switch to ios for some time now, there are many major changes on that side that are badly overdue.

PR #789, opened 2021-07-28
Add chainType field and moonbeam ethereum networks

joelamouche:


Tbaut:
@joelamouche not sure if you know about [Stylo](https://github.com/stylo-app/stylo), it's my fork of Signer that is still in TS (and Rust for data signing), and still supports Ethereum. I would love it if we could make it usable for Moonriver, the main problem I see is that Metamask doesn't support external signers, and without this, there's no usage for Moonriver/Moonbeam. This feature has been a WIP for 2 years now :D https://github.com/MetaMask/metamask-extension/pull/6267

Slesarew:
@Tbaut please take this work in and let me know when I could close this and related issues; I'm a bit overwhelmed by irrelevant things that make tracking ongoing stuff here difficult but I want to help this initiative to survive.

Tbaut:
totally @Slesarew, if I were you, I'd close this since it'll never make it into Signer.

joelamouche:
@Tbaut No I didnt know about it. Looks interesting! So it's a live app on the store?

>the main problem I see is that Metamask doesn't support external signers, and without this, there's no usage for Moonriver/Moonbeam

Not sure what the problem is here. The idea was to use QR code based signing

Tbaut:
> Not sure what the problem is here. The idea was to use QR code based signing

If Moonbeam uses Metamask, and Metamask doesn't allow QR signing, that's the problem :)

joelamouche:
@Tbaut The polkadot-js app allows QR signing directly. So if Stylo can read a QR code and sign it, it can be used directly with the polkadot-js app

Tbaut:
Ah for this yes absolutely, btw I tried to reach out on Element, I think it'd work better for this conversation!

joelamouche:
Ok great

PR #793, opened 2021-08-24
major: Android native port

Slesarew:
Android app. Uses Jetpack Compose as framework.

- [x] Rust backend bridge
- [x] Onboarding
- [x] Basic navigation
- [x] Key management
- [x] Seeds encryption
- [x] Multiframe QR scanner
- [x] Transaction handling
- [x] History screen
- [ ] Secure network status indicator


vas3k (approved):
As Kirill said, doing a review for +36,911 -172,926 PR in a language you do not know is almost impossible, so we decided to make a test build and test all the necessary workflows: seed creation, metadata update, exporting, transaction parsing, transaction signing, extension integration

I spend some time testing it on Westend, all features seems to work well. UI is very ugly and buggy, as expected :D

vas3k:
However, it is still impossible to use the app without @Slesarew being there to tell you to "turn on the dark mode" or "restart the app to make camera work". We should make a note in the README that the version is unusable, because for now we decided to focus on the iOS version

Slesarew:
> However, it is still impossible to use the app without @Slesarew being there to tell you to "turn on the dark mode" or "restart the app to make camera work". We should make a note in the README that the version is unusable, because for now we decided to focus on the iOS version
> 
> Current builds in GooglePlay/AppStore will remain old until public release

https://github.com/paritytech/parity-signer/blob/ios-native-port/README.md#system-requirements

That PR will be merged first so both will have it.

PR #813, opened 2021-09-10
feat: docs and comments in transaction preview

Slesarew:
This is a set of small improvements to backend.

They should be implemented in UI

~- [ ] prepare to use metadata v14~
- [x] transaction preview can show documentation
- [x] when signing transaction, user can add comment to it

...and stuff in linked issues that happened to be added here as well


vas3k (approved):
Tested the last build from TestFlight, works for me. Noticed minor changes in the UI, no other visible changes (and bugs, which is good).

Issue #827, opened 2021-09-22
Zeroize secrets in rust code


Slesarew:
Check that everything is cleared properly before releasing.

Maybe some automatic testing? (should not be needed in zeroize abstractions are used)

Slesarew:
related issue reported https://github.com/paritytech/substrate/issues/9870

Slesarew:
This pretty much should be resolved by https://github.com/paritytech/substrate/issues/8089 and #844

PR #844, opened 2021-10-15
feat: db atomization and certificates logic

Slesarew:
Major backend modification.

- Database operations are now atomic; everything that requires confirmation is prepared before being shown and gets pushed in 1 action.
- Transaction parser modified to prepare the db changes in 1 atomic write; after-approval "signer" logic greatly simplified.
- Certificate trust logic brought to sensible state with 3 tiers - unverified, individually verified, general verifier.
- Onboarging in verified and jailbroken mode is available.
- Networks with individual network verifier could be locked if the certificate was compromised.
- Zeroizing confirmed everywhere, but is not implemented properly upstream (see #827 )
- Tests for backend in github action!
- Finally package rust code as workspace
- Address names effectively disabled
- Active alert indicator that disables seed creation on network connect
- Select between normal and certificate-less reset


vas3k (approved):
> +11,934 −25,169

LGTM 😆 

Ok, really, I can't test everything in here, I just tested the TestFlight build as usual: creating keys, importing to companion, signing transactions, updating metadata. We need to do a proper testing after you finish all the refactorings and we get proper UI mockups.

PR #873, opened 2021-11-08
feat: qr_reader_pc crate: opencv video input and output 

kosrk:
### Features:
1. Added video output for easy aiming at the QR code
2. Added a parser of program arguments for camera selection
3. Added tests (correctness of reading arguments and reading a single QR)
4. Added documentation (`readme.md` and internal)
5. The selected crates technically allow for cross-platform compatibility (not tested yet)

### New crates:
1. Video capture is based on `opencv` (because it supports more backends, than `rscam` and allows you to make the cross-platform application)
2. The video output is based on `opencv`

### The prospects:
It is possible to replace the `quircs` crate with an opencv qr detector (remove dependencies on crate `image` and improve performance), but so far it has a byte-type QR reading error (UTF-8 decoding error. с++ std::string -> rust String).
This can be solved by making changes to the opencv source code and requires creating an issue in the opencv repository. 
Issue has been created in opencv crate repository: https://github.com/twistedfall/opencv-rust/issues/304

### Notes:
This pull request has a higher priority than `qr_reader_pc crate: nokhwa video input and minifb video output` #875 pull request.

PR #875, opened 2021-11-08
feat: qr_reader_pc crate: nokhwa video input and minifb video output

kosrk:
### Notes:
This pull request has a lower priority than `feat: qr_reader_pc crate: opencv video input and output` #873 pull request. This PR can be considered as an alternative if opencv is not suitable for some reasons.

### Features:
1. Added video output for easy aiming at the QR code
2. Added a parser of program arguments for camera selection and configuration
3. Added tests (correctness of reading arguments and reading a single QR)
4. Added documentation (`readme.md` and internal)
5. The selected crates technically allow for cross-platform compatibility (not tested yet)

### New crates:
1. Video capture is based on `nokhwa` crate (because it supports more backends, than `rscam` and allows you to make the cross-platform application)
2. The video output is based on `minifb` crate


rust/qr_reader_pc/src/lib.rs
```diff
@@ -1,104 +1,229 @@
...
+    fps: u32,
+}
 
-    let mut camera = match Camera::new("/dev/video0") {
+/// Main cycle of video capture.
+/// Returns a string with decoded QR message in HEX format or error.
+///
+/// # Arguments
+///
+/// * `camera_settings` - CameraSettings struct that holds the camera parameters
```
Slesarew, line 35: Clean the camera buffer somewhere in the beginning and in the end; camera immediately parses last frame from previous scan resulting in instant "success".

kosrk: Nokhwa crate does not have an API for working with a camera buffer (and the frame read methods are blocking), but to avoid such a problem, I can add skipping few frames at the start of the program. In an alternative PR #873 there are more opportunities to solve this problem.

kosrk: 1b48054030ad98f363cbba779de5911b0295cd6f dropping frames at the beginning

PR #906, opened 2022-01-14
chore: Dependencies update 20220114

chevdor:


Slesarew (approved):
We need to do something about gitignoring rust/signer/Cargo.toml but still leaving it for CI plain `cargo test`; later

Slesarew (approved):
We need to do something about gitignoring rust/signer/Cargo.toml but still leaving it for CI plain `cargo test`; later

PR #910, opened 2022-01-17
test: test for presence of sufficient crypto in payload

Slesarew:
Just adds a test to make sure Signer can indeed sign its own metadata and export it as QR code. This is an example of how we can do that minimalistically.


Slesarew:
File rust/navigator/src/tests.rs has all the meaningful changes here

vas3k (approved):

PR #911, opened 2022-01-17
feat: android: screens (ports new app structure to android)

Slesarew:
Just add all required UI elements in Compose framework in this PR.


.github/workflows/rust-test.yml
```diff
@@ -20,4 +20,4 @@ jobs:
         run:            git submodule update --init --recursive
       - name:           Rust build
```
TriplEight, line 21:
```suggestion
      - name:           cargo test
```

.github/workflows/rust-test.yml
```diff
@@ -20,4 +20,4 @@ jobs:
...
         shell:          bash
-        run:            cd rust && cargo test && cd ..
+        run:            cd rust && cp os-specific-lock/android/Cargo.lock . && cargo test && cp os-specific-lock/ios/Cargo.lock . && cargo test && cd ..
```
TriplEight, line 23:
```suggestion
        run: |
          cd rust
          cp os-specific-lock/android/Cargo.lock .
          cargo test
          cp os-specific-lock/ios/Cargo.lock .
          cargo test
```

TriplEight: or maybe even divide them by two tests that would run in parallel

TriplEight (dismissed):
checked the CI part

.github/workflows/rust-test.yml
Slesarew: Done; now ios test is running on macos as ios will be built on that arch. Unfortunately, this flakes all tests with disk access operations. If I don't solve this soon, I'll move ios tests back to linux, but we should do something about this in longer term.

goldsteinsveta (approved):

.gitignore
```diff
@@ -84,6 +84,8 @@ ios/NativeSigner/Database/Database/*
 # rust
 rust/target
 rust/*/target
+rust/Cargo.lock
```
Vovke, line 87: Doesn't look like we should add Cargo.lock to .gitignore ([ref](https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html))

Slesarew: We should, see here

https://github.com/paritytech/parity-signer/blob/2fcbe3c23d2d217a11e334d55e11a14605b1a258/scripts/build.sh#L32

https://github.com/paritytech/parity-signer/blob/2fcbe3c23d2d217a11e334d55e11a14605b1a258/scripts/build.sh#L79

This is a hack, for Cargo does not support this target-specific building, and we have different Cargo.toml in final build for different archs, because android requires dynamic JNI interface and ios requires static C lib.

Slesarew: Although I'm planning to change this scripting system further soon, to separate build stages and introduce convenient `just` file to manage all this.

Vovke (approved):

PR #921, opened 2022-01-28
feat: backend: generate_message warns if metadata seems useless

Slesarew:
Small but useful feature: warning if metadata could not really be used to parse signable transactions due to incomplete extensions set


rust/definitions/src/metadata.rs
```diff
@@ -176,7 +184,23 @@ pub fn runtime_metadata_from_vec(meta_vec: &Vec<u8>) -> Result<RuntimeMetadata,
...
-/// Struct for processing 
+/// Function to check if the v14 metadata has all signed extensions required for transaction decoding.
+/// True if extensions are incomplete.
+/// Currently, the decoding of the transaction demands that metadata version, network genesis hash,
+/// and era are among signed extensions. Otherwise, a ParserMetadataError would occur on decoding.
+/// However, we can not simply forbid the loading of the metadata without required set of
+/// signed extensions into Signer.
+/// This function should be used for warnings only on generate_message side and during metadata
+/// loading into Signer.
+fn v14_warn (metadata_v14: &RuntimeMetadataV14) -> bool {
```
prybalko, line 195: Looking at the function name only, it's not obvious that it returns `bool`. Maybe `is_v14_signed` or `is_v14_ok` would be a better option

prybalko (approved):

rust/definitions/src/metadata.rs
Slesarew: Changed this below and also used hashmap to catch duplicates of extensions (paranoid but important!); please have another look

prybalko (approved):

Issue #920, opened 2022-01-28
Off by one issue with progress bar when scanning multipart gif


pepyakin:
When reading one of those metadata gifs I noticed that the progress bar may overshoot. For example, with a gif with 250 parts the scanning would finish when read either 251 or 252 parts.

That's on iOS with the App Store version.

Slesarew:
This is inevitable - raptor erasure codes have probabilistic nature, there is some chance that you'll need slightly more frames. Although the number shown now is an estimate either way, we can tweak it up or down a bit. Maybe we should bump it up a few numbers, just to lower expectations?

pepyakin:
I guess.

Can't we use percentage though?

Slesarew:
We can, but then we'll have something like "101%" in unfortunate cases, which is worse I suppose; and we can't predict how many frames are needed because it depends on which frames are captured next. I had % in early drafts of this interface, frame count *feels* more informative somehow.

pepyakin:
> We can, but then we'll have something like "101%" in unfortunate cases,

It's not necessary though: the label can be clamped to 0-100% or even 0-99%.

Slesarew:
>clamped to 0-100% or even 0-99%

That would not be informative though; I've seen once a corrupted transaction that just did kept counting frames; without proper indication just how much extra frames were caught, this would cause user to suspect that every video that slowed down near completion is corrupt and cause needless frustration, especially as metadata size grows. We should eventually discover some optimal mechanism to auto-reset failed transactions, but that should be based on more testing and measurement and probably on user research.

Maybe we should indicate better that target frame count is just an estimate? @goldsteinsveta

goldsteinsveta:
I'd suggest
- to keep the frame count
- add % as a subtitle, clamped to 99%, same for the width of the bar
- add units to what we measure with frame count and percentage, so that 252/251 does not look too weird and 98% does not overpromise how soon the scanning will be done.

<img width="320" alt="ContextMenu - Key" src="https://user-images.githubusercontent.com/8782596/153062008-06d95754-abd3-4007-86b5-168c13c380d8.png">

Happy to learn more about the QR movies to replace the placeholder strings!

Slesarew:
> add % as a subtitle
@goldsteinsveta  I liked the idea initially, but now I just realized that this is showing too much to user without really telling anything new - it's 3rd indicator for the same value on screen.

[...]

PR #924, opened 2022-02-01
fix: backend: increase presented expected frames count

Slesarew:
Issue #920


goldsteinsveta (dismissed):

goldsteinsveta:
While we here, could you wrap the multipart status menu into the same styles as i.g. "Add Seed" menu? Thanks!

goldsteinsveta (approved):

PR #929, opened 2022-02-16
feat: backend: seed recovery logic moved to rust

varovainen:
Seed recovery screen logic moved to rust.
- [x] backend change
- [x] Android
- [x] iOS


goldsteinsveta (changes requested):
**On connected devices:**

- [x] Android – Generate seed, recover seed, backup seed and derive key should not be possible. It's necessary to show "Network connected" error on calling these functions
- [x] iOS, Android – Backup seed – Error message is unclear. Please show "`Network connected! Seeds are not available now. Please enable airplane mode and disconnect all cables to access the seed phrase`"
- [x] iOS, Android – Backup seed – Seed phrase container, also in the error state, should take the full width, that's optional but it would be so nice!

goldsteinsveta (approved):

PR #936, opened 2022-02-28
build: add justfile

Slesarew:


justfile
```diff
@@ -0,0 +1,103 @@
...
+    ./build.sh android
+    cd ..
+
+build_ios:
+    #!/usr/bin/env bash
+    cd scripts
+    ./build.sh ios
+    cd ..
+
+# build all - this is probably impossible
```
TriplEight, line 35: it's possible https://cheatography.com/linux-china/cheat-sheets/justfile/
image

TriplEight (changes requested):
It's not clear to me why do you need another CI system

justfile
Slesarew: yes, these commands could be called of course, but we do not have a toolchain combination that will really build both on any platform yet. iOS build requires macos for which we havent's established or tested any android toolchain at the moment. Hopefully this will change.

justfile
```diff
@@ -0,0 +1,103 @@
...
+#build: build_ios build_android
+
+#do all automatable tests for ios
+test_rust_ios: switch_to_ios
+    #!/usr/bin/env bash
+    cd rust
+    cargo test
+
+#do all automatable tests 
+test_rust_android: switch_to_android
```
chevdor, line 45:
```suggestion
test_rust_android: _switch_to_android
```

montekki:
CI is ok atm, it is unlikely that this will be used in the future

PR #938, opened 2022-02-28
fix: Fix the docs publishing

TriplEight:
- Now https://paritytech.github.io/parity-signer/ is available and serving the docs
- previous jobs are being cancelled
- rust jobs are cached
- cargo installs are cached
- `cargo nextest` is used instead of `cargo test`

some practices are described here https://github.com/paritytech/ci_cd/wiki/HowTo:-GitHub-Actions-best-practices


Slesarew (approved):

TriplEight:
Note: https://github.com/paritytech/parity-signer/runs/5365103611?check_suite_focus=true#step:15:2 and https://github.com/paritytech/parity-signer/runs/5365103609?check_suite_focus=true#step:15:2 might be caused by `cd`

Slesarew (approved):

PR #955, opened 2022-03-16
fix: work around to launch camer for caption on mac

EgorPopelyaev:
As discussed with Alexander @Slesarew here is a pr for the workaround to be able to launch camera for reading. Initial issue I was facing on my macbook was that qc_reader_pc was not able to start failing with the following error: 
Error: "QR reading error. Zero frame size."

### DIFF
--- a/rust/qr_reader_pc/src/lib.rs
+++ b/rust/qr_reader_pc/src/lib.rs
@@ -101,7 +101,7 @@ fn create_camera(camera_index: i32, width: u32, height: u32) -> anyhow::Result<v
     let mut frame = Mat::default();
 
     match camera.read(&mut frame) {
-        Ok(_) if frame.size()?.width > 0 => Ok(camera),
+        Ok(_) if frame.size()?.width >= 0 => Ok(camera),
         Ok(_) => Err(anyhow!("Zero frame size.")),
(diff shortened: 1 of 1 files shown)


prybalko:
`qr_reader_pc` works well on my mac as is. 
Does `frame.size()?.width` always return zero for you? Is it a known opencv issue?

EgorPopelyaev:
@prybalko Yep, I was not able to run it at all on locally. I haven't found any related issues for opencv so far. I have macOS Monterey 12.2.1 and opencv 4.54 locally

prybalko:
I've also got Monterey 12.2.1, but opencv 4.5.5 (installed it from brew). I can't replicate the issue so far

EgorPopelyaev:
I'll upgrade my mac (there is a new version of OS came up) and opencv over the weekend and test it again. It can be easily that something is wrong with my setup

Slesarew:
> I'll upgrade my mac (there is a new version of OS came up) and opencv over the weekend and test it again. It can be easily that something is wrong with my setup

Any progress on this?

EgorPopelyaev:
Looks like the issue is with the mac's on Intel architecture. I moved yesterday to the new one on M1 and the problem has gone. So, I'll this PR for now.

PR #962, opened 2022-03-21
refactor: code cleanup, docs, bumping dependencies

varovainen:


rust/constants/Cargo.toml
```diff
@@ -1,13 +1,19 @@
 [package]
 name = "constants"
 version = "0.1.0"
-authors = ["vera"]
```
montekki, line 4: afair it has to say
```toml
authors = ["Parity Technologies <admin@parity.io>"]
```
everywhere

rust/constants/src/test_values.rs
```diff
@@ -0,0 +1,117 @@
+/// Identicons and related pictures, hex encoded vector of png data
+///
+/// Empty 30x30 transparent png image,
+/// used in cases when identicon generation failed or public key does not exist
+pub const EMPTY_PNG: &str = "89504e470d0a1a0a0000000d494844520000001e0000001e08060000003b30aea20000002e49444154789cedcd410100200c0021ed1f7ab6381f8302dc99393f8833e28c3823ce8833e28c3823ce8833fbe20724cf59c50a861d5c0000000049454e44ae426082";
```
montekki, line 5: either `include_str` or `include_bytes` and convert to necessary data in code here and in all the following constants.

rust/signer/src/lib.rs
```diff
@@ -59,16 +59,6 @@ export! {
...
 
-    @Java_io_parity_signer_models_SignerDataModel_substratePathCheck
-    fn path_check(
-        seed_name: &str,
-        path: &str,
-        network: &str,
-        dbname: &str
-    ) -> String {
-        db_handling::interface_signer::dynamic_path_check(dbname, seed_name, path, network)
-    }
```
montekki, line 70: It looks like this breaks the iOS build, take a look at the error, check the CI log:

```
❌  Undefined symbols for architecture x86_64
> Symbol: _path_check
> Referenced from: NativeSigner.MDeriveKey.updateDerivationCheck(path: Swift.String, dbName: Swift.String) -> NativeSigner.DerivationCheck? in MDeriveKey.o


❌  ld: symbol(s) not found for architecture x86_64


❌  clang: error: linker command failed with exit code 1 (use -v to see invocation)

** BUILD FAILED **
```

Slesarew: Yes, this is result of merge error; should just be reverted

rust/constants/Cargo.toml
```diff
@@ -1,12 +1,19 @@
...
-edition = "2018"
+edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+hex = {version = "0.4.3", optional = true}
```
montekki, line 9: looks like `hex` can be a [dev dependency](https://doc.rust-lang.org/rust-by-example/testing/dev_dependencies.html) so there would be no need for this optional feature

rust/db_handling/Cargo.toml
```diff
@@ -1,30 +1,34 @@
...
+sp-runtime = {git = "https://github.com/paritytech/substrate", default-features = false, optional = true}
+tiny-bip39 = {version = "0.8.0", default-features = false, optional = true}
+zeroize = {version = "1.5.3", optional = true}
+
+[features]
+default = ["test"]
+
+active = ["constants/active", "defaults/active", "definitions/active", "sp-core", "sp-runtime", "zeroize"]
+signer = ["blake2-rfc", "constants/signer", "defaults/signer", "definitions/signer", "hex", "plot_icon", "qrcode_static", "tiny-bip39", "sp-core", "sp-runtime", "zeroize"]
+test = ["active", "constants/test", "signer"]
```
montekki, line 31: It would be better to break into multiple lines with one feature per line, see [this](https://github.com/paritytech/substrate/blob/master/frame/grandpa/Cargo.toml#L46) for example

montekki (approved):

rust/constants/Cargo.toml
varovainen: `test_values.rs` contents are used in other crates tests (`db_handling`, `navigator`, `transaction_parsing`, `transaction_signing`), so the `hex` propagation is needed, I suppose. Maybe we should rename the feature from `test` to `integration_test` or something, it's really confusing.

Slesarew (approved):

PR #982, opened 2022-03-30
chore: move android CI from gitlab to GHA

montekki:


Slesarew:
After this is done, @goldsteinsveta should get an apk from #929 so that we can merge it and move forward

Slesarew (approved):

PR #987, opened 2022-04-04
refactor: run clippy on code and add a github workflow

montekki:
Another part of #962. This is the `clippy` part.


montekki:
I have placed a couple of `allow` directives in this one in places that would require a deeper refactor.

Slesarew (approved):

PR #989, opened 2022-04-06
chore: adds cargo deny workflow

montekki:
I think it would be a good idea to check at least the advisories.


Slesarew:
Let's fix and merge this after #962, merging anything there becomes too challenging already

.github/workflows/rust-cargo-deny.yml
```diff
@@ -0,0 +1,29 @@
...
+    strategy:
+      matrix:
+        checks:
+          - advisories
+
+    continue-on-error:    ${{ matrix.checks == 'advisories' }}
+
+    steps:
+      - name:               Checkout Sources
+        uses:               actions/checkout@v2
```
sergejparity, line 23:
```suggestion
        uses:               actions/checkout@v3
```
I suggest to use latest version of this GHA

Slesarew (approved):

PR #999, opened 2022-04-13
feature: uniffi: initial commit

montekki:
An attempt to use [uniffi-rs](https://mozilla.github.io/uniffi-rs/) in order to create a typesafe and compact interface between UI and rust backend

### DIFF
--- a/rust/db_handling/Cargo.toml
+++ b/rust/db_handling/Cargo.toml
@@ -19,7 +19,6 @@ qrcode_static = {path = "../qrcode_static", optional = true}
 regex = "1.5.4"
 sled = "0.34.6"
 sp-core = {git = "https://github.com/paritytech/substrate", default-features = false, features = ["full_crypto"], optional = true}
-sp-runtime = {git = "https://github.com/paritytech/substrate", default-features = false, optional = true}
 tiny-bip39 = {version = "0.8.0", default-features = false, optional = true}
 zeroize = {version = "1.5.3", optional = true}
 
@@ -30,7 +29,6 @@ active = [
 	"defaults/active",
 	"definitions/active",
 	"sp-core",
-	"sp-runtime",
 	"zeroize",
 ]
 signer = [
@@ -43,7 +41,6 @@ signer = [
 	"qrcode_static",
 	"tiny-bip39",
 	"sp-core",
-	"sp-runtime",
 	"zeroize",
 ]
 test = [
--- a/rust/db_handling/src/interface_signer.rs
+++ b/rust/db_handling/src/interface_signer.rs
@@ -4,12 +4,12 @@ use hex;
 use parity_scale_codec::Encode;
 use plot_icon::EMPTY_PNG;
 use sp_core::{sr25519, Pair};
-use sp_runtime::MultiSigner;
 use std::collections::HashMap;
 use zeroize::{Zeroize, ZeroizeOnDrop};
 
 use constants::{HISTORY, MAX_WORDS_DISPLAY, TRANSACTION};
 use definitions::{
+    crypto::MultiSigner,
     error::{AddressGenerationCommon, ErrorSource},
     error_signer::{DatabaseSigner, ErrorSigner, InterfaceSigner, NotFoundSigner, Signer},
     helpers::{
@@ -99,9 +99,9 @@ fn preferred_multisigner_identicon(multisigner_set: &[MultiSigner]) -> String {
         let mut got_ecdsa = None;
         for x in multisigner_set.iter() {
             match x {
-                MultiSigner::Ed25519(_) => got_ed25519 = Some(x.to_owned()),
-                MultiSigner::Sr25519(_) => got_sr25519 = Some(x.to_owned()),
-                MultiSigner::Ecdsa(_) => got_ecdsa = Some(x.to_owned()),
+                MultiSigner::Ed25519 { .. } => got_ed25519 = Some(x.to_owned()),
+                MultiSigner::Sr25519 { .. } => got_sr25519 = Some(x.to_owned()),
+                MultiSigner::Ecdsa { .. } => got_ecdsa = Some(x.to_owned()),
             }
         }
         if let Some(a) = got_sr25519 {
@@ -483,9 +483,9 @@ pub fn print_new_seed(seed_name: &str) -> Result<String, ErrorSigner> {
             ))
         }
     };
-    let hex_identicon = hex::encode(make_identicon_from_multisigner(&MultiSigner::Sr25519(
-        sr25519_pair.public(),
-    )));
+    let hex_identicon = hex::encode(make_identicon_from_multisigner(&MultiSigner::Sr25519 {
+        public: sr25519_pair.public().into(),
+    }));
     Ok(format!(
         "\"seed\":\"{}\",\"seed_phrase\":\"{}\",\"identicon\":\"{}\"",
         seed_name, seed_phrase, hex_identicon
--- a/rust/definitions/src/history.rs
+++ b/rust/definitions/src/history.rs
@@ -20,12 +20,11 @@
 use blake2_rfc::blake2b::blake2b;
 use parity_scale_codec::{Decode, Encode};
 use sled::IVec;
-use sp_runtime::MultiSigner;
 #[cfg(feature = "signer")]
 use std::convert::TryInto;
 
 use crate::{
-    crypto::Encryption,
+    crypto::{Encryption, MultiSigner},
     keyring::VerifierKey,
     metadata::MetaValues,
     network_specs::{
@@ -44,9 +43,9 @@ use crate::{
 /// Contains network name, network version, metadata hash.
 #[derive(Decode, Encode, PartialEq, Clone)]
 pub struct MetaValuesDisplay {
-    name: String,
-    version: u32,
-    meta_hash: Vec<u8>,
+    pub name: String,
+    pub version: u32,
+    pub meta_hash: Vec<u8>,
 }
 
 impl MetaValuesDisplay {
@@ -88,10 +87,10 @@ impl MetaValuesDisplay {
 /// of address used for `SufficientCrypto` generation.  
 #[derive(Decode, Encode, PartialEq, Clone)]
 pub struct MetaValuesExport {
-    name: String,
-    version: u32,
-    meta_hash: Vec<u8>,
-    signed_by: VerifierValue,
+    pub name: String,
+    pub version: u32,
+    pub meta_hash: Vec<u8>,
+    pub signed_by: VerifierValue,
 }
 
 impl MetaValuesExport {
@@ -117,9 +116,9 @@ impl MetaValuesExport {
 /// Event content for importing or removing network specs  
 #[derive(Decode, Encode, PartialEq, Clone)]
 pub struct NetworkSpecsDisplay {
-    specs: NetworkSpecs,
-    valid_current_verifier: ValidCurrentVerifier,
-    general_verifier: Verifier,
+    pub specs: NetworkSpecs,
+    pub valid_current_verifier: ValidCurrentVerifier,
+    pub general_verifier: Verifier,
 }
 
 impl NetworkSpecsDisplay {
@@ -154,8 +153,8 @@ impl NetworkSpecsDisplay {
 /// `SufficientCrypto` generation.  
 #[derive(Decode, Encode, PartialEq, Clone)]
 pub struct NetworkSpecsExport {
-    specs_to_send: NetworkSpecsToSend,
-    signed_by: VerifierValue,
+    pub specs_to_send: NetworkSpecsToSend,
+    pub signed_by: VerifierValue,
 }
 
 impl NetworkSpecsExport {
@@ -182,9 +181,9 @@ impl NetworkSpecsExport {
 /// Event content for setting network verifier
 #[derive(Decode, Encode, PartialEq, Clone)]
 pub struct NetworkVerifierDisplay {
-    genesis_hash: Vec<u8>,
-    valid_current_verifier: ValidCurrentVerifier,
-    general_verifier: Verifier,
+    pub genesis_hash: Vec<u8>,
+    pub valid_current_verifier: ValidCurrentVerifier,
+    pub general_verifier: Verifier,
 }
 
 impl NetworkVerifierDisplay {
@@ -220,8 +219,8 @@ impl NetworkVerifierDisplay {
 /// Contains hash of SCALE-encoded types data and types information [`Verifier`].
 #[derive(Decode, Encode, PartialEq, Clone)]
 pub struct TypesDisplay {
-    types_hash: Vec<u8>,
-    verifier: Verifier,
+    pub types_hash: Vec<u8>,
+    pub verifier: Verifier,
 }
 
 impl TypesDisplay {
@@ -255,8 +254,8 @@ impl TypesDisplay {
 /// used for `SufficientCrypto` generation.  
 #[derive(Decode, Encode, PartialEq, Clone)]
 pub struct TypesExport {
-    types_hash: Vec<u8>,
-    signed_by: VerifierValue,
+    pub types_hash: Vec<u8>,
+    pub signed_by: VerifierValue,
 }
 
 impl TypesExport {
@@ -292,11 +291,11 @@ impl TypesExport {
 /// - genesis hash of the network within which the address is  
 #[derive(Decode, Encode, PartialEq, Clone)]
 pub struct IdentityHistory {
-    seed_name: String,
-    encryption: Encryption,
-    public_key: Vec<u8>,
-    path: String,
-    network_genesis_hash: Vec<u8>,
+    pub seed_name: String,
+    pub encryption: Encryption,
+    pub public_key: Vec<u8>,
+    pub path: String,
+    pub network_genesis_hash: Vec<u8>,
 }
 
 impl IdentityHistory {
@@ -329,17 +328,17 @@ impl IdentityHistory {
 #[derive(Decode, Encode, PartialEq, Clone)]
 pub struct SignDisplay {
     /// raw `Vec<u8>` transaction that user either tried to sign or signed  
-    transaction: Vec<u8>,
+    pub transaction: Vec<u8>,
 
     /// name for the network in which transaction is generated,
     /// as it is recorded in the network specs and network metadata  
-    network_name: String,
+    pub network_name: String,
 
     /// address that has generated and signed the transaction  
-    signed_by: VerifierValue,
+    pub signed_by: VerifierValue,
 
     /// user entered comment for transaction
-    user_comment: String,
+    pub user_comment: String,
 }
 
 impl SignDisplay {
@@ -361,9 +360,15 @@ impl SignDisplay {
     /// Get raw transaction, network name, and [`Encryption`] from [`SignDisplay`]  
     pub fn transaction_network_encryption(&self) -> (Vec<u8>, String, Encryption) {
         let encryption = match &self.signed_by {
-            VerifierValue::Standard(MultiSigner::Ed25519(_)) => Encryption::Ed25519,
-            VerifierValue::Standard(MultiSigner::Sr25519(_)) => Encryption::Sr25519,
-            VerifierValue::Standard(MultiSigner::Ecdsa(_)) => Encryption::Ecdsa,
+            VerifierValue::Standard {
+                multi_signer: MultiSigner::Ed25519 { .. },
+            } => Encryption::Ed25519,
+            VerifierValue::Standard {
+                multi_signer: MultiSigner::Sr25519 { .. },
+            } => Encryption::Sr25519,
+            VerifierValue::Standard {
+                multi_signer: MultiSigner::Ecdsa { .. },
+            } => Encryption::Ecdsa,
         };
         (
             self.transaction.to_vec(),
@@ -420,17 +425,17 @@ impl SignDisplay {
 #[derive(Decode, Encode, PartialEq, Clone)]
 pub struct SignMessageDisplay {
     /// decoded message
-    message: String,
+    pub message: String,
 
     /// name for the network in which message transaction is generated,
     /// as it is recorded in the network specs and network metadata  
-    network_name: String,
+    pub network_name: String,
 
     /// address that has generated and signed the message  
-    signed_by: VerifierValue,
+    pub signed_by: VerifierValue,
 
     /// user entered comment for message
-    user_comment: String,
+    pub user_comment: String,
 }
 
 impl SignMessageDisplay {
@@ -474,60 +479,78 @@ impl SignMessageDisplay {
 #[derive(Decode, Encode, Clone)]
 pub enum Event {
     /// Network metadata was added
-    MetadataAdded(MetaValuesDisplay),
+    MetadataAdded {
+        meta_values_display: MetaValuesDisplay,
+    },
 
     /// Network metadata was removed
-    MetadataRemoved(MetaValuesDisplay),
+    MetadataRemoved {
+        meta_values_display: MetaValuesDisplay,
+    },
 
     /// User has generated [`SufficientCrypto`](crate::crypto::SufficientCrypto)
     /// with one of Signer addresses for `load_metadata` update
-    MetadataSigned(MetaValuesExport),
+    MetadataSigned {
+        meta_values_export: MetaValuesExport,
+    },
 
     /// Network specs were added
-    NetworkSpecsAdded(NetworkSpecsDisplay),
+    NetworkSpecsAdded {
+        network_specs_display: NetworkSpecsDisplay,
+    },
 
     /// Network specs were removed
-    NetworkSpecsRemoved(NetworkSpecsDisplay),
+    NetworkSpecsRemoved {
+        network_specs_display: NetworkSpecsDisplay,
+    },
 
     /// User has generated [`SufficientCrypto`](crate::crypto::SufficientCrypto)
     /// with one of Signer addresses for `add_specs` update
-    NetworkSpecsSigned(NetworkSpecsExport),
+    NetworkSpecsSigned {
+        network_specs_export: NetworkSpecsExport,
+    },
 
     /// Network verifier with [`ValidCurrentVerifier`] was set for network
-    NetworkVerifierSet(NetworkVerifierDisplay),
+    NetworkVerifierSet {
+        network_verifier_display: NetworkVerifierDisplay,
+    },
 
     /// General verifier was set up
-    GeneralVerifierSet(Verifier),
+    GeneralVerifierSet { verifier: Verifier },
 
     /// Types information was added
-    TypesAdded(TypesDisplay),
+    TypesAdded { types_display: TypesDisplay },
 
     /// Types information was removed
-    TypesRemoved(TypesDisplay),
+    TypesRemoved { types_display: TypesDisplay },
 
     /// User has generated [`SufficientCrypto`](crate::crypto::SufficientCrypto)
     /// with one of Signer addresses for `load_types` update
-    TypesSigned(TypesExport),
+    TypesSigned { types_export: TypesExport },
 
     /// User has generated signature for a transaction
-    TransactionSigned(SignDisplay),
+    TransactionSigned { sign_display: SignDisplay },
 
     /// User tried to generate signature for a transaction, but failed to enter
     /// a valid password
-    TransactionSignError(SignDisplay),
+    TransactionSignError { sign_display: SignDisplay },
 
     /// User has generated signature for a message
-    MessageSigned(SignMessageDisplay),
+    MessageSigned {
+        sign_message_display: SignMessageDisplay,
+    },
 
     /// User tried to generate signature for a message, but failed to enter
     /// a valid password
-    MessageSignError(SignMessageDisplay),
+    MessageSignError {
+        sign_message_display: SignMessageDisplay,
+    },
 
     /// User generated a new address
-    IdentityAdded(IdentityHistory),
+    IdentityAdded { identity_history: IdentityHistory },
 
     /// User removed an address
-    IdentityRemoved(IdentityHistory),
+    IdentityRemoved { identity_history: IdentityHistory },
 
     /// All identities were wiped
     IdentitiesWiped,
@@ -540,23 +563,23 @@ pub enum Event {
     ResetDangerRecord,
 
     /// New seed was created (stored value here is the seed name)
-    SeedCreated(String),
+    SeedCreated { seed_created: String },
 
     /// User opened seed backup, and seed phrase for shortly shown as a plain
     /// text on screen (stored value here is the seed name)
-    SeedNameWasShown(String), // for individual seed_name
+    SeedNameWasShown { seed_name_was_shown: String }, // for individual seed_name
 
     /// A warning was produces and displayed to user
-    Warning(String),
+    Warning { warning: String },
 
     /// User has entered wrong password
     WrongPassword,
 
     /// User has manually added entry to history log
-    UserEntry(String),
+    UserEntry { user_entry: String },
 
     /// System-generated entry into history log
-    SystemEntry(String),
+    SystemEntry { system_entry: String },
 
     /// History was cleared
     HistoryCleared,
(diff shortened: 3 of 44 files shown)


montekki:
A very basic MVP for `android` is in, the majority of the new changes are new types that mirror the existing ones by either making struct fields public or unnamed enum fields into named ones. Things to do from here:

- [ ] Bring the bridge to ios
- [ ] Unify the majority of said mirroring types
- [ ] Turn the big and fat `json` string returned from navigation and other places a proper type

rust/signer/src/lib.rs
```diff
@@ -14,124 +14,116 @@
...
-        definitions::history::print_all_events()
+fn history_get_warnings(dbname: &str) -> anyhow::Result<bool, anyhow::Error> {
+    db_handling::helpers::get_danger_status(dbname).map_err(|e| e.anyhow())
+}
+
+fn history_acknowledge_warnings(dbname: &str) -> anyhow::Result<(), anyhow::Error> {
+    db_handling::manage_history::reset_danger_status_to_safe(dbname).map_err(|e| e.anyhow())
+}
+
+fn history_entry_system(event: Event, dbname: &str) -> anyhow::Result<(), anyhow::Error> {
```
Slesarew, line 106: This is supposed to allow UI to write generic strings into history. Should accept string and convert it to "system entry" event inside. The only usage now is logging "logging errors", not sure if it is useful at all, but let's just keep for now.

rust/definitions/src/navigation.rs
```diff
@@ -0,0 +1,659 @@
...
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct TransactionNetworkInfo {
+    pub network_title: String,
+    pub network_logo: String,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct TransactionAuthor {
```
Slesarew, line 190: should have `has_password` field and be generally convertable to Address type (so they both can be visualized with the same UI component without extra effort)

rust/db_handling/Cargo.toml
```diff
@@ -22,6 +22,10 @@ sp-core = {git = "https://github.com/paritytech/substrate", default-features = f
 sp-runtime = {git = "https://github.com/paritytech/substrate", default-features = false, optional = true}
 tiny-bip39 = {version = "0.8.0", default-features = false, optional = true}
 zeroize = {version = "1.5.3", optional = true}
+log = { version = "0.4" }
```
Slesarew, line 25: ios build complains that this is unused

rust/definitions/src/navigation.rs
```diff
@@ -0,0 +1,659 @@
...
+#[derive(Clone, Debug, PartialEq)]
+pub struct MMetadataRecord {
+    pub specs_version: String,
+    pub meta_hash: String,
+    pub meta_id_pic: String,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct MNetworkDetails {
+    pub base58prefix: u16,
```
Slesarew, line 334: Even this should be String on settings screen. None of this is really processed or sent further in UI.

rust/definitions/src/navigation.rs
```diff
@@ -0,0 +1,659 @@
...
+    PasswordConfirm { f: MPasswordConfirm },
+    SignatureReady { f: MSignatureReady },
+    EnterPassword { f: MEnterPassword },
+    LogRight { f: MLogRight },
+    ManageNetworks { f: MMManageNetworks },
+    TypesInfo { f: MTypesInfo },
+    Text { f: String },
+    NewSeedMenu,
+    NetworkDetailsMenu,
+    ManageMetadata,
```
Slesarew, line 530: MManageMetadata type was here

rust/definitions/src/navigation.rs
montekki: The reason is how far do the type guarantees go and I would always rather have the guarantees go further (and in this case over the FFI bridge) rather than not. Any UI framework is expected to be perfectly capable of turning a `u8` or a `u32` into a text or do whatever the UI wants to do about showing the actual data to the user. There may be many examples where supplying the actual types over the bridge and actually showing them to the user in UI will definitely be benifitial.

* a hash. You may be willing to display `H256` the way `Display` implementation for it does, you may want to display it as the `LowerHex` implementation does or even display the first part of the hash depending on the situation in the UI.
* a timestamp. Take a look at how the work is done now with timetamps, how it is displayed in the UI and so on. Right now it is sent as a string which is painful because it actually may be displayed in different formats and values due to being in different timestamps etc. R.n. you have to parse a timestamp string and turn it into a proper java type.
* an image. this is obvious.

I will always say that the guarantee is always better then no guarantee and this is the hill i am willing to die on.

Slesarew: I certainly agree with you on image type, of course it should not be passed as string. I agree on timestamp in some cases - if we need any functionality like localization or timezone conversion (which we don't need - at least it is not planned yet). I completely disagree about types only purpose of which is being fed into `Text()` - as having rust control how the final conversion is done is just way more reliable and maintainable than letting UI framework do that; furthermore, we can guarantee that the conversion is then done identically in all platforms and changes propagate to them naturally. If you look at the app as it is now, you can find divergence in some elements already, as our 2 frameworks have barely noticeable disagreements on how to interpret some things here and there; and in addition to that, I have forgotten to correct them synchronously.

One solution to keep some type safety would be to implement conversion to `String` (or, if we are going to that trouble, maybe markdowned AttributedString or something like that to make things prettier?) in rust so that both frameworks can call it similarly; however I doubt it will have good impact on idiomacy. We can try that and then cut the intermediate operation out if it is the only thing done to the value indeed.

rust/navigator/src/lib.rs
```diff
@@ -35,23 +35,23 @@ lazy_static! {
...
-            (*state).generate_json(&details)
-        }
+        Ok(mut state) => state.perform(action, details_str, secret_seed_phrase),
         Err(TryLockError::Poisoned(_)) => {
             //TODO: maybe more grace here?
             //Maybe just silently restart navstate? But is it safe?
             panic!("Concurrency error! Restart the app.");
         }
-        Err(TryLockError::WouldBlock) => "".to_string(),
+        Err(TryLockError::WouldBlock) => Err("WouldBlock".to_string()),
```
Slesarew, line 54: Just don't output this error and make this back into unfallible. This is hardware debouncing, or, in other words, solution for concurrency involving a human. This is not an error that should be shown or logged anywhere. This is also silently rejected on native frontend side.

android/app/src/main/java/io/parity/signer/components/HistoryCardExtended.kt
```diff
@@ -37,114 +40,120 @@ fun HistoryCardExtended(card: JSONObject) {
 				danger = true
 			)
 		}
-		"general_verifier_added" -> {
+		is Event.GeneralVerifierSet -> {
+			val hex = eventVal.verifier.v.let {
+				when (it) {
+					is VerifierValue.Standard -> {
```
Slesarew, line 46: This is not right. General verifier could only be *set* to `Standard`, this should not be allowed to be optional, even on rust side.

Slesarew: My wrong, same event sets it to none on wipe-general-verifier reset

rust/definitions/src/navigation.rs
```diff
@@ -0,0 +1,688 @@
...
+    pub events: Vec<Event>,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct MLog {
+    pub log: Vec<History>,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct MEventMaybeDecoded {
```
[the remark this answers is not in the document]
montekki: I would suggest doing this refactor in a separate further PR, anyway rn they differ only here, in the 2 Transaction variants.

rust/definitions/src/navigation.rs
```diff
@@ -0,0 +1,665 @@
...
+    pub order: u8,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct MManageNetworks {
+    pub networks: Vec<MMNetwork>,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct MSCContent {
```
Slesarew, line 411: This was supposed to tell user what is being signed. Something like `Types` or `Network specs for westend` or `matedata 9090 for westend`. We could do an enum with empty variant for "types", "network" with its name and "metadata" with name and version.

rust/definitions/src/navigation.rs
```diff
@@ -0,0 +1,665 @@
...
+    pub ttype: String,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct MSufficientCryptoReady {
+    pub author_info: Address,
+    pub sufficient: Vec<u8>,
```
Slesarew, line 418: This comes empty now

rust/definitions/src/navigation.rs
```diff
@@ -0,0 +1,661 @@
...
+}
+
+#[derive(PartialEq, Debug, Clone)]
+pub enum ScreenNameType {
+    H1,
+    H4,
+}
+
+#[derive(PartialEq, Debug, Clone)]
+pub enum ShieldAlert {
```
Slesarew, line 79: Explanation on this alert.

Rust side knows only whether the shield alarm was triggered or reset. Thus, values here should be only `past` or `none` (or `safe` or something like that). Source of truth for this is database, alert is set to `past` when actual state moves to `active` and set to `none` when acknowledge button is pressed in shield modal from state `past`

The frontend controls whether `active` state is triggered in shortest possible path. `active` state triggers on network detection and clears on network loss. Only triggering to 'active' sends signal to rust side.

Thus, logic in UI is something like:

```
if (alert) {
  ShowDisconnectNetwork()
} else {
  if (past) {
    ShowAlertWithAcknowledgeButton()
  } else {
    ShowAllIsSafe()
  }
}

File: rust/definitions/src/navigation.rs [hunk 12]
```diff
@@ -0,0 +1,661 @@
...
+    Active,
+    Past,
+}
+
+#[derive(PartialEq, Debug, Clone)]
+pub enum AlertData {
+    Shield { f: Option<ShieldAlert> },
+    ErrorData { f: String },
+    Confirm,
```
Slesarew, line 88: Here the idea was to control all confirmation message from single place. Alert data should have following structure then:

```
struct Contirm {
  line1: Option<String>
  line2: Option<String>
  yes_message: Option<String>
  no_message: Option<String>
}
```
Thus UI will default to something like 

>CONFIRM
>Are you sure?
>yes/no

AFAIR this is not really implemented on Rust side yet and many of these alerts are still called by UI. We should clearly change soon that because bugs in those are potential security-related bugs. Not sure if we should already start preparing for this, please create an issue so that we do not forget if postponing.

rust/definitions/src/navigation.rs
montekki: what is `if(alert)` what exactly are you proposing to match on then?

montekki: how exactly in types are you proposing to move it into `active` state when you are requesting to remove the `active` variant from the type?

Slesarew: > what is `if(alert)` what exactly are you proposing to match on then?

corrected



> how exactly in types are you proposing to move it into `active` state when you are requesting to remove the `active` variant from the type?

It will only exist in UI, as it does in released version. It pretty much is just a check for network availability that is completely on native framework side.

rust/definitions/src/navigation.rs
```diff
@@ -0,0 +1,661 @@
...
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct MManageNetworks {
+    pub networks: Vec<MMNetwork>,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum MSCContent {
+    LoadTypes { types: String, pic: Vec<u8> },
```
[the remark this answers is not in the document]
Slesarew: That is correct; all I'm saying is that I'm not using these and I don't see how these fields could be useful in UI. We could leave this alone for now until designers make this modal more sane.

rust/navigator/src/navstate.rs
```diff
@@ -610,954 +601,1299 @@ impl State {
...
-                Alert::Error => format!("\"error\":\"{}\"", errorline),
-                Alert::ErrorDisplay => format!("\"error\":\"{}\"", errorline),
-                Alert::Empty => "".to_string(),
-                Alert::Shield => "\"shield_state\":\"unknown\"".to_string(),
+            let alert_data = match new_navstate.alert {
+                Alert::Error | Alert::ErrorDisplay => Some(AlertData::ErrorData {
+                    f: format!("{{\"error\":\"{}\"}}", errorline),
+                }),
+                Alert::Empty => None,
+                Alert::Shield => Some(AlertData::Shield { f: None }),
```
Slesarew, line 1877: ok I see the issue now; this just was not implemented in backend yet.

What should be here:
```
match db_handling::helpers::get_danger_status(dbname) {
  Ok(a) => AlertData::Shield(a)
  Err(e) => AlertData::Error(...)
}
```

For now returning constant will work since there is native implementation that calls `get_danger_status` but it would be nice to get rid of that and do it properly. This will require dropping more alerts from native to backend and maybe this will have to wait till next round.

montekki: Done

rust/navigator/src/navstate.rs
```diff
@@ -610,954 +602,1307 @@ impl State {
...
             //Prepare alert details
             //Important! No errors could be handled in this block!
-            let alert_details = match new_navstate.alert {
-                Alert::Error => format!("\"error\":\"{}\"", errorline),
-                Alert::ErrorDisplay => format!("\"error\":\"{}\"", errorline),
-                Alert::Empty => "".to_string(),
-                Alert::Shield => "\"shield_state\":\"unknown\"".to_string(),
+            let alert_data = match new_navstate.alert {
+                Alert::Error | Alert::ErrorDisplay => Some(AlertData::ErrorData {
+                    f: format!("{{\"error\":\"{}\"}}", errorline),
```
Slesarew, line 1875: Remove jsoning here; just output string

rust/navigator/src/navstate.rs
```diff
@@ -610,954 +602,1307 @@ impl State {
...
+                }),
+                Alert::Empty => None,
+                Alert::Shield => match get_danger_status(dbname) {
```
Slesarew, line 1878: for some reason this always sends `.Past` state in swift, even when calling `get_danger_status` directly sends `false`

.github/workflows/android-app.yml
```diff
@@ -35,6 +35,13 @@ jobs:
           toolchain:    stable
           override:     true
 
+      - name:                 Install uniffi_bindgen
+        uses:                 actions-rs/install@v0.1
+        with:
+          crate:              uniffi_bindgen
+          version:            0.17.0 # Must be in sync with version in Cargo.toml
```
Slesarew, line 42: Another important thing about this PR - we should update "how to build Signer" page of manual - what prerequisites are, what should be pre-installed, and that uniffi_bindgen should match Cargo.toml.

rust/navigator/src/navstate.rs
```diff
@@ -610,954 +602,1305 @@ impl State {
...
+                                network_title: i.title,
+                                network_logo: i.logo,
+                            }),
                         },
-                        transaction_parsing::Action::Stub(content, _, _) => format!("\"content\":{{{}}},\"type\":\"stub\"", content),
-                        transaction_parsing::Action::Read(content) => format!("\"content\":{{{}}},\"type\":\"read\"", content),
                     }
-                },
+                }
                 Screen::SeedSelector | Screen::SelectSeedForBackup => {
```
Slesarew, line 1561: These should be two disctinct screens, as they have different actions. They will probably also have different UI later.

rust/navigator/src/navstate.rs
```diff
@@ -610,954 +602,1305 @@ impl State {
...
-                    }
-                },
+                    let seed_name_cards =
+                        db_handling::interface_signer::get_all_seed_names_with_identicons(
+                            dbname,
+                            &self.seed_names,
+                        )
+                        .unwrap();
+                    let f = MSeeds { seed_name_cards };
+                    ScreenData::SeedSelector { f }
```
Slesarew, line 1569: Thus if screen is SelectSeedForBackup, screen data should be SelectSeedForBackup.

Slesarew: This used to work when we had separate `Screen` variable; not anymore

Slesarew:
If you change all `[u8; 32]` to `H256`, you should also change it in `definitions/src/keyring.rs` and further on. We should double-check that it does not change update QR codes format though (it should not, but we should check).

The question is - are you absolutely sure there will not be other `genesis_hash` formats? Technically it seems possible.

montekki:
> The question is - are you absolutely sure there will not be other `genesis_hash` formats? Technically it seems possible.

No but what I am absolutely sure about is that we are unlikely to meet the `[u8; 12345]` or `[u8; 3]` genesis hash types. Also all existing network and the substrate template use `H256` so for now this typelevel restriction looks sane.

montekki:
Using anything but `H256` [does not look possible atm](https://docs.rs/frame-system/latest/frame_system/pallet/trait.Config.html#associatedtype.Hash)

rust/db_handling/src/interface_signer.rs
```diff
@@ -193,40 +190,43 @@ pub fn print_identities_for_seed_name_and_network(
...
+                multiselect,
+            },
+        )
+        .collect();
 
-    Ok(format!(
-        "\"root\":{{{}}},\"set\":{},\"network\":{{\"title\":\"{}\",\"logo\":\"{}\"}}",
-        root_print, other_print, network_specs.title, network_specs.logo
-    ))
+    Ok((root, set, network_specs.title, network_specs.logo))
```
Slesarew, line 229: I suppose we leave these tuples for next refactor.

rust/db_handling/src/interface_signer.rs
```diff
@@ -759,27 +842,14 @@ impl SeedDraft {
...
-                out.push(',')
-            }
-            out.push_str(&format!("{{\"order\":{},\"content\":\"", i));
-            out.push_str(x.word());
-            out.push_str("\"}");
-        }
-        out.push(']');
-        out
+    pub fn draft(&self) -> Vec<String> {
+        self.saved.iter().map(|w| w.word().to_string()).collect()
```
Slesarew, line 846: Does this ever reallocate? This should be zeroizeable; `pust_str` was here previously to be certain that secrets do not go into macro. Just double-check that this is clean and preallocate vectors and strings if needed.

Ideally we should also have an `unsafe` direct memory access test for this at some point, even if it is inevitably flaky.

android/app/src/main/java/io/parity/signer/screens/SettingsScreen.kt
```diff
@@ -25,27 +24,34 @@ import io.parity.signer.uniffi.MSettings
...
  */
 @Composable
-fun SettingsScreen(settings: MSettings, signerDataModel: SignerDataModel) {
+fun SettingsScreen(
+	settings: MSettings,
+	button1: (Action) -> Unit,
+	isStrongBoxProtected: () -> Boolean,
+	getAppVersion: () -> String,
+	wipeToFactory: () -> Unit,
+	alertState: State<AlertData?>
```
Slesarew, line 33: This should be moved to backend later

goldsteinsveta:
Android bugs:

- [ ] Key display name in Key an Tx screens – Seed display name displayed incorrectly (encoded?). Should be the same as seed display name
- [ ] Derived Key screen – derivation path should be on the same line as seed display name
- [ ] Recover Seed – input field does not work
- [ ] Backup – always claims network is connected. Can't backup seed anymore

rust/definitions/src/keyring.rs
```diff
@@ -30,6 +30,7 @@
 //!
 use parity_scale_codec::{Decode, Encode};
 use sled::IVec;
+use sp_core::H256;
```
varovainen, line 33: Let's replace VerifierKey inner content with H256 as well. Because VerifierKey is just the genesis hash.

rust/definitions/src/history.rs
```diff
@@ -155,24 +146,14 @@ impl NetworkSpecsExport {
...
-    }
 }
 
 /// Event content for setting network verifier
 #[derive(Debug, Decode, Encode, PartialEq, Clone)]
 pub struct NetworkVerifierDisplay {
-    genesis_hash: Vec<u8>,
-    valid_current_verifier: ValidCurrentVerifier,
-    general_verifier: Verifier,
+    pub genesis_hash: Vec<u8>,
```
varovainen, line 154: This is also network genesis hash, so should be H256.

rust/definitions/src/history.rs
```diff
@@ -256,38 +229,40 @@ impl TypesExport {
...
+    /// The name of the seed.
+    pub seed_name: String,
+    /// [`Encryption`] scheme of the seed.
+    pub encryption: Encryption,
+    /// Public key.
+    pub public_key: Vec<u8>,
+    /// - path with soft (`/`) and hard (`//`) derivatinos only, **without** password  
+    pub path: String,
+    /// - genesis hash of the network within which the address is  
+    pub network_genesis_hash: Vec<u8>,
```
varovainen, line 246: And here H256 too. And further on.
Also, not sure if we should do it, but meta_hash and types_hash are 32 bytes as defined, so there are no other length options for them either.

rust/db_handling/src/tests.rs
```diff
@@ -286,137 +519,219 @@ fn derive_prep_alice_collided_with_password() {
...
+    .unwrap();
+    let expected_key = MDeriveKey {
+        seed_name: "Alice".to_string(),
+        network_title: "Westend".to_string(),
+        network_logo: "westend".to_string(),
+        network_specs_key: "01e143f23803ac50e8f6f8e62695d1ce9e4e1d68aa36c1cd2cfd15340213f3423e"
+            .to_string(),
+        suggested_derivation: "//secret///abracadabra".to_string(),
+        keyboard: false,
+        derivation_check: None, //TODO "collision":{"base58":"5EkMjdgyuHqnWA9oWXUoFRaMwMUgMJ1ik9KtMpPNuTuZTi2t","path":"//secret","has_pwd":true,"identicon":"<alice_sr25519_//secret///abracadabra>","seed_name":"Alice"}
```
varovainen, line 539: This should be done before merging

rust/db_handling/src/tests.rs
```diff
@@ -286,137 +519,219 @@ fn derive_prep_alice_collided_with_password() {
...
-        .unwrap()
-        .replace(&empty_png(), r#"<empty>"#)
-        .replace(&types_known(), r#"<types_known>"#);
-    let expected_element = r#"{"event":"types_removed","payload":{"types_hash":"d091a5a24a97e18dfe298b167d8fd5a2add10098c8792cba21c39029a9ee0aeb","types_id_pic":"<types_known>","verifier":{"public_key":"","identicon":"<empty>","encryption":"none"}}}"#;
-    assert!(
-        history_printed.contains(expected_element),
-        "\nReceived history: \n{}",
-        history_printed
-    );
+    let types = show_types_status(dbname).unwrap();
```
varovainen, line 647: Why not new `MTypesInfo`? Doesn't really matter, but looks somewhat confusing.

rust/db_handling/src/tests.rs
```diff
@@ -678,8 +992,67 @@ fn test_generate_default_addresses_for_alice() {
...
+                        .as_ref(),
+                )
+                .unwrap(),
+            ),
+            AddressDetails {
+                seed_name: "Alice".to_string(),
+                path: "".to_string(),
+                has_pwd: false,
+                network_id: vec![
+                    NetworkSpecsKey::from_hex(&hex::encode(&[
```
varovainen, line 1011: Let's change the numbers to hex, because it is just kusama, polkadot, and westend genesis hashes.

rust/navigator/src/tests.rs
```diff
@@ -968,623 +3282,1424 @@ fn flow_test_1() {
...
+                        base58: "5HGiBcFgEBMgT6GEuo9SA98sBnGgwHtPKDXiUukT6aqCrKEx".to_string(),
+                        identicon: alice_sr_westend_0().to_vec(),
+                        has_pwd: false,
+                        path: "//westend//0".to_string(),
+                        swiped: false,
+                        multiselect: false,
+                    },
+                ],
+                root: MSeedKeyCard {
+                    seed_name: String::new(), // TODO: "Alice".to_string(),
```
varovainen, line 3794: This should be fixed before merge.

rust/navigator/src/tests.rs
```diff
@@ -968,623 +3282,1424 @@ fn flow_test_1() {
...
-        "LongTap",
+        Action::LongTap,
         "01e655361d12f3ccca5f128187cf3f5eea052be722746e392c8b498d0d18723470",
         "",
+    )
+    .unwrap();
+    // remove keys in multiselect mode
+    let action = do_action(Action::RemoveKey, "", "").unwrap().unwrap();
+
+    if let ScreenData::Keys { ref mut f } = expected_action.screen_data {
```
varovainen, line 3957: Here, since `.remove()` is non-atomic, the number of the removed element looks somewhat confusing. Maybe at this point it already makes sense to redefine the `expected_action` variable.

rust/definitions/src/navigation.rs
```diff
@@ -0,0 +1,659 @@
...
+
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct MNetworkCard {
+    pub title: String,
+    pub logo: String,
+}
+
+// TODO: This has to have a custom default.
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct MSeedKeyCard {
```
varovainen, line 164: `Address` should not have `multiselect` key. `MSeedKeyCard` should probably have fields `address` and `mode`.
enum Mode {
    Plain,
    Swiped(_),
    Multiselect(_),
}
`swiped` and `multiselect` exclude each other and if we go for proper types, those bools that were for json simplicity only are no longer needed. This can wait for later, as now the whole thing works properly.

rust/navigator/src/tests.rs
```diff
@@ -1763,360 +5101,1376 @@ fn flow_test_1() {
...
+            real_json
+        );
 
-    // What if the database is not initiated properly? This should have been a separate test, but mutex.
-    populate_cold_nav_test(dbname).unwrap(); // no init after population
-    let real_json = do_action("NavbarSettings", "", "");
-    let expected_json = r#"{"screen":"Settings","screenLabel":"","back":false,"footer":true,"footerButton":"Settings","rightButton":"None","screenNameType":"h4","modal":"Empty","alert":"Empty","screenData":{"error":"Could not find general verifier."},"modalData":{},"alertData":{}}"#;
-    assert!(real_json == expected_json, "Switched to Settings from Log with non-initiated database. Expected Settings screen with error on screen, and no alerts (we should still allow to reset Signer), got:\n{}", real_json);
+        // What if the database is not initiated properly? This should have been a separate test, but mutex.
+        populate_cold_nav_test(dbname).unwrap(); // no init after population
```
varovainen, line 6469: This piece should be back. Having uninitiated database was quite a common problem (for users who did not reset Signer before updating as instructed, but still), we need this test to make sure they have a way out.

rust/transaction_parsing/src/tests.rs
```diff
@@ -282,85 +696,475 @@ fn parse_transaction_1() {
...
+        assert_eq!(author_info, author_info_known);
         assert!(!has_pwd, "Expected no password");
+        assert_eq!(network_info, network_info_known);
     } else {
         panic!("Wrong action {:?}", output)
     }
     fs::remove_dir_all(dbname).unwrap();
 }
 
+/* TODO: does this actually test anything meaningful?
```
varovainen, line 1163: Not anymore, I suppose. This was mostly to test for punctuation errors in json. Off it goes?
We still need some test to see that all cards are displayed as expected on screen though.

rust/transaction_signing/src/tests.rs
```diff
@@ -103,11 +126,43 @@ fn print_db_content(dbname: &str) -> String {
         let verifier_key = VerifierKey::from_ivec(&verifier_key_vec);
```
varovainen, line 126: Verifier details should be eventually simplified here to `general` and `custom(..)`. Identicons are now tested in `navigator`, json-like syntax is not used anywhere else.

rust/signer/src/ffi_types.rs
```diff
@@ -0,0 +1,110 @@
...
+pub type Sr25519Public = sp_core::sr25519::Public;
+pub type EcdsaPublic = sp_core::ecdsa::Public;
+pub type MultiSigner = sp_runtime::MultiSigner;
+pub type H256 = sp_core::H256;
+
+impl UniffiCustomTypeConverter for sp_runtime::MultiSigner {
+    type Builtin = String;
+
+    fn into_custom(val: Self::Builtin) -> uniffi::Result<Self> {
+        Ok(serde_json::from_str(&val)?)
```
Slesarew, line 43: This should be done like in `db_handling::helpers::get_general_verifier` with separate fields for pubkey, identicon and encryption. With default upstream serde implementation, this results in base58 representation for nonsence network; thus history starts with weird things.

Slesarew (changes requested):

rust/definitions/src/navigation.rs
```diff
@@ -0,0 +1,659 @@
...
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct MEnterPassword {
+    pub author_info: Address,
+    pub counter: u32,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct MLogRight {
+    pub checksum: String,
```
Slesarew, line 497: This stores some json; should store only its value as string

rust/signer/src/ffi_types.rs
montekki: I've added a hack that sends `vec[key, encryption]` as strings for now. This should be properly encoded as `MVerifierDetails` I guess but for now I have failed to make it work with uniffi machinery

rust/definitions/src/history.rs
```diff
@@ -155,24 +146,14 @@ impl NetworkSpecsExport {
...
-            "{},\"signed_by\":{}",
-            &self.specs_to_send.show(),
-            export_complex_single(&self.signed_by, |a| a.show_card())
-        )
-    }
 }
 
 /// Event content for setting network verifier
 #[derive(Debug, Decode, Encode, PartialEq, Clone)]
 pub struct NetworkVerifierDisplay {
```
varovainen, line 153: Shows network genesis hash instead of verifier public key in log message.

montekki: i think fixed now

rust/db_handling/src/interface_signer.rs
```diff
@@ -411,26 +430,45 @@ pub fn derive_prep(
...
-    match collision {
-        Some((multisigner, address_details)) => {
-            let address_base58 = print_multisigner_as_base58(&multisigner, Some(network_specs.base58prefix));
-            let hex_identicon = hex::encode(make_identicon_from_multisigner(&multisigner));
-            let collision_display = format!("\"base58\":\"{}\",\"path\":\"{}\",\"has_pwd\":{},\"identicon\":\"{}\",\"seed_name\":\"{}\"", address_base58, address_details.path, address_details.has_pwd, hex_identicon, seed_name);
-            Ok(format!("\"seed_name\":\"{}\",\"network_title\":\"{}\",\"network_logo\":\"{}\",\"network_specs_key\":\"{}\",\"suggested_derivation\":\"{}\",\"collision\":{{{}}}", seed_name, network_specs.title, network_specs.logo, hex::encode(network_specs_key.key()), suggest, collision_display))
-        },
-        None => Ok(format!("\"seed_name\":\"{}\",\"network_title\":\"{}\",\"network_logo\":\"{}\",\"network_specs_key\":\"{}\",\"suggested_derivation\":\"{}\"", seed_name, network_specs.title, network_specs.logo, hex::encode(network_specs_key.key()), suggest)),
-    }
+    let derivation_check = collision.map(|(multisigner, address_details)| {
```
Slesarew, line 436: derivation_check field should not be an option: it should always be present and explicitly tell UI, whether button is pressable and if so what should it do; even when there are no collisions, it should be there. Right now this behavior results in following bug: 

1. create seed without seed key 
2. try to derive key - button is grey (incorrect)
3. type a character and remove it - button is blue (correct)

This should behave exactly as `db_handling::interface_signer::dynamic_path_check` derivation_check constructor

varovainen:
Genesis hash is displayed sometimes in uppercase (`History`), sometimes in lowercase (`Network details` screen). Should be made uniform.

android/app/src/main/java/io/parity/signer/components/HistoryCard.kt
```diff
@@ -135,12 +136,30 @@ fun HistoryCard(card: Event, timestamp: String) {
 			}
 		}
 		is Event.NetworkVerifierSet -> {
+			var line3 =
```
Slesarew, line 139: Also should be fixed in HistoryCardExtended

PR #1001, opened 2022-04-19
build: make uniffi buildable for ios

Slesarew:
Just buggy boilerplate for now. Making sure it links at all. We can keep it nicely isolated in this branch until all is ready.


Slesarew:
reverted shieldalert to native-based logic; there is something elusive happening with shield state.

PR #1006, opened 2022-04-21
fix: rename root keys to seed keys in ui

Slesarew:
We will need to fix this in internal code as well later, after ongoing large refactor #999


goldsteinsveta:
/parity-signer/android/app/src/main/java/io/parity/signer/components/KeyCard.kt has a mention of root keys in comments. should be seed keys

PR #1008, opened 2022-04-21
Adjusted Docker file and sign_android.sh to be build and signe android  signer.apk 

EgorPopelyaev:


docker/Dockerfile
```diff
@@ -1,20 +1,19 @@
...
+LABEL maintainer="Parity Technologies <admin@parity.io>"
 
 ENV ANDROID_SDK_ROOT /opt/android-sdk-linux
 ENV ANDROID_HOME /opt/android-sdk-linux
 ENV NDK_HOME /opt/android-ndk
-ENV ANDROID_NDK_VERSION r23b
 ENV JAVA_HOME /usr/lib/jvm/java-11-openjdk-amd64/
 
 RUN apt-get -y update && apt-get -y install default-jdk wget unzip curl clang
+RUN apt-get -y install python git
```
rcny, line 11: Combine it into the one layer, e.g. `RUN apt-get -y update && apt-get -y install default-jdk wget unzip curl clang python git`

docker/Dockerfile
```diff
@@ -25,20 +24,7 @@ ENV PATH ${PATH}:${ANDROID_SDK_ROOT}/platform-tools:${ANDROID_SDK_ROOT}/cmdline-
...
-    unzip -q android-ndk-${ANDROID_NDK_VERSION}-linux.zip && \
-# move to its final location
-    mv ./android-ndk-${ANDROID_NDK_VERSION} ${NDK_HOME} && \
-# remove temp dir
-    cd ${NDK_HOME} && \
-    rm -rf /opt/android-ndk-tmp
-
-# add to PATH
-ENV PATH ${PATH}:${NDK_HOME}
+RUN echo "y" | sdkmanager --install "ndk;24.0.8215888" --sdk_root=${ANDROID_SDK_ROOT}
```
TriplEight, line 27: It's better to combine all the `RUN` operations into one layer

docker/Dockerfile
```diff
@@ -1,20 +1,19 @@
 # This dockerfile is based on Bitrise but with a lot of extra cruft we don't need removed
 FROM ubuntu:focal
-MAINTAINER Parity Technologies <admin@parity.io>
+LABEL maintainer="Parity Technologies <admin@parity.io>"
```
TriplEight, line 3: A good practice would be write more metadata as we do pretty much everywhere [else](https://github.com/paritytech/scripts/blob/master/dockerfiles/ci-linux/Dockerfile#L8)

TriplEight (changes requested):

docker/Dockerfile
EgorPopelyaev: Here, when I do so, I get a strange behavior while installing jdk. It prompts to chose a region and while building an image stuck at this point. When it separate, it goes without asking for it. Don't know exactly why it behaves like this.

docker/Dockerfile
EgorPopelyaev: 👍 I'll adjust it accordingly

docker/Dockerfile
TriplEight: ` ENV DEBIAN_FRONTEND=noninteractive` solves this

docker/Dockerfile
TriplEight: apparently, as long as you're not publishing it, and I've never found if you do, this is not necessary.

EgorPopelyaev: yep, basically, this image is intended only for internal use, to be able to have a quick set up to build an Android version

PR #1045, opened 2022-05-17
docs: fix adding new network docs

jak-pan:
Fix docs about adding new networks (multiple typos with add_spec instead of add_specs)


Slesarew (approved):
thank you

PR #1051, opened 2022-05-23
fix: reduce qr size, switched to indexed colors

varovainen:
After switching to `png` crate in multiframe QR generation, all QR code files became much larger with same content and frame number. Switched here to 1-bit indexed colors instead of 8-bit grayscale, to get smaller size files. QRs are still done in two colors, main and back, from `constants`. These could be any colors, not only grayscale.


prybalko (approved):
Now it's even more lightweight than it used to be. Nice 👍

PR #1054, opened 2022-05-24
fix: android filter out non qr barcodes

montekki:


Slesarew (approved):
This does not seem to cause backwards compatibility issues. Does not cause noticeable performance boost, unfortunately, but this is certainly right thing to do.

PR #1065, opened 2022-06-02
fix: android: scan screen always on

### LINKED ISSUE #1041: Android - Prevent screen from going off when scanning
Scanning Metadata takes a while, a lot longer than the automatic screen sleep timer (prob 30s to 1min depending on users).
Generally, I expect users to place their phone in front of their screen, and leave it there rather than holding it for several minutes.

Because of this, it would be great if we you had the ability to prevent the phone to turn into sleep mode, and in the same time, keep the QR reading on.

montekki:
Fixes #1041


Slesarew (changes requested):
I think it would be a nice UX move to only enable this if at least 1 frame was captured.

User accidentally left phone in scan mode - go to sleep mode;

User started scanning - keep awake

Slesarew (approved):
The check could be hidden into ScanProgressBar which checks for exactly same condition, but it does not matter now and things will probably change later.

PR #1069, opened 2022-06-03
fix: remove message signing until final form is decided

varovainen:
Processing messages (payload code `0x03`) is temporarily blocked.
It will be back as soon as the exact content of the raw message is discussed, approved, and written in standard.
Related tests are modified or commented.


prybalko:
Superseded by #1332

PR #1070, opened 2022-06-06
fix: backend: verifier key hash type

### LINKED ISSUE #1060: Use H256 for all 32-byte items.
Let's replace VerifierKey inner content with H256 as well. Because VerifierKey is just the genesis hash.

_Originally posted by @varovainen in https://github.com/paritytech/parity-signer/pull/999#discussion_r884669091_

montekki:
Fixes #1060

### DIFF
--- a/rust/db_handling/src/helpers.rs
+++ b/rust/db_handling/src/helpers.rs
@@ -507,7 +507,7 @@ pub fn remove_network(
     let general_verifier = get_general_verifier(database_name)?;
     let network_specs = get_network_specs(database_name, network_specs_key)?;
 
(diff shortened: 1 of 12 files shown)


varovainen:
Just realized something. If the `&[u8]` that we try to get into H256 has incorrect size, `primitive_types` source panics ([here](https://github.com/paritytech/parity-common/blob/master/fixed-hash/src/hash.rs#L188)). We can add size-check right before the transformation.

Slesarew (approved):
Now we should change genesis_hash and probably more into this type.

PR #1097, opened 2022-06-08
ci: rust test all features

Slesarew:
This test is failed by certain modifications in #1092 and I nearly forgot about it; thus this should be automated.

Maybe this should be done in parallel, maybe on both platforms.

### DIFF
--- a/.github/workflows/rust-test-android.yml
+++ b/.github/workflows/rust-test-android.yml
@@ -58,3 +58,6 @@ jobs:
         run:                  |
           cd rust
           cargo nextest run --retries 2 --locked
+          cargo nextest --no-default-features --features signer --retries 2 --locked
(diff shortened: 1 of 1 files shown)


.github/workflows/rust-test-android.yml
```diff
@@ -58,3 +58,6 @@ jobs:
         run:                  |
           cd rust
           cargo nextest run --retries 2 --locked
+          cargo nextest --no-default-features --features signer --retries 2 --locked
+          cargo test --no-default-features --features active --retries 2 --locked
+          cargo test --no-default-features --retries 2 --locked
```
montekki, line 63: btw why both `test` and `nextest` are used?

montekki (approved):

PR #1119, opened 2022-06-13
refactor: integration tests separated, optional path hot db

varovainen:
- separated `db_handling` integration tests
- added optional path for `default_hot` (will need later, for tests in `generate_message`)

Note: as Rust does not allow optional `dev-dependencies` (apparently, please advise if you know better), with linter `#![deny(unused_crate_dependencies)]` tests keep failing on features other than default `"test"`. Found no way around it better than is currently (after all, when `db_handling` is addressed from elsewhere, `dev-dependencies` are ignored and linter passes ok).


montekki (approved):
I believe that in the long run we should aim for complete removal of the `test` feature as it is sort of non-idiomatic

PR #1120, opened 2022-06-13
fix: settings screen should never fail

Slesarew:
Settings screen should always allow alert to be discarded. I've added inline comments to important places, because even though those are tested areas, we've fixed this before and broke again, and even `expect` messages did not save us.

### DIFF
--- a/rust/navigator/src/navstate.rs
+++ b/rust/navigator/src/navstate.rs
@@ -27,6 +27,7 @@ use definitions::{
         ErrorSigner, ExtraAddressKeySourceSigner, InputSigner, InterfaceSigner, Signer,
     },
     keyring::{AddressKey, NetworkSpecsKey},
+    network_specs::Verifier,
     users::AddressDetails,
 };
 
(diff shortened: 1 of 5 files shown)


rust/navigator/src/navstate.rs
```diff
@@ -1885,7 +1900,9 @@ impl State {
                 Ok(sd) => sd,
                 Err(e) => {
                     errorline.push_str(&<Signer>::show(&e));
-                    new_navstate.alert = Alert::Error;
+                    new_navstate.alert = Alert::ErrorDisplay; //This is special error used only
+                                                              //here; please do not change it to
+                                                              //`Alert::Error` or app may get stuck
```
montekki, line 1905:
```suggestion
                    //This is special error used only
                    //here; please do not change it to
                    //`Alert::Error` or app may get stuck
                    new_navstate.alert = Alert::ErrorDisplay; 
```

montekki (approved):

PR #1124, opened 2022-06-14
fix: paths to ndk were fixed in upstream

montekki:
mozilla/rust-android-gradle#91 was fixed, let's give it a try


Slesarew (approved):
Magic!

PR #1133, opened 2022-06-20
ci: restructure the docs pipeline

TriplEight:
- places generating user-docs after rustdocs are built, so it's possible to check the links between them
- publishes user-docs to https://paritytech.github.io/parity-signer/user-docs/
- publishes rustdocs to i.e. https://paritytech.github.io/parity-signer/rustdocs/constants/index.html (it's better to choose an initial page and place an index.html like [this](https://github.com/paritytech/substrate/blob/master/scripts/ci/gitlab/pipeline/build.yml#L157))
- empties `gh-pages` branch after every deploy (on master) with `force_orphan: true`

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -71,8 +71,7 @@ jobs:
           mv ./docs/book/html/ ./user-docs
           echo "Removing everything but the docs dirs"
           shopt -s extglob
-          rm -vrf !("rustdocs"|"user-docs"|"README.md")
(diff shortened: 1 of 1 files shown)


radupopa2010 (approved):

sergejparity (approved):
LGTM.
Looks like only required statuses checks needs to be updated in the repository.

TriplEight:
yup, right before the merge

PR #1134, opened 2022-06-20
fix: proper hash types

### LINKED ISSUE #1060: Use H256 for all 32-byte items.

montekki:
I think that this about does it

Fixes #1060


rust/transaction_parsing/src/check_signature.rs
```diff
@@ -26,9 +26,10 @@ pub fn pass_crypto(
             // Ed25519 crypto was used by the verifier
             let (pubkey, data) = match data.get(3..35) {
                 Some(a) => {
-                    let into_pubkey: [u8; 32] =
-                        a.try_into().expect("fixed size should fit in array");
-                    (ed25519::Public::from_raw(into_pubkey), &data[35..])
+                    let into_pubkey = H256::from(
```
varovainen, line 29: These are public keys, H256 should not be here, as there is no hashing. Just below is `ecdsa` with public key length 33.

docs/src/development/UOS.md
```diff
@@ -275,7 +275,7 @@ Thus, the transaction structure could also be represented as:
 
 Bold-marked transaction pieces are used in the blob for which the signature is
 produced. If the blob is short, 257 bytes or below, the signature is produced
-for it as is. For blobs longer than 257 bytes, 32 byte hash (`blake2-rfc`) is
+for it as is. For blobs longer than 257 bytes, 32 byte hash (`blake2_256`) is
```
varovainen, line 278: Do you mean changing direct use of `blake2-rfc` crate to Substrate wrapper? Or this is just for more clear wording here? The `transaction_signing` (where the long transaction hashing is done) still has direct `blake2-rfc`.

Slesarew: checked signing westend lorem ipsum remark on-chain by Alice//westend using this branch; works correctly.

prybalko (commented):
Should `pub network_genesis_hash: Vec<u8>,` in `IdentityHistory` also be `H256`?

krodak (approved):

PR #1137, opened 2022-06-20
ci: fix removing the non-docs

TriplEight:


.github/workflows/docs.yml
```diff
@@ -68,14 +68,15 @@ jobs:
         # executed only on master so PRs can use cache
         if:                   ${{ github.ref == 'refs/heads/master' }}
         run:                  |
-          echo "Removing everything but the docs dirs"
-          shopt -s extglob
-          rm -vrf !("./docs/book/html/")
+          echo "Removing everything but the docs dir"
+          mv -v ./docs/book/html/ /tmp/html/
```
Slesarew, line 72: shouldn't we remove whatever might happen to be at /tmp/html/ first just in case?

TriplEight: well, these runners are supposed to be always clean spot instances™, but you're right

Slesarew (approved):

PR #1143, opened 2022-06-22
test: make possible running tests from individual crates

Slesarew:
Make possible to run `cargo test` from `transaction_*` crates.


montekki:
It is not necessary to run it _from_ these crates, you should be able to do `cargo test -p transaction_parsing` but atm it is broken because of the `test` feature:

```
cargo test -p transaction_parsing

...

error[E0432]: unresolved imports `db_handling::cold_default::populate_cold`, `db_handling::cold_default::populate_cold_no_metadata`, `db_handling::cold_default::populate_cold_no_networks`
 --> transaction_parsing/src/tests.rs:6:20
  |
6 |     cold_default::{populate_cold, populate_cold_no_metadata, populate_cold_no_networks},
  |                    ^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^^^^^^^^ no `populate_cold_no_networks` in `cold_default`
  |                    |              |
  |                    |              no `populate_cold_no_metadata` in `cold_default`
  |                    no `populate_cold` in `cold_default`

Slesarew:
The reason behind `test` feature is that some unit tests here work kind of like integration tests - test database is first generated through hot-side code and then tested on `signer` feature. The best solution would probably be to make these tests integration tests instead, however that would imply integrating crates with the same crates however with different features. The other option is to place it into navigation test, but I am strongly against that, because it already is almost 10k lines of code and ~5 minutes of time, moving tests there will just decrease readability.

I propose we just fix these tests like this until we do that.

The other thing to note is that "test" feature includes operations that result in invalid objects being generated to test for corrupted states of system - like generation of invalid metadata, spoiled db, etc. Where should this code be if not under "test" feature? Clearly it should not be available in normal app builds.

montekki (approved):

PR #1147, opened 2022-06-27
fix: ios: restore alerts functionality

Slesarew:
Alerts in iOS got broken on last update (crashing the app on alert call or dismiss); this restores them.

To test: start app and call alerts (push on shield, try to type wrong password on transaction, etc.)


ios/NativeSigner/Screens/MainScreenContainer.swift
```diff
@@ -24,74 +24,74 @@ struct MainScreenContainer: View {
...
-                    )
-                    AlertSelector(
-                        alertData: data.actionResult.alertData
+                    .gesture(
+                        DragGesture().updating($dragOffset, body: { (value, state, transaction) in
+                            if value.startLocation.x < 20 && value.translation.width > 100 {
+                                data.pushButton(action: .goBack)
+                            }
+                        })
                     )
```
prybalko, line 75: Just curious, any reason not to pass `.environmentObject(data)` to `AlertSelector`?

ios/NativeSigner/Screens/MainScreenContainer.swift
```diff
@@ -24,74 +24,74 @@ struct MainScreenContainer: View {
...
-                            increment: { seedName, details in
-                                let seedPhrase = data.getSeed(seedName: seedName)
-                                if seedPhrase != "" {
-                                    data.pushButton(action: .increment, details: "1", seedPhrase: seedPhrase)
-                                }
-                            }
+                            getSeedForBackup: {seedName in return data.getSeed(seedName: seedName, backup: true)},
+                            sign: {seedName, comment in data.sign(seedName: seedName, comment: comment)}
+                        )
+                        AlertSelector(
```
prybalko, line 65: Just curious, any reason not to pass .environmentObject(data) to AlertSelector?

prybalko: > there is no environment object anymore

I'm curious why it's not used. Is it a bad practice?

Slesarew: In general, no.

But in this particular case, we've moved truth source onto particular screen and instead of passing the whole object, we pass only item-specific data and closures. This makes it easier to use screen previews and in general test things separately, since now there is no need to mock the whole data, only screen-specific part of it.

Global variables make sense if it's indeed something small that's used in many screens, but even network detection alert is does not fit to this description.

prybalko (approved):

montekki (approved):

PR #1157, opened 2022-06-29
chore: ios: add swiftlint

Slesarew:
Add popular linter for iOS code.


prybalko (commented):
It might be a good idea to enforce linter checks. Maybe a precommit hook or CI job checking that `swiftlint` is happy.

Slesarew:
> It might be a good idea to enforce linter checks. Maybe a precommit hook or CI job checking that `swiftlint` is happy.

Or fail ios build run on warnings

montekki (approved):

PR #1170, opened 2022-07-17
chore: rewrite error handling

montekki:
An project-wide overarching error type system does not look like sustainable. For instance, this is one of the things that prevents separate crates from being easily reused. On top of that it uses non-idiomatic trait system which makes things even more complicated. Refactoring is also a PITA.

Instead, make every crate have its own error type (as in for example `substrate`), use `thiserror` to avoid boilerplate and do other fixes.


rust/defaults/src/lib.rs
```diff
@@ -96,6 +91,23 @@ pub const DEFAULT_VERIFIER_PUBLIC: [u8; 32] = [
     0x04, 0x85, 0xc7, 0x2b, 0x4e, 0xcf, 0x3c, 0x59, 0x9e, 0xcc, 0xa6, 0x99, 0x8f, 0x39, 0xbd, 0x57,
 ];
 
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
```
prybalko, line 95: I'd propose to move it to its module just to be consistent with other crates.

prybalko (approved):

PR #1192, opened 2022-08-02
chore: ios setup development scheme

krodak:
## Purpose ℹ️ 
- We should have an ability to run application locally and be able to mock connectivity state during development.
- Protocol-oriented approach and dependency injection should be used to enable unit testing.

## Scope 🔭
- replace inline use of `NWPathMonitor` with mockable adapter
- introduce dependency injection for `NWPathMonitor` adapter
- add `RuntimePropertiesProvider` to act as a wrapper around `Environmental Variables` accessible for given `Scheme`
- add `NativeSigner-Dev` scheme 
- setup unit test target by adding bridging header to test target
- add unit test coverage
- remove empty test target
- remove dead code related to Bluetooth connectivity as it was never used

## Screenshots 🖼️ 
XCode now features two separate schemes:
CleanShot 2022-08-02 at 14 06 59@2x


montekki (approved):
Great! merge when green

A little nit: please do not force-push your branches in the future, the workflow across the codebase is generally creating merge commits and adding new commits on top of that which makes re-reviewing "changes since last seen" to the PR easier.

krodak:
@montekki got it, thanks for information, got used to rebase + force push approach a bit too much 😅 🙏🏻

prybalko (approved):

vas3k (approved):

vas3k (approved):
Thanks for not forgetting to update the README :)

PR #1198, opened 2022-08-03
refactor: extract database operations from SignerDataModel and reduce code smell within SignerDataModel

krodak:
## Purpose ℹ️ 
This PR aims to start decoupling process for `SignerDataModel` class that is currently too massive and handles to many responsibilities.
This is achieved through use of composition and protocol-based dependency injection to enable testing for `SignerDataModel` further down the road 

## Scope 🔭
- `DatabaseMediator` that extracts and decouples local database-related operations from within `SignerDataModel`, following SRP rule
- unit test coverage for `DatabaseMediator`
- relevant refactor for main part of `SignerDataModel` to remove hardcoded in-line `String` values, to make code more DRY and follow good practices (smaller functions)
- removing `UITestsTarget` as it's empty now and returns warnings when running unit tests


ios/NativeSignerTests/Core/Database/DatabaseMediatorTests.swift
```diff
@@ -0,0 +1,308 @@
...
+        let result = subject.recreateDatabaseFile()
+
+        // Then
+        XCTAssertTrue(result)
+    }
+}
+
+// MARK: - Mocks
+
+final class FileManagingProtocolMock: FileManagingProtocol {
```
krodak, line 236: For now I'm creating these mocks within my won personal project that has [Sourcery](https://github.com/krzysztofzablocki/Sourcery) setup for mock generation, I'll remove those after Sourcery is integrated within Signer and those mocks can be autogenerated locally during compilation time, so no need to check them out into the repo. 
They will follow exact same syntax that you can see for below mocks 🙏🏻

montekki (approved):

PR #1204, opened 2022-08-04
refactor: remove redundancy from HistoryCardExtended

krodak:
## Purpose ℹ️ 
`HistoryCardExtended` features mostly copy-and-paste code from `HistoryCard`, this PR addresses that. 
Additionally views hierarchy featured unnecessary `HStack` that is removed in this PR

## Scope 🔭
- use `HistoryCard` logic where application within `HistoryCardExtended`
- simplify required datasets by making `timestamp` and `line2` optionals
- remove `HStack`
- add utility method to keep `[UInt8] -> String` conversion DRY

## Screenshots 🎥 
Before / After to make sure nothing is broken (as we still don't have UI tests)

| Before | After | 
|-|-|
|   |   |


montekki:
1. check the Tx case (afair the Extended version mostly exists for them)
2. any pr changing UI behavior has to happen for both platforms at the same time otherwise UI is going to fork endlessly

krodak:
1) @montekki I tried `key created` as it also use extended designs, here it is, all works fine, details are properly presented on `Event details` screens:

|||
|-|-|
| IMG_5848 |  IMG_5849 |



2) There are absolutely no changes in design, `HStack` was there completely unnecessary, just clogging internal views structures for no reason, rest of refactor just aims to reduce copy-paste within code 🙏🏻

montekki (approved):

PR #1208, opened 2022-08-08
chore: remove gitignored files

prybalko:
Files in `ios/NativeSigner/Generated/` are autorenerated, so there is no point to keep them.

This filder is already in .gitignore
https://github.com/paritytech/parity-signer/blob/master/.gitignore#L86


krodak:
Thanks @prybalko, I wasn't sure whether this was intentional or not 😅 🙏🏻 
For new files that are generated, I've added all of them to this folder, so `.gitignore` is properly triggered for them 👍🏻

krodak (approved):
Thanks @prybalko, I wasn't sure whether this was intentional or not 😅 🙏🏻 
For new files that are generated, I've added all of them to this folder, so `.gitignore` is properly triggered for them 👍🏻

montekki (approved):

PR #1213, opened 2022-08-10
ci: iOS unit tests GA and removal of redundant actions

krodak:
## Purpose ℹ️ 
This PR aims at providing GA that would build iOS app and run its unit tests.

In addition, I'm proposing removal of following GA:
- `swiftlint.yml` - we are running `swiftlint lint --strict` as part of iOS build phases, so there is no need for separate action for that, this will be triggered on newly added action during iOS build time

## Issues
I've tried to remove all warnings that were triggered within `action.yml`, but I'm unable to remove one that's visible below:
CleanShot 2022-08-10 at 08 47 12@2x

## Screenshots 📱 
CleanShot 2022-08-10 at 08 44 40@2x


.github/workflows/ios-unit-tests.yml
```diff
@@ -52,27 +56,25 @@ jobs:
         uses:                 actions-rs/install@v0.1
         with:
           crate:              uniffi_bindgen
-          version:            0.19.3 # Must be in sync with version in Cargo.toml
```
montekki, line 55: The version and the comment were here for a reason.

krodak: Fixed, sorry for that; removed trying to silence action warnings 🙏🏻

.github/workflows/ios-unit-tests.yml
```diff
@@ -57,22 +61,21 @@ jobs:
 
       - name:                 Rust Cache
         uses:                 Swatinem/rust-cache@v2.0.0
-        with:
-          working-directory:  rust
```
montekki, line 61: This is going to break the cache on rust builds and as a result within the same PR _all_ rust deps are going to be rebuilt every time a new commit is pushed to a branch.

krodak: I've removed it in my file as I was inspecting GA warnings: 
CleanShot 2022-08-10 at 10 06 20@2x

CleanShot 2022-08-10 at 10 07 20@2x

Maybe different value should be there @montekki ?

.github/workflows/ios-unit-tests.yml
```diff
@@ -1,15 +1,16 @@
...
 on:
   pull_request:
   push:
     branches:
       - master
       - stable
+
 jobs:
-  check:
-    name:                     Check on MacOS
```
prybalko, line 11: `Check on MacOS` name is marked as a required action for this repo. Without it, it's not possible to merge any PRs. The name can be changed in the settings, but I might prefer the old name. I think`Run iOS Unit Tests` doesn't hint that linter checks also run there.

krodak:
Changed this PR so it actually edits `actions.yml` instead of adding new one, hence also not requiring any changes in general GA setup 🙏🏻 I've started with fresh GA while I still wasn't aware what actions.yml is doing exactly, so it turned out not to be best approach 😅 
PR should be much clearer now, thanks for all the feedback

.github/workflows/actions.yml
```diff
@@ -23,12 +24,15 @@ jobs:
           fetch-depth:        50
           submodules:         'recursive'
 
+      - name:                 Setup - Xcode
+        run:                  sudo xcode-select -switch '/Applications/Xcode_13.4.1.app/Contents/Developer' && /usr/bin/xcodebuild -version
```
prybalko, line 28: would we need to edit this line with every Xcode release?

krodak: No, we only need to update this once this XCode version won't be able to properly build app version or one of Homebrew dependencies will require XCode version bump.
So far default for Github Action env is 13.2, but `swiftgen` requires `13.3` hence manual bump to newest version 🙏🏻

montekki (approved):

prybalko (approved):

PR #1214, opened 2022-08-10
ci: update Rust Cache step in all GA actions

krodak:
## Purpose ℹ️ 

To fix Rust Cache step on GAs

Did a test run here, seems to save cache properly: https://github.com/paritytech/parity-signer/actions/runs/2831246280

### DIFF
--- a/.github/workflows/android-app.yml
+++ b/.github/workflows/android-app.yml
@@ -57,7 +57,7 @@ jobs:
       - name:           Rust Cache
         uses:           Swatinem/rust-cache@v2.0.0
         with:
-          working-directory: rust
+          workspaces:   rust -> rust/target
 
       - name:           Gradle Cache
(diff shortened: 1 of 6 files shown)


.github/workflows/android-app.yml
```diff
@@ -125,7 +125,7 @@ jobs:
       - name:           Rust Cache
         uses:           Swatinem/rust-cache@v2.0.0
         with:
-          working-directory: rust
+          workspaces:   rust -> rust/target
```
montekki, line 128: There is another one in this file still left.

montekki (approved):
```
Cache restored successfully
Restored from cache key "v0-rust-check-ef8eb32af1beb44c578cddab0a98848581008893-b64491d63f9c974483d5cf70273bc28e3f2371c3".
```
Seems to work now

PR #1218, opened 2022-08-16
ci: Fastlane and Testflight setup for GA workflow

krodak:
## Purpose ℹ️ 
This PR aims at providing GA that would build iOS app and distribute it through TestFlight

## Scope 🔭
- App Store Connect configured with a newly created certificate
- all required Github Secrets added, including dedicated API Key and base64 encoded certificate
- [Fastlane](https://fastlane.tools) setup for TestFlight distribution
- GA workflow that wraps whole process

## TODO ☑️ 
- [x] Discuss and agree on what should be triggers to run this GA (merge to `/release/*` branch, cron with daily / weekly builds, merging to `master`, "dispatch_workflow" with some user-based limitations)
- [x] Prepare Notion document describing the process
- [x] Discuss whether we need to create `Issue` to mimic the process on Android side: https://github.com/paritytech/parity-signer/issues/1224


.github/workflows/testflight.yml
```diff
@@ -0,0 +1,106 @@
...
+          crate:              uniffi_bindgen
+          version:            0.19.3 # Must be in sync with version in Cargo.toml
+          use-tool-cache:     true
+
+      - name:                 Rust Cache
+        uses:                 Swatinem/rust-cache@v2.0.0
+        with:
+          workspaces:         rust -> rust/target
+
+      - name:                 Install the Apple certificate and provisioning profile
```
krodak, line 72: Simplified version of GitHub provided step run to include certificate in runner's temporary keychain: https://docs.github.com/en/actions/deployment/deploying-xcode-applications/installing-an-apple-certificate-on-macos-runners-for-xcode-development

ios/fastlane/Appfile
```diff
@@ -0,0 +1,4 @@
+app_identifier  "io.parity.NativeSigner"
+apple_id        "krzysztof@parity.io"
```
krodak, line 2: All relevant actions are done through API key, this is only used if there is a need to refresh Provisioning Profiles

ios/fastlane/lanes/lane_tests.rb
```diff
@@ -0,0 +1,15 @@
...
+desc "- 'scheme : <value>' to define scheme to test"
+desc " "
+desc "Example usage: fastlane test_build scheme:'NativeSigner'"
+lane :test_build do |options|
+  scheme = options[:scheme]
+  clear_derived_data
+  scan(
+    clean: true,
+    scheme: scheme,
+    device: "iPhone 13",
```
prybalko, line 12: Would it make sense to run tests on the latest iPod touch? I know it's not supported any more, but that's what people have

.github/workflows/testflight.yml
```diff
@@ -0,0 +1,106 @@
+name: Distribute TestFlight Build
+
+# on:
+#  workflow_dispatch: 
+#
+# on:
```
prybalko, line 6: > Discuss and agree on what should be triggers to run this GA (merge to /release/* branch, cron with daily / weekly builds, merging to master, "dispatch_workflow" with some user-based limitations)

I think that depends on who is going to use those builds. If it's primarily maintainers, then I think it's ok to build it on every `master` commit.

krodak: Posted new Notion page with voting on that 🙏🏻

ios/fastlane/lanes/lane_tests.rb
krodak: iPod Touch gen 7 will be removed from available simulators in September for newer XCode versions, given iOS 16 release.
As we don't have device-specific unit tests for now and I don't think we'll have for foreseeable future, only requirement here is to use device that is accessible in all versions of XCode 🙏🏻

montekki (approved):

PR #1219, opened 2022-08-16
chore: add integration tests to generate_message

prybalko:
Add integrations tests for `generate_message` CLI that doesn't do RPC calls


rust/generate_message/src/lib.rs
```diff
@@ -1065,14 +1065,16 @@
 //!
 //! `$ cargo run meta-at-block --url wss://westend-rpc.polkadot.io --block
 //! 780812df50c4006d1865742269fe4ca339c097e61d6279cce91ebc58f5aebada`
-#![deny(unused_crate_dependencies)]
+#![deny(unused)]
```
prybalko, line 1068: There is a false-positive using `unused_crate_dependencies` https://github.com/rust-lang/rust/issues/72686

rust/generate_message/src/lib.rs
```diff
@@ -1096,6 +1098,11 @@ use specs::gen_add_specs;
 mod error;
 pub use error::{Error, Result};
 
+lazy_static! {
+    pub(crate) static ref HOT_DB_PATH: PathBuf =
+        PathBuf::from(std::env::var("HOT_DB_PATH").unwrap_or_else(|_| HOT_DB_NAME.to_string()));
+}
```
montekki, line 1104: Would be cool to change these hardcoded paths to configuratble ones, so if this is starting to get changed I would propose just an `honest` cli flag instead of this env that would have a default value. That way it would have a `--help` doc and would be more usable by anyone

montekki: So that would look like you add flags to top-level parsing structure


https://github.com/paritytech/parity-signer/blob/7ebe0d3eb45bc390ea2da7d715fbbcf2b7422224/rust/generate_message/src/parser.rs#L16-L20

that would replace these constant values.

prybalko: That was my initial approach as well. 

However, `generate_message` is used not only as a cli, but as a library as well.  I could add a `--hot-db-path` flag with a default value, but that would only work for cli consumers. Whoever is using `generate_message` as a library (e.g. `signer` crate or `metadata-portal`) would need to pass the value explicitly. 

Crates within `parity-signer` would need to have `constants` as a dependency just to import `HOT_DB_NAME` from there. 

For external users, it's even more confusing, because some commands (like `make` or `sign`) don't require a database, but users would be forced to pass `hot-db-path` argument anyway.

Example of `generate_message` usage:
https://github.com/paritytech/metadata-portal/blob/master/cli/src/updater/generate.rs#L28

montekki: yeah well probably i was wrong about command-wide args and they can only be subcommand-specific flags for those subcommands that use them. And then the API that exists for use-cases like the portal could have an `Option`-al path value or something. In any case I am not a big fan of this

```rust
pub const HOT_DB_NAME: &str = "../database/database_hot";
```

We can do so much better.

prybalko: That makes sense. Fixed

rust/generate_message/src/helpers.rs
```diff
@@ -70,10 +80,14 @@ pub fn network_specs_from_entry(
...
+pub fn try_get_network_specs_to_send<P>(
     network_specs_key: &NetworkSpecsKey,
-) -> Result<Option<NetworkSpecsToSend>> {
-    let database = open_db(HOT_DB_PATH.to_str().unwrap())?;
+    db_path: P,
+) -> Result<Option<NetworkSpecsToSend>>
+where
+    P: AsRef<Path>,
+{
+    let database = open_db(db_path.as_ref().to_str().unwrap())?;
```
montekki, line 90: So there is no any reason left for `open_db` and others to take `&str` as a param now, it can also be generic over `AsRef<Path>` just like `sled::open` is.

prybalko: Totally agree, but that would trigger changes in many other places and it might be hard to review this PR.

prybalko: Right. My concern was that it will trigger changes in many places, and it'll harder to review this PR.

rust/generate_message/tests/make.rs
```diff
@@ -0,0 +1,35 @@
...
+fn it_signs() {
+    let unsigned = format!("{}/sign_me_load_metadata_polkadotV30", FOLDER);
+    let unverified = format!("{}/load_metadata_polkadotV30_unverified.txt", EXPORT_FOLDER);
+    let signed = format!(
+        "{}/load_metadata_polkadotV30_Alice-sr25519.txt",
+        EXPORT_FOLDER
+    );
+    remove_if_exists(&unverified);
+    remove_if_exists(&unsigned);
+    remove_if_exists(&signed);
```
montekki, line 17: Are all the tests racing on the same directories and file paths? If so maybe they should run in their isolated tmpdirs each?

prybalko: That makes sense. Thanks! Fixed

rust/generate_message/tests/unwasm.rs
```diff
@@ -0,0 +1,21 @@
+pub mod common;
+use crate::common::{assert_cmd_stdout, assert_files_eq, setup, teardown};
+
+use std::path::PathBuf;
+
+#[test]
+fn it_unwasm() {
+    let files_dir = PathBuf::from("./tests/it_unwasm");
```
montekki, line 8: _probably_ should make all these relative variables as absolute with `env!(CARGO_MANIFEST_DIR).push("./tests/it_unwasm")` but I cannot really come up with a counterexample to the relative dir off the top of my head.

montekki (approved):
lgtm ditto some nits

rust/generate_message/tests/derivations.rs
```diff
@@ -0,0 +1,21 @@
+pub mod common;
+use crate::common::{assert_cmd_stdout, assert_files_eq, remove_if_exists, setup, teardown};
+
+use std::path::PathBuf;
+
+#[test]
+fn it_derives() {
+    let files_dir = PathBuf::from("./tests/it_derives");
```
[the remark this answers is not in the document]
prybalko: 👍

rust/generate_message/tests/make.rs
```diff
@@ -0,0 +1,33 @@
...
+    let files_dir = PathBuf::from("./tests/it_signs");
+    setup(&files_dir);
+    let cmd = format!(
+        "load-metadata -f -a --hot-db-path {0} --files-dir {0}",
+        files_dir.to_string_lossy()
+    );
+    assert_cmd_stdout(&cmd, "");
+
+    let sign_cmd = format!(
+        "make --goal text --crypto sr25519 --msg load-metadata --payload sign_me_load_metadata_polkadotV30 --files-dir {0} --export-dir {0}",
```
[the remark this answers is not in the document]
prybalko: Done

PR #1222, opened 2022-08-17
chore: introduce PR template

krodak:
## Purpose ℹ️ 
This PR introduces new PR template as requested by @montekki 

## Scope 🔭
- add new single template to be used for PRs with optional sections

## Discussion🎙
Maybe it's better to have separate templates and parametrise PRs with something like `ios`, `mobile`, `backend`, but I think just simple one with optional sections might be a better start point 🙏🏻

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,0 +1,45 @@
+## Purpose ℹ️
+<!-- What is the goal of the proposed change? i.e.
+This PR aims to address issue: #1234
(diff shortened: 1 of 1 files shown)


montekki (approved):
I would say that probably emojis are not strictly necessary, so you could just remove them instead of trying to fix the `spellcheck` build. Other than that LGTM

prybalko (approved):

PR #1223, opened 2022-08-18
style: Refactored and UI updated tab bar

krodak:
## Purpose
This PR aims to update tab bar application-wide, according to new design, while providing better SwiftUI implementation and testability

## Scope
- `Tab` and `TabViewModel` that reflect UI state for given tab element
- `TabBarView` and `TabBarButton` that encapsulate updated UI using more reusable approach and moving most of mapping into view model builder and view model
- extracting `selectedTab` to be able to narrow dependency for `TabBarView` to lower coupling
- add unit test coverage to view model builder, `Tab` mapping

## Screenshots

| Before | After |
|-|-|
|||
|||


prybalko (approved):
👍

PR #1229, opened 2022-08-22
chore: include build.sh equivalent into Xcode build process

krodak:
## Purpose
This PR aims to remove need to manually run `build.sh` script and include logic from it within Xcode build process

## Scope
- `ios/scripts/build_libsigner.sh` that takes part of `build.sh` and make it possible to run it within Xcode `Build Phases`
- `ios/scripts/generate_database.sh` that takes another part of `build.sh` and make it possible to run it within Xcode `Build Phases`
- add custom `Build Rule` which uses `uniffi-bindgen` to generate `.swift` files based on `.udl` files added to `Compile Sources` `Build Phase`
- updating Fastlane and GA workflows to reflect change
- updating README.md to reflect change

## Discussion
Solution is mostly based on https://mozilla.github.io/uniffi-rs/swift/xcode.html and on their sample project that includes build script - most importantly, this part of script managed to resolve all the issues I had with integrating with Xcode 😂 
```bash
# XCode tries to be helpful and overwrites the PATH. Reset that.
PATH="$(bash -l -c 'echo $PATH')"
```

## Screenshots
CleanShot 2022-08-22 at 12 06 53@2x
CleanShot 2022-08-22 at 12 07 00@2x

CleanShot 2022-08-22 at 12 07 06@2x


prybalko (approved):
Tested. Works fine 👍

montekki (approved):
It would be cool if the build could detect the paths to swiftlint and others so on M1 this would not be necessary

```
sudo ln -s /opt/homebrew/bin/swiftgen /usr/local/bin/swiftgen
sudo ln -s /opt/homebrew/bin/swiftformat /usr/local/bin/swiftformat
sudo ln -s /opt/homebrew/bin/swiftlint /usr/local/bin/swiftlint
```

like $(which swiftlint) or something.

Other than that, seems to work, LGTM

krodak:
@montekki I'll try to capture this in separate PR, would indeed be good to have 👍🏻

PR #1232, opened 2022-08-23
ci: pin gha versions

sergejparity:
## Purpose
Improvement of security of GitHub Actions usage.

## Scope
In accordance with https://github.com/paritytech/ci_cd/wiki/Policies-and-regulations:-GitHub-Actions-usage-policies GitHub Actions used by the project. Either pinned to latest versions in case of trusted vendors. Or pinned to the hash of commit pointing to the current latest stable version.

## Related issues
https://github.com/paritytech/ci_cd/issues/114
https://github.com/paritytech/ci_cd/issues/464


montekki (approved):
Probably dependabot config should also be updated?

sergejparity:
> Probably dependabot config should also be updated?

It already has GHA in it https://github.com/paritytech/parity-signer/blob/efd9d13835d36fb9114fff8a238efb5f64f7a271/.github/dependabot.yml#L13-L17

montekki:
What I meant is how the dependabot bumps have to work with the new policy?

sergejparity:
There is no need to change dependabot config. It should offer updated version in the same way, it is clever enough to offer new tag to actions which pinned with the tag, and new hash to ones pinned with hash.
Goal is to attract some attention to changes in third party GHA's with the hope that someone will take a quick look to what changed at least :)

PR #1238, opened 2022-08-26
chore: improve QR payload parser

prybalko:
## Purpose

Improve QR payload parser

### Format Specification:
https://paritytech.github.io/parity-signer/development/UOS.html#qr-code-structure

## Scope

- use [nom](https://github.com/Geal/nom)
- re-write `get_payload` function

## Discussion
I'd propose re-write parser in small chunks. One format - one PR


montekki (approved):
Seems to work

PR #1240, opened 2022-08-29
feat: runtime SVG support on iOS with SwiftUI component

krodak:
## Purpose
This PR aims to showcase support for SVG, loading them from generic `Data` in runtime

## Scope
- addition of https://github.com/exyte/SVGView through SPM
- example `SVGIdenticon` UI component that takes `Data` as input
- example Preview

## Screenshots

| Example preview | 
|-|
||


.github/workflows/actions.yml
```diff
@@ -34,6 +34,14 @@ jobs:
           brew install swiftformat
           brew install xcbeautify
 
+      - name: Get cached Swift Packages managed by Xcode
+        uses: actions/cache@v2
```
prybalko, line 38: Any reason not to use `@v3` version?

Here is an example:
https://stackoverflow.com/a/72276931

krodak: Not really, wasn't aware of v3, just used my usual script, thanks, bumping it now! 🙏🏻

krodak:
Closed in favour of https://github.com/paritytech/parity-signer/pull/1453

Issue #1155, opened 2022-06-29
Parsers improvements


montekki:
Code that handles user inputs (metadata, txs, qrs) should be improved.

It should not
1. AND, XOR or OR things with random undocumented numeric values such as `0x1000000`.
2. Panic on bad user inputs.
3. match against undocumented values that no one understands what their values are or how stable these values are from one chain to another.

It should
1. At least be fuzzed.

https://github.com/paritytech/parity-signer/blob/c54813197fd9e627d0a3d46b6a65ab233bf4803b/rust/transaction_parsing/src/lib.rs#L58-L69

https://github.com/paritytech/parity-signer/blob/224d3dc034993106c8eafd333cabb4057725f38b/rust/qr_reader_phone/src/process_payload.rs#L45-L57

https://github.com/paritytech/parity-signer/blob/224d3dc034993106c8eafd333cabb4057725f38b/rust/qr_reader_phone/src/process_payload.rs#L101-L111

Slesarew:
> AND, XOR or OR things with random undocumented numeric values such as 0x1000000

This is well documented: https://paritytech.github.io/parity-signer/development/UOS.html#raptorq-multipart-payload

> Panic on bad user inputs.

Where does it panic? Please point me or propose bad user input

> match against undocumented values that no one understands what their values are or how stable these values are from one chain to another.

It is well-documented here. https://paritytech.github.io/parity-signer/development/UOS.html

Please advise

montekki:
a bad user input is easy to construct by hand

```patch
diff --git a/rust/signer/src/lib.rs b/rust/signer/src/lib.rs
index 801482c2..b1b0c503 100644
--- a/rust/signer/src/lib.rs
+++ b/rust/signer/src/lib.rs
@@ -219,5 +219,10 @@ ffi_support::define_string_destructor!(signer_destroy_string);
 
 #[cfg(test)]
 mod tests {
-    //use super::*;
+    use crate::qrparser_try_decode_qr_sequence;
+
+    #[test]
+    fn try_decode_qr_test() {
+        qrparser_try_decode_qr_sequence("[\"aa\"]", true).unwrap();
+    }
 }

```

reliably gets you a panic

```
running 1 test
thread 'tests::try_decode_qr_test' panicked at 'range end index 4 out of range for slice of length 1', library/core/src/slice/index.rs:73:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test tests::try_decode_qr_test ... FAILED
[...]

PR #1247, opened 2022-08-31
chore: #1155 improve QR reader parser

prybalko:
## Purpose

improve readablilty and relibility of QR reader parser

## Parser specs
https://paritytech.github.io/parity-signer/development/UOS.html#raptorq-multipart-payload

## Scope

- re-write paser using `nom`
- add fuzzing (it takes hours to run, so I didn't put it in CI)
- fix documentation

## Screenshots

| Before |  After |
|-|-|
|image|image
|


rust/qr_reader_phone/src/parser.rs
```diff
@@ -1,9 +1,67 @@
...
+        //
+        // Thus, if there are no additional packets, probability is ~ 0.99609.
+        // If one additional packet is added, it is ~ 0.99998.
+        // It was decided to add one additional packet in the printed estimate, so that
+        // the user expectations are lower.
+        self.size / (self.payload.len() as u32) + 1
+    }
+}
+
+impl TryFrom<&Vec<u8>> for RaptorqFrame {
```
montekki, line 33: `impl TryFrom<&[u8]>` ?

rust/qr_reader_phone/src/parser.rs
```diff
@@ -1,9 +1,67 @@
...
+    }
+}
+
+pub(crate) struct LegacyFrame {
+    pub(crate) total: u16,
+    pub(crate) index: u16,
+    pub(crate) data: Vec<u8>,
+}
+
+impl TryFrom<&Vec<u8>> for LegacyFrame {
```
montekki, line 52: and here

rust/qr_reader_phone/src/parser.rs
prybalko: I get
```
 the trait `From<&Vec<u8>>` is not implemented for `RaptorqFrame`
```

montekki: Yes everywhere you call `try_from(&payload)` you would have to call `try_from(payload.as_ref())` and it should work

rust/qr_reader_phone/fuzz/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "qr_reader_phone-fuzz"
+version = "0.0.0"
+authors = ["Automatically generated"]
+publish = false
+edition = "2018"
```
montekki, line 6: The regular stuff like everywhere else should be used

rust/qr_reader_phone/src/parser.rs
prybalko: Thanks! But is there a reason to prefer one over the other? Or it's more a matter of taste?

montekki: No actually it is quite useful. When you choose `Vec<u8>` on a type-level you are binding yourself to work only with this single type. When you choose `&[u8]` you are "saying" something like "i can work with any type that can act as a contiguous piece of memory containing bytes". For idiomatic examples you can look at how the `std::io` traits are composed. For example check out the [`std::io::Write`](https://doc.rust-lang.org/std/io/trait.Write.html) trait.

prybalko: Awesome, thanks!

montekki (approved):

PR #1249, opened 2022-09-01
style: Key Set screen redesign

krodak:
## Purpose
This PR provides redesign for `Key Set` screen and add new UI component, being Navigation Bar as featured in Figma.

## Scope
- `DerivedKeyRow` that reflects single Derived Key cell in Key Set screen
- `KeyDetailsView` that reflects Key Set screen
- `NavigationBarView` component that mimics UI and behaviour of system navigation bar
- minimal set of changes within `NewSeedMenu` to enable presentation from new `KeySetList` through `sheet` presentation - UI was not modified, so it looks bad, but `NewSeedMenu` is next component that will be rewritten fully, so this state is just transitory to enable proper action performance. 

### Static preview

| Light mode | Dark mode |
|-|-|
|||

### Screen interactions

| Light mode | Dark mode |
|-|-|
|||

### Navigation bar

| Light mode | Dark mode |
|-|-|
|||


montekki:
Why does the `create derived key` button exist on the screen with all keys? It is not clear to which key it is going to be applied if any

krodak:
@montekki I guess this is question regarding updated designs 🤷🏻‍♂️ This is details screen for single key set.
If you change a network with button next to `Derived Key` header, and then tap on `Create Derived Key`, you'd create key for that network.
As for key, when you are on create new key screen, I think you have that info 
CleanShot 2022-09-01 at 12 04 14@2x

Dmitry-Borodin (approved):

montekki (approved):

PR #1250, opened 2022-09-01
fix: Removing requirement of biometric

Dmitry-Borodin:
Removed requirement of strong biometric authentication for new android devices. 
User can now use any authentication that it set on device before unlocking secure storage to get the master key.


krodak (approved):
LGTM 🚀

montekki (approved):

PR #1251, opened 2022-09-01
feat: expose secret export function

prybalko:
## Purpose

Expose `generate_secret_key_qr` function that generates QR code with secret key

## Scope

Add `generate_secret_key_qr` to `signer.udl`


prybalko:
@krodak please let me know if that works

montekki (approved):

rust/signer/src/signer.udl
```diff
@@ -550,6 +550,9 @@ namespace signer {
     [Throws=ErrorDisplayed]
     string qrparser_try_decode_qr_sequence([ByRef] string data, boolean cleaned);
 
+    [Throws=ErrorDisplayed]
+    MKeyDetails generate_secret_key_qr([ByRef] string dbname, [ByRef] MultiSigner multisigner, [ByRef] string expected_seed_name, [ByRef] string network, [ByRef] string seed_phrase, string? pwd);
```
krodak, line 554: Q1: Do we need to pass `dbname`? I remember @montekki mentioning that this should be done just in the first call from mobile apps. While it might not make sense to refactor state machine, maybe we shouldn't add this to new calls.

Q2: How do we get `MultiSigner` in this context? Do we need to retain some data from state machine in previous screens? If so, could you point to which data model / machine state exactly?

Q3: is `network` a network name or `id`? Also, which model / machine state it should come from? 

Q4: how do we get password of key is password protected? Do we have UI flow designed for that?

prybalko: A1: If we don't pass `dbname`, then it needs to be stored somewhere. Normally we store everything in the database, but `dbname` _is_ the path to the database. I can figure something out, but that would require some refactoring. 

A2: I'll look into that

A3: `network` is a hex `network_id`. Not sure where it's coming from. I'll look into that as well

A4: user enters the password. We don't store it aanywhere. No UI flow for it at the moment.

krodak: A1 - no worries, I'll pass it somehow, as this mobile app is still one fat singleton 😂 
A2 / A3 - if it's something that I could get from `MKeysCard`, `MSeeds`, `MKeys`, then I'd be able to pass this data through view hierarchy through Service Locator anti-pattern, as there are no connections between views due to state machine

krodak: @montekki maybe you'd have more info on A2 / A3 🙏🏻

montekki: 1. you have to pass dbname in this case but we can dodge that in the future when I finish #1265 
1. it should be somewhere in the card types in the view or you could probably look it up
1. i have no idea

rust/db_handling/src/identities.rs
```diff
@@ -1294,17 +1294,21 @@ fn prepare_secret_key_for_export(
...
+    db_path: P,
     multisigner: &MultiSigner,
     expected_seed_name: &str,
-    network_specs_key: &NetworkSpecsKey,
+    network_specs_key_hex: &str,
     seed_phrase: &str,
-    pwd: Option<&str>,
-) -> Result<MKeyDetails> {
-    let network_specs = get_network_specs(database_name, network_specs_key)?;
+    mut pwd: Option<String>,
```
Dmitry-Borodin, line 1303: would be nice to add simple comment  to which parameters it is. Like what should we pass to pwd.

rust/signer/src/signer.udl
krodak: A2 - I don't see any references to `MultiSigner` within iOS repo right now, beside this generated code:
```swift
public typealias MultiSigner = [String]
...
public enum VerifierValue {
    case standard(m: MultiSigner)
}
```
And only code that refers to that is:
```swift

extension VerifierValue {
    func show() -> String {
        switch self {
        case let .standard(value):
            return value[0]
        }
    }
}
```
And only use of that is in `Logs` screen for 2 types of cards: `MEventMaybeDecoded` -> `.transactionSignError`, `.transactionSigned`

So this won't be data that can be "fetched" easily when exporting private key, especially that it's based on assumption we'd have those types of items in `Logs`.

So we'll probably need some handler to get it

rust/db_handling/src/identities.rs
prybalko: I try to make arg names self-explanatory, so it's not nessesary add any comments. You're right, `pwd` is not quite clear name, I can rename it to `key_password` or `secret_password`

rust/signer/src/signer.udl
prybalko: 2. I made `generate_secret_key_qr` to take `public_key` instead of `multisigner`. It's accessable from `MKeyDetails`. 
3. I renamed `network` to `network_specs_key_hex`. It comes from `MDeriveKey`.

Let me know if that helps

Dmitry-Borodin: I would add information about where to get public_key and network_specs_key_hex to a comment. Comments in code are not great, but I think much better than searching for this information in github PR comments.

prybalko: Makes sense, added

Dmitry-Borodin (approved):

krodak (approved):
@prybalko I can confirm that this works and generates valid QR code, let's merge it 🎉

prybalko:
> @prybalko I can confirm that this works and generates valid QR code, let's merge it 🎉

Great! I'll fix the android part and will merge the PR

PR #1263, opened 2022-09-05
fix: update android runner image to fix compilation issue in github a…

Dmitry-Borodin:
I still don't know the root cause, but this task change fixing issue with github actions and it runs basic unit tests (which we have none, but we need to check project compiles)

### DIFF
--- a/.github/workflows/android-app.yml
+++ b/.github/workflows/android-app.yml
@@ -6,11 +6,10 @@ on:
     branches:
       - master
       - stable
-      - change-android-runner
 
 jobs:
   build:
-    name: Build android
+    name: Build
     runs-on:            ubuntu-22.04
     steps:
       - name:           Cancel Previous Runs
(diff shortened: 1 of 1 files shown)


.github/workflows/android-app.yml
```diff
@@ -6,11 +6,12 @@ on:
     branches:
       - master
       - stable
+      - change-android-runner
```
krodak, line 9: ✂️

.github/workflows/android-app.yml
```diff
@@ -6,11 +6,12 @@ on:
...
 
 jobs:
   build:
-    name: Build
-    runs-on:            ubuntu-latest
+    name: Build android
```
krodak, line 13:
```suggestion
    name: Build
```

As `Build` is required step in repo settings and we shouldn't change it

Dmitry-Borodin: Fixed, 
but I see this is the only flow and job called Build. Do you know which requirement it is?

montekki (approved):

PR #1266, opened 2022-09-06
feat: Move android to root folder

Dmitry-Borodin:
Moved android project to a root folder so that we can just open root in android studio and work with all the files seamelessly.


montekki:
DB Blobs shouldn't have made it into the repo, they are generated at build time and ignored somewhere in the .gitignore, please check.

Also maybe it is not such a good idea since the android app and the ios app are unlikely to both live at the root of the project and if that is the case I see no reason to make android a first-class citizen with that regard cC @krodak

krodak:
No issues on iOS development - if for Android setup it's easier to go through code like that - for iOS we are only using XCode (there is `[AppCode](https://www.jetbrains.com/objc/)` but with SwiftUI it doesn't make sense to use) and if I go through Rust code I'm using Visual Studio Code, so on my side current structure for `ios` folder and setup makes sense 👍🏻

krodak:
No issues on iOS development - if for Android setup it's easier to go through code like that - for iOS we are only using XCode (there is `[AppCode](https://www.jetbrains.com/objc/)` but with SwiftUI it doesn't make sense to use) and if I go through Rust code I'm using Visual Studio Code, so on my side current structure for `ios` folder and setup makes sense 👍🏻

montekki (approved):

Issue #1301, opened 2022-09-20
Uniffi doesn't construct exceptions properly in JVM


Dmitry-Borodin:
ErrorDisplay is defined and Str as a Subclass. 
While in case of error ErrorDisplayed$Str is thrown and calling code receiving 
java.lang.RuntimeException: java.lang.reflect.InvocationTargetException

Exception example 
```
Process: io.parity.signer, PID: 9371
                                                                                                    java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
                                                                                                    	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:502)
                                                                                                    	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)
                                                                                                    Caused by: java.lang.reflect.InvocationTargetException
                                                                                                    	at java.lang.reflect.Method.invoke(Native Method)
                                                                                                    	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
                                                                                                    	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930) 
                                                                                                    Caused by: io.parity.signer.uniffi.ErrorDisplayed$Str
                                                                                                    	at io.parity.signer.uniffi.FfiConverterTypeErrorDisplayed.read(signer.kt:5427)
                                                                                                    	at io.parity.signer.uniffi.FfiConverterTypeErrorDisplayed.read(signer.kt:5422)
                                                                                                    	at io.parity.signer.uniffi.FfiConverter$DefaultImpls.liftFromRustBuffer(signer.kt:149)
                                                                                                    	at io.parity.signer.uniffi.FfiConverterRustBuffer$DefaultImpls.liftFromRustBuffer(signer.kt:161)
                                                                                                    	at io.parity.signer.uniffi.FfiConverterTypeErrorDisplayed.liftFromRustBuffer(signer.kt:5422)
                                                                                                    	at io.parity.signer.uniffi.FfiConverterTypeErrorDisplayed.liftFromRustBuffer(signer.kt:5422)
                                                                                                    	at io.parity.signer.uniffi.FfiConverterRustBuffer$DefaultImpls.lift(signer.kt:162)
                                                                                                    	at io.parity.signer.uniffi.FfiConverterTypeErrorDisplayed.lift(signer.kt:5422)
[...]

PR #1303, opened 2022-09-20
feat: show rust-caused error message in android logs when app is cras…

Dmitry-Borodin:
…hing

Workaround for #1301 
So we at least always see what was the message since it's not in Exception.getMessage(), only in class member String.

Reasoning:
Currently we don't handle rust errors. Some of them are recoverable, and in best case we should handle those and provide options to restore the app.
In a meanwhile, when we crash, this PR is about at least to print an error message to the logs. Because currently it is possible to see the error message 's' in debugging breakpoint only, which is sub-optimal.

cc @montekki  added section reasoning above ^^


montekki:
i don't understand the logic here, why would this exception be unhandled.


_every_ error situation in rust side that is not actually terminal: for example user has added an invalid phrase or something. As such this

https://github.com/paritytech/parity-signer/blob/aee88ac1b9a21d75918b2811e6fdb83660d79392/rust/signer/src/lib.rs#L94-L100

is turned to this:

https://github.com/paritytech/parity-signer/blob/aee88ac1b9a21d75918b2811e6fdb83660d79392/rust/signer/src/signer.udl#L542-L543

And will throw the exception type `ErrorDisplayed`. But obviously there should be graceful error handling designed but just un-catching and terminating probably is not the best solution. This is more true for when all the errors that ended up in alert UI elements when will also and up in these exceptions when we move from the stateful navigation.

Dmitry-Borodin:
1) Currently JVM Runtime exception is thrown, not to our exception. That is something we need to fix in #1301 before being able to simply catch FFI exceptions. Now I can catch only all possible exceptions, but then it's not clear what to do with them.

2) I see some of exceptions are not really restorable. Like when DB is broken. Do we already have list of errors and how they can be restored by any chance? This is nice task to be done. Just I don't see why we shouldn't at least show error message when we didn't catch something.

cc @krodak what do you do for FFI errors?

krodak:
@Dmitry-Borodin they are just displayed in standard Alert component for now, I plan to present them in redesigned bottom sheet component

montekki (approved):

PR #1324, opened 2022-09-28
chore: Added connection and request timeout for the WsClientBuilder

pgolovkin:
## Purpose
The main goal is to fix Metadata portal problem. It fails because can't get response from some nodes in time.
The solution is to change timeout and request timeout.

## Scope
-- Changed the connection timeout to 60 seconds and request timeout to 120 seconds

## TODOs

- [ ] Maybe connection and request parameters should be passed to the API?

### DIFF
--- a/rust/generate_message/src/fetch_metadata.rs
+++ b/rust/generate_message/src/fetch_metadata.rs
@@ -117,6 +117,9 @@ lazy_static! {
     static ref PORT: Regex = Regex::new(r"^(?P<body>wss://[^/]*?)(?P<port>:[0-9]+)?(?P<tail>/.*)?$").expect("known value");
 }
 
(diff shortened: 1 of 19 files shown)


rust/generate_message/src/fetch_metadata.rs
```diff
@@ -156,6 +157,8 @@ fn address_with_port(str_address: &str) -> String {
 #[tokio::main]
 pub async fn fetch_info(str_address: &str) -> Result<FetchedInfo> {
     let client = WsClientBuilder::default()
+				.connection_timeout(Duration::from_secs(60))
+				.request_timeout(Duration::from_secs(120))
```
montekki, line 161: Create const values

montekki (approved):

prybalko (approved):

PR #1341, opened 2022-10-07
chore: bump clap

montekki:
## Purpose

Bumps `clap` dependency to new major version `4.0`.

### DIFF
--- a/rust/generate_message/src/parser.rs
+++ b/rust/generate_message/src/parser.rs
@@ -627,7 +627,7 @@ pub enum Crypto {
 
 #[derive(clap::Args, Debug, Clone)]
 #[clap(group(clap::ArgGroup::new("signature")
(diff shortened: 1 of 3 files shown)


prybalko:
`cargo check --features clap/deprecated` shows many deprecation warnings. I think it's better to fix them. I can do that

krodak (approved):

rust/generate_message/src/parser.rs
```diff
@@ -471,14 +471,14 @@ impl std::fmt::Display for Set {
 #[derive(clap::Args, Debug)]
 pub struct Make {
     /// payload
-    #[clap(long, name = "msg", value_parser)]
```
prybalko, line 474: `value_parser` is a default value => no need to specify. 
`Name` is the same as a variable name => no need to specify.

prybalko (approved):

PR #1343, opened 2022-10-09
feat: export address info

montekki:
## Purpose

This introduces an API to export information about all addresses and
keys Signer knows of to external wallets via animated QR codes.


rust/navigator/src/lib.rs
```diff
@@ -104,3 +110,11 @@ pub fn update_seed_names(seed_names: Vec<String>) {
         }
     }
 }
+
+pub fn export_key_info(dbname: &str) -> Result<MKeysInfoExport, String> {
```
prybalko, line 114: I think tests might be helpful for this

rust/db_handling/src/identities.rs
```diff
@@ -96,6 +97,63 @@ lazy_static! {
...
+
+#[derive(Encode, Decode)]
+pub struct ExportAddrsV1 {
+    pub addrs: Vec<AddrInfo>,
+}
+
+#[derive(Encode, Decode)]
+pub struct AddrInfo {
+    /// Name of the key
+    pub name: String,
```
pgolovkin, line 113: What is the name? As I see in the Parity Signer UI there is a root key name and others are dervied keys for the specific network. What will be encoded or composed in this parameter?

rust/db_handling/src/identities.rs
```diff
@@ -96,6 +97,63 @@ lazy_static! {
...
+
+    /// Address in the network.
+    ///
+    /// This is either `ss58` form for substrate-based chains or h160 form for ethereum based
+    /// chains
+    pub address: String,
```
pgolovkin, line 119: Shouldn't the public key be here also?

montekki: As we discussed with the Ethereum keys case I see no need for UI to know the public key and to address things by public key. Public key "equals" `AddressId32` only for `sr25519` and `ed25519`, it does not for `ecdsa` that goes into `MultiSignature` and it does not for ethereum addrs.

rust/signer/src/lib.rs
```diff
@@ -223,6 +223,10 @@ fn history_seed_name_was_shown(seed_name: &str, dbname: &str) -> anyhow::Result<
         .map_err(|e| format!("{}", e))
 }
 
+fn export_key_info(dbname: &str) -> anyhow::Result<MKeysInfoExport, String> {
```
Dmitry-Borodin, line 226: Please expose existing function to get bitmap qr code from vec<u8> as well. In addition to current request. It will be easier to use it for now rather than inventing logic on android side. 
There is no native solution and all decent libraries I see areworking with string exclusively. 
I can request it 1 by 1 in parallel or as an array in 1 shot - doesn't matter.

montekki: Done

rust/signer/src/lib.rs
```diff
@@ -223,6 +223,17 @@ fn history_seed_name_was_shown(seed_name: &str, dbname: &str) -> anyhow::Result<
...
```
Dmitry-Borodin, line 226: To export only selected keys, which unique parameters of those keys will be needed here?
Is "`seedName`: String" unique? Because we have only MSeeds to show those screens on UI right now.

rust/db_handling/src/identities.rs
```diff
@@ -96,6 +97,168 @@ lazy_static! {
...
+
+    /// Public key of the root key.
+    ///
+    /// `None` if user has created no root keys.
+    pub multisigner: Option<MultiSigner>,
+
+    /// Derived keys.
+    pub derived_keys: Vec<AddrInfo>,
+}
+
```
montekki, line 123: @pgolovkin the structures have been updated to show hierarchical information, that should suffice your usecases

rust/navigator/src/lib.rs
```diff
@@ -61,3 +64,16 @@ pub fn update_seed_names(seed_names: Vec<String>) -> Result<()> {
...
 }
+
+/// Export key info with derivations.
+pub fn export_key_info(
+    dbname: &str,
+    selected_names: Option<Vec<String>>,
+) -> Result<MKeysInfoExport> {
+    let export_all_addrs = export_all_addrs(dbname, selected_names)?;
+
+    let data = [&[0x53, 0xff, 0xde], export_all_addrs.encode().as_slice()].concat();
```
montekki, line 75: @pgolovkin the data in the QR would now have this prefix, it is not yet described in the docs. It is needed for compatibility with UOS protocol so two Signer devices can talk to each other.

rust/signer/src/lib.rs
krodak: @Dmitry-Borodin api has been updated with optional string array to handle that

prybalko (approved):

PR #1350, opened 2022-10-11
fix: no panics on not found keys

montekki:
## Purpose

The code unwraps unnecessarily on not found keys. Also some parts of the
UI use public keys for addressing instead of addresses. Untils the addressing
issue is fixed, support both:
 * Search for an Ethereum addr
 * Search for public addr

In the future this has to be replaced with typed addresses instead of byte blobs or `String`s.


prybalko (approved):
Tested with steps:
1. wipe all data
1. Scan QR with `moonbase` network specs and approve it
1. Create new key set
1. Go to Settings -> Networks -> moonbase-etherium
1. Hit `...`, select sign network specs
1. Select `//moonbase` key

Works well

PR #1354, opened 2022-10-12
chore: bump opencv version 

montekki:
## Purpose

Bumps `opencv` to a new version and uses `QrCodeDetector`.


.github/workflows/android-app.yml
```diff
@@ -10,7 +10,7 @@ on:
 jobs:
   build:
     name: Build
-    runs-on:            ubuntu-22.04
+    runs-on:            ubuntu-latest
```
sergejparity, line 13:
```suggestion
    runs-on:            ubuntu-20.04
```
I would suggest to use `ubuntu-20.04` explicitly, as it is expected that `ubuntu-latest` will start pointing to `ubuntu-22.04` soon. https://github.blog/changelog/2022-08-09-github-actions-ubuntu-22-04-is-now-generally-available-on-github-hosted-runners/

.github/workflows/android-app.yml
```diff
@@ -75,7 +75,7 @@ jobs:
 
   debug:
     name: Build unsigned APKs
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
```
sergejparity, line 78:
```suggestion
    runs-on:            ubuntu-20.04
```
I would suggest to use `ubuntu-20.04` explicitly, as it is expected that `ubuntu-latest` will start pointing to `ubuntu-22.04` soon. https://github.blog/changelog/2022-08-09-github-actions-ubuntu-22-04-is-now-generally-available-on-github-hosted-runners/

sergejparity, line 78:
```suggestion
    runs-on: ubuntu-20.04
```
I would suggest to use `ubuntu-20.04` explicitly, as it is expected that `ubuntu-latest` will start pointing to `ubuntu-22.04` soon. https://github.blog/changelog/2022-08-09-github-actions-ubuntu-22-04-is-now-generally-available-on-github-hosted-runners/

montekki: I would actually want to switch to `22.04` but even with that the build would not work it looks like

krodak (approved):

PR #1363, opened 2022-10-16
chore: simplify nav data models

prybalko:
## Purpose
Remove repetition in nav data models

## Scope
- reuse `Address` model
- add `MAddressCard` 
- reuse `MKeysCard` instead of `MSeedKeyCard`
- reuse `MSCId` instead of `MSCAuthorPlain`
- reuse `NetworkSpecs` in `OrderedNetworkSpecs`

## ToDo
To unite:
- [x] `NetworkSpecs` & `NetworkSpecsToSend`
- [x] `MSeedKeyCard` & `MKeysCard`
- [x] `MSCId` & `MSCAuthorPlain`


rust/db_handling/src/identities.rs
```diff
@@ -1362,16 +1363,10 @@ where
 
     let style = address_details.encryption.identicon_style();
     let address = Address {
-        base58: print_multisigner_as_base58_or_eth(
-            multisigner,
-            Some(network_specs.base58prefix),
-            address_details.encryption,
-        ),
```
montekki, line 1369: Why is address not the part of `Address`? It _literally_ the address, be it `ss58` or ethereum addr? I would suggest leaving it here but renaming to something address-format agnostic

montekki: The address could probably be an enum with variants `ss58` and `eth`  to add a bit of structure

prybalko: > It literally the address

Not really. It's an address + encryption. 
`print_multisigner_as_base58_or_eth` depends on `network_specs`, but `AddressDetails` has a list of `NetworkSpecsKey`.
https://github.com/paritytech/parity-signer/blob/master/rust/definitions/src/users.rs#L39

So, in general, it's not possible to define `ss58` having only `AddressKey`

prybalko: Agreed. And we better have this in a separate PR.

montekki:
Also have to be united:

* `NetworkSpecs`  & `NetworkSpecsToSend`
* `MSeedKeyCard` & `MKeysCard`
* `MSCId` & `MSCAuthorPlain`

Dmitry-Borodin:
Why do we need to introduce MAddressCard from rust side?
It seems to me that we should get data model which is Address and whether user should be able to multiselect or not should come from where this address is shown as UI element and backend should not define view characteristics, only data.
New keys and multiselect keys are even 2 separate views, because use cases are different, they are not mixed with each other.

prybalko:
> backend should not define view characteristics

Absolutely. But I think it deserves a separate PR, this one is already too big IMO

montekki (approved):

Issue #1360, opened 2022-10-13
Next button out of screen


Tbaut:
I noticed this on a build from `master` from yesterday.
This is a on Pixel 3a 

the spacing at the top isn't ideal. The Next button is only partly visible, and I can't scroll to have the button in full.

![image](https://user-images.githubusercontent.com/33178835/195588657-2f459783-a345-478d-be48-323843c0e46d.png)

Dmitry-Borodin:
Thank you. That is recent regression, will fix shortly

PR #1371, opened 2022-10-18
fix: android wrong paddings in screen with documents

Dmitry-Borodin:
## Purpose
hotfix #1360


android/src/main/java/io/parity/signer/MainActivity.kt
```diff
@@ -168,7 +168,12 @@ fun SignerApp(signerDataModel: SignerDataModel) {
 			}
 			OnBoardingState.No -> {
 				if (shieldAlert.value == AlertState.None) {
-					Scaffold { padding ->
+					Scaffold(
+						modifier = Modifier
+							.navigationBarsPadding()
+							.captionBarPadding()
```
Dmitry-Borodin, line 174: This is the actual fix

montekki (approved):

PR #1383, opened 2022-10-24
feat: Android export keys info

Dmitry-Borodin:
## Purpose
Export key info implemented on Android

## Scope
Key info screen redesigned
Key info export screens, navigation and qr code added
Key details screen is mostly implemented and disabled with feature flag, it's UI is out of scope in this PR

Note looks like UI is not so smooth and around 15ms is needed just to update each image.  But it works fine as delay is much bigger.
Please let me know if you'll notice measurable performance issues.

| Bright | Dark |
|-|-|
| image | image |
| image | image |
| image | image |
| image | image |


tetekinandrey (approved):

montekki:
We should wait for testing from @pgolovkin and if all good, then we can merge

pgolovkin:
Works well. Shouldn't the `Export private key` be renamed to anything else? There is address export not a private key export.

montekki:
"Export Private Key" is wrong, should be "Export Key Sets"

Dmitry-Borodin:
Thank you, fixed forgotten localisation for menu.

PR #1397, opened 2022-11-03
feat: bulk signing

montekki:
## Purpose

Sign bulks of transactions.


rust/db_handling/src/identities.rs
```diff
@@ -100,6 +100,40 @@ lazy_static! {
     static ref REG_PATH: Regex = Regex::new(r"^(?P<path>(//?[^/]+)*)(///(?P<password>.+))?$").expect("known value");
 }
 
+#[derive(Clone, Encode, Decode)]
+pub enum SignaturesBulk {
+    #[codec(index = 4)]
+    V1(SignaturesBulkV1),
```
montekki, line 106: @pgolovkin this is essentially the type that is proposed to use to provide bulks of signatures to external apps. The current variant starts with the byte `0x04` in order not to conflict with the existing format of exporting single signatures which is essentially `MultiSignature::encode()` and would start with bytes in set `[0x00, 0x01, 0x02]` for respective signature scheme variants in `MultiSignature`.

rust/generate_message/src/helpers.rs
```diff
@@ -686,6 +688,38 @@ pub fn generate_key_info_export_to_qr<P: AsRef<Path>>(
...
+    .collect();
+    let v1_bulk = TransactionBulkV1 {
+        encoded_transactions,
+    };
+    let bulk = TransactionBulk::V1(v1_bulk);
+
+    let payload = [&[0x53, 0xff, 0x04], bulk.encode().as_slice()].concat();
+
+    match output_format {
+        Goal::Qr => generate_qr_code(&payload, chunk_size, 8, dst_file)?,
```
prybalko, line 715: It's not obvious what `8` means. Maybe make it a constant?

rust/generate_message/src/helpers.rs
```diff
@@ -686,6 +688,38 @@ pub fn generate_key_info_export_to_qr<P: AsRef<Path>>(
...
+    tx_count: usize,
+    chunk_size: u16,
+    from: String,
+    output_format: Goal,
+) -> Result<()> {
+    let encoded_transactions = (0..tx_count).map(|_| {
+
+let line = "0102".to_string() + 
+    &from +
+"a80403004adb5312dbd3a1bc28610deb1fb631110bbcccb6e7fb97e18501509d5565ea760b00a014e3322615010000682400000e000000e143f23803ac50e8f6f8e62695d1ce9e4e1d68aa36c1cd2cfd15340213f3423e2faa9938ec10c2627d0fd5d20214aba3bf281e94a6a8f50d19b4e0ce3b253d65e143f23803ac50e8f6f8e62695d1ce9e4e1d68aa36c1cd2cfd15340213f3423e";
```
prybalko, line 703: Is it some test data?

rust/definitions/src/navigation.rs
```diff
@@ -13,31 +13,44 @@ pub struct SeedNameWithIdenticon {
...
+
 /// Enum containing card sets for four different outcomes:
 /// importing derivations (Derivations), signing (Sign),
 /// accepting (Stub) and reading, for example, in case of an error (Read)
 #[derive(PartialEq, Eq, Debug, Clone)]
 pub enum TransactionAction {
     Derivations {
-        content: TransactionCardSet,
-        network_info: OrderedNetworkSpecs,
+        content: Box<TransactionCardSet>,
```
prybalko, line 38: Are there any benefits of making it boxed?

montekki: Yeah that fixes a `clippy` lint about the hugely different sizes of `enum` variants here. They become different since one of the params becomes `Vec` in this changeset as you may see.

rust/generate_message/src/helpers.rs
montekki: I believe that the changes to `generate_message` can be rolled back before merging this anyway but yes

rust/signer/src/signer.udl
```diff
@@ -492,7 +492,7 @@ interface ScreenData {
     Settings(MSettings f);
     Log(MLog f);
     LogDetails(MLogDetails f);
-    Transaction(MTransaction f);
+    Transaction(sequence<MTransaction> f);
```
Dmitry-Borodin, line 495: Any type transactions? Or we can any multiple "signature" transactions or any single one?
If second I would have separate data model for it.

rust/transaction_parsing/src/lib.rs
```diff
@@ -56,15 +58,44 @@ where
...
+    match bulk {
+        TransactionBulk::V1(b) => {
+            let mut actions = vec![];
+
+            let mut checksum = 0;
+            for t in &b.encoded_transactions {
+                let encoded = hex::encode(t);
+                let encoded = "53".to_string() + &encoded;
+                let action = parse_transaction(&encoded, &dbname, true)?;
+                if let TransactionAction::Sign {
```
Dmitry-Borodin, line 85: If bulk transactions can e any transactions, shouldn't we handle all transaction actions here?

montekki: So at the moment we are suppose that a bulk has a no-nonsense content and contains only `Sign` variant and ignores the rest. But I want to reduce the surface of the changeset and keep the code as brief as possible by re-using the existing solutions, so no specific new variant is added to the types, just the existing variants are enhanced to contain a set of `Sign` transactions instead of one if that makes sense.

Dmitry-Borodin: Ok, then We need to add this check on platform code and probably to UI, that user can understand that multisign should work only for signing transactions. cc @tetekinandrey

rust/db_handling/src/identities.rs
```diff
@@ -100,6 +100,40 @@ lazy_static! {
...
+impl From<SignaturesBulkV1> for SignaturesBulk {
+    fn from(s: SignaturesBulkV1) -> Self {
+        Self::V1(s)
+    }
+}
+
+#[derive(Clone, Encode, Decode)]
+pub struct SignaturesBulkV1 {
+    /// `MultiSignature`s in encoded form.
+    signatures: Vec<Vec<u8>>,
```
montekki, line 118: @pgolovkin I am going to change this to typed signatures: `Vec<MultiSignature>` if that is ok with you

krodak (approved):
LGTM 🚀

PR #1429, opened 2022-11-21
feat: import derivations

prybalko:
## Purpose
Import derivations from QR code

## Scope
Rust:
- alter `TransactionAction::Derivations` navstate to include data from QR
- introduce `SeedKeysPreview` model to hold data for frontend
- add `import_derivations` api call to import derivations
- add `populate_derivations_has_pwd` API

iOS:
- update `TCDerivations` to reflect current design
- add error states UI
- add Rust calls to import derivations
- add UI flows for adding missing network and recover key set

## TODOs
Rust:
- [x] Handle `goForward` state machine event
- [x] fix unoptimized nexted loops

iOS:
- [x] UI for list of derived keys
- [x] UI for error states
- [x] full logic flow for import
- [x] logic flow for recover key set
- [x] logic flow for add missing network
- [x] logic flow for recover key set
- [x] logic flow for add missing network
- [x] fix issues on data refresh on Key Set List

## Screenshots
| UI states | Error States |
|-|-|
|||
| ||
|||
|||
|| |
| |

## Technical Notes
After scanning import derived keys QR code, we always need to fake navigation with `.goBack` action, as even if camera is dismissed without import, navigation "forward" already happened.


krodak:
@prybalko can we make `hasPwd` non-optional? Is there any situation we don't know whether given key has password? 
CleanShot 2022-12-23 at 11 12 03@2x

Also, in which situation `derivationPath` would be `nil`?

prybalko:
@krodak good questions.

`hasPwd` is optional because it's impossible to know if path has a password based purely on QR import payload. The mobile app is responsible for calling the `populate_derivations_has_pwd` function with the seeds in order to determine the value of `hasPwd` for a given derivation path. If the value of `hasPwd` is `nil` after this function is called, it means that there is no seed associated with the derivation path and the import should be ignored.

`derivationPath` is `nil` for the root key.

rust/definitions/src/qr_transfers.rs
```diff
@@ -1,31 +1,30 @@
...
 //! through the QR codes maintaining the air-gap. In addition to transactions
 //! that could be parsed and signed, this includes:
-//! - adding new networks,  
-//! - updating the metadata of existing networks,  
-//! - updating the types information,  
-//! - bulk-importing the derivations  
+//! - adding new networks,
+//! - updating the metadata of existing networks,
+//! - updating the types information,
+//! - bulk-importing the derivations
```
montekki, line 9: It would be nice for reviewers not to make a huge patch even larger by including unrelated whitespace changes into it.

rust/navigator/src/actions.rs
```diff
@@ -24,6 +24,7 @@ pub enum Action {
     NetworkSelector,
     NextUnit,
     PreviousUnit,
+    ChangeNetwork,
```
montekki, line 27: That is probably a bad merge; was removed in previous PRs

rust/navigator/src/navstate.rs
```diff
@@ -1786,6 +1763,7 @@ impl State {
                 Action::NetworkSelector => self.handle_network_selector(),
                 Action::NextUnit => self.handle_next_unit(),
                 Action::PreviousUnit => self.handle_previous_unit(),
+                Action::ChangeNetwork => self.handle_change_network(details_str),
```
montekki, line 1766: This was removed

rust/navigator/src/navstate.rs
```diff
@@ -817,6 +776,28 @@ impl State {
...
+                    Err(e) => {
+                        new_navstate.alert = Alert::Error;
+                        let _ = write!(&mut errorline, "{}", e);
+                    }
+                }
+            }
+        }
+        (new_navstate, errorline)
+    }
+
```
montekki, line 800: This was removed

rust/signer/src/signer.udl
```diff
@@ -240,6 +240,7 @@ enum Action {
     "NetworkSelector",
     "NextUnit",
     "PreviousUnit",
+    "ChangeNetwork", // used for tests
```
montekki, line 243: Should be removed.

montekki (changes requested):

rust/transaction_parsing/src/derivations.rs
```diff
@@ -1,44 +1,129 @@
...
+fn prepare_derivations_v1<P>(db_path: P, export_info: ExportAddrsV1) -> Result<Vec<SeedKeysPreview>>
+where
+    P: AsRef<Path>,
+{
+    let mut result = Vec::new();
+    for seed_info in export_info.addrs {
+        let mut derived_keys = vec![];
+        for addr_info in seed_info.derived_keys {
+            let multisigner =
+                base58_or_eth_to_multisigner(&addr_info.address, &addr_info.encryption)?;
```
montekki, line 56: And if you have just one eth addr in the pack it will nuke the whole process here.

montekki: This should not be done at all, the correct way would be something like

1. Arrive at `MultiSigner` value with `seed phrase` + `AddrInfo.derivation_path`;
2. Use `AddrInfo.genesis_hash` + `AddrInfo.encryption` to query the network info and get ss58prefix
3. if needed get the address and compare it to `addr_info.address` for correctness.

tldr: `address` field is only intended for frontends such as Nova or Polkadot-JS to use to do their thing they do where addresses and not keys matter.

rust/db_handling/src/identities.rs
```diff
@@ -290,32 +321,65 @@ fn import_all_addrs_v1<P: AsRef<Path>>(
...
+                MultiSigner::Sr25519(p) => sr25519_signers.get(&p),
+                MultiSigner::Ed25519(p) => ed25519_signers.get(&p),
+                MultiSigner::Ecdsa(p) => ecdsa_signers.get(&p),
+            };
+            if seed_name.is_none() {
+                // unknown seed, skipping
+                continue;
             }
+            let seed_name = seed_name.unwrap();
+            let seed_phrase = seeds.get(seed_name.as_str()).unwrap();
```
montekki, line 337: Please avoid `.unwrap()` in prod code

rust/db_handling/src/identities.rs
```diff
@@ -290,32 +321,65 @@ fn import_all_addrs_v1<P: AsRef<Path>>(
...
+                        Err(e) => {
+                            full_address.zeroize();
+                            return Err(Error::SecretStringError(e));
+                        }
+                    }
+                }
+            };
+            let multisigner =
+                base58_or_eth_to_multisigner(&derived_key.address, &derived_key.encryption)?;
+            derived_key.has_pwd = Some(multisigner_pwdless != multisigner);
```
montekki, line 379: so essentially here you have to go from multisigner_pwdless to address and compare it to `derived_key.address`, not the other way around. This way you in case of Ethereum will go from ecdsa multisigner -> to 20-byte hex address.

rust/navigator/src/actions.rs
prybalko: This is intentional.
The previous implementation of the derivation import changed the network navstate, but the new one does not. However, I need a way to mimic the old behavior for `flow_test_1`, so `ChangeNetwork` is very useful for this purpose.

rust/transaction_parsing/src/derivations.rs
prybalko: > 1. Arrive at MultiSigner value with seed phrase + AddrInfo.derivation_path;

This woudn't work for passworded paths.

prybalko: Ok, but it woudn't work for passworded paths.
The only ways to get a `MultiSigner` for a passworded path are:
1. ask user for a password
2. docode `address`

am I missing anything?

rust/definitions/src/helpers.rs
```diff
@@ -230,6 +230,26 @@ fn print_ethereum_address(public: &ecdsa::Public) -> String {
...
+            let pubkey = sr25519::Public::from_ss58check(base58_or_eth)?;
+            Ok(MultiSigner::Sr25519(pubkey))
+        }
+        Encryption::Ethereum | Encryption::Ecdsa => {
+            let pubkey = ecdsa::Public::from_ss58check(base58_or_eth)?;
+            Ok(MultiSigner::Ecdsa(pubkey))
+        }
+    }
+}
+
```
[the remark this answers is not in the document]
prybalko: Thanks for the pointer, fixed

rust/transaction_parsing/src/derivations.rs
prybalko: Ok, lets use `seed phrase + AddrInfo.derivation_path` for passwordless paths.

rust/db_handling/src/identities.rs
prybalko: Fixed

prybalko:
I made quite a few changes to fix the pointed flaws in busness logic as well as in data structures for frontend. 
Now the flow looks like this:
1. user scans the QR code, UI gets `ExportAddrsContainer` data structure with raw data.
2. call `get_derivations_preview` with given `ExportAddrsContainer` and all available seeds. Get `SeedKeysPreviewSummary` with with separated `importable_keys`, `already_existing_keys` and `non_importable_keys` fields. 
3. call `import_derivations` with the same  `SeedKeysPreviewSummary`

##### ToDo:
- [ ] fix tests

montekki:
@prybalko I've reverted your patch; keep it stashed somewhere. @krodak and I have decided that the frontend structures should not be changed at least before the RC; I will try to do the minimal fixes w.r.t. pre-patch code.

krodak:
@Dmitry-Borodin could we fix android build here to merge?

android/src/main/java/io/parity/signer/screens/scan/transaction/transactionElements/TCDerivatoins.kt
```diff
@@ -1,47 +0,0 @@
-package io.parity.signer.screens.scan.transaction.transactionElements
```
Dmitry-Borodin, line 1: @montekki  is that intentional that you removed this file and restored old one?

montekki (approved):

PR #1441, opened 2022-11-25
feat: More typed errors hotfix

montekki:
## Purpose

Make a small subset of errors typed for UI to better present them to the user.


Dmitry-Borodin:
Would like to have some enumeration of recoverable errors and clear separation between Error - user should do something and Exceptions like MutexPoisoned.
Another option is to mark specific function calls with recoverable errors only this function can cause.

krodak:
Closing this one as #1621 addresses this already

PR #1445, opened 2022-11-28
fix: android bulk signing seed passed

Dmitry-Borodin:
## Purpose
passe all seeds for signing multitransactions.

FYI @montekki this will pass the same seed multiple times if it's the same for multiple transactions. Is it ok?


montekki:
> FYI @montekki this will pass the same seed multiple times if it's the same for multiple transactions. Is it ok?

Exactly

But all the rest has also to be tested & implemented if it is not yet

Dmitry-Borodin:
@montekki  implemented full old flow and tested. I think this is ready to merge and I'll start working on new designs for those screens right away

montekki (approved):

PR #1458, opened 2022-12-08
feat: remove qr generation from rust

montekki:
## Purpose

Do not generate QR images on backend side.


krodak:
@Dmitry-Borodin could you update Android app to always generate QR codes on Android side and update to `QrCode` data model?

Dmitry-Borodin:
There is no simple solution on android to generate QR code. It is easier to use rust util function otherwise I"ll need to reimplement the same in kotlin, which doesn't make sense. 
I think it's great thing to not have qr generation as part of business logic requests, but we can keep it as external utils function as it does the job well.

montekki:
> I think it's great thing to not have qr generation as part of business logic requests, but we can keep it as external utils function as it does the job well.

I am pretty sure there should be some 3-d party library that can be used just as it is in case of iOS. Covid pandemic alone has created thousands of apps that generate or read QR codes one way or another, so it is hard for me to believe that this has to be written from scratch.

Dmitry-Borodin:
> > I think it's great thing to not have qr generation as part of business logic requests, but we can keep it as external utils function as it does the job well.
> 
> I am pretty sure there should be some 3-d party library that can be used just as it is in case of iOS. Covid pandemic alone has created thousands of apps that generate or read QR codes one way or another, so it is hard for me to believe that this has to be written from scratch.

Yes there are jhava libraries mostly quite old and big, but not something like in iOS where platform API have just single call to generate it. 
I will need to evaluate those libraries, think if they are capable to do what we want (tune colors, add images et.c.) and it is not something I can do in an hour. I think we can postpone removing this utils function from rust untill we will need this new logic on android and priority to implement it. I don't think this is even release blocker. 
It is probably few days task to do.

rust/definitions/src/navigation.rs
```diff
@@ -430,6 +430,35 @@ pub enum MSCContent {
     AddSpecs { f: MSCNetworkInfo },
 }
 
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum QrData {
```
Dmitry-Borodin, line 434: I"m not sure that backend should decide which qr code is sensitive it it's model. 
I will ignore this information on android and will always make qr code in private key screen look sensitive and in key details - normal.

Dmitry-Borodin:
#1466 my changes

rust/definitions/src/navigation.rs
krodak: I think it's fine, as this is like property in database model telling you property of a given object, it does not tell you how to display it.

PR #1476, opened 2022-12-14
feat: update workflows to limit triggers

krodak:
## Purpose
Limit triggers for CI workflows for given actions

## TODO
- [x] Confirm that we don't need to disable `Required` in repo Settings to not requite CI workflow that is disabled for given path


prybalko:
As you can see, now both "skip" and "normal" CI jobs run. That's because `paths` and `paths-ignore` works differently.
 
> When *all* the path names match patterns in paths-ignore, the workflow will not run

> If at least *one* path matches a pattern in the paths filter, the workflow runs. 

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths

I'm not sure how to solve this elegantly.

krodak:
@prybalko @montekki I know it looks weird now, but I think this is what we are looking for.
For each workflow we are only using `paths` so workflow will get triggered for this specific paths only
- each "real" workflow includes relevant paths
- each "skip" workflow includes relevant paths with `!` prefix, as [explained here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-and-excluding-paths)
 - each workflow now also has its workflow file listed - as any change within workflow, should trigger check.
    - this additionally shields us from one side effect of `!` prefix explained in docs: `If you define a path with the ! character, you must also define at least one path without the ! character. If you only want to exclude paths, use paths-ignore instead.` - hence, if we have `!rust/**` and `'.github/workflows/android-app-skip.yml'` now `!` will work correctly
    
 As last note - yeah, I know for this PR **all workflows are triggered** - but this is to be expected as we list workflows paths in `paths` in every workflow file

montekki (approved):

PR #1494, opened 2022-12-23
feat: iOS: Network Details

krodak:
## Purpose
This PR updates `Network Details` screen and adds an ability to add metadata from that screen.
Some screens available from this screen - `Sign Sufficient Crypto` (signing metadata with key set), `Network Details Menu` still need to be redesigned

## Scope
- New screen for `Network Details`
- Updated flow and design for `Remove Metadata` interaction
- Ability to add new metadata directly from this screen - one caveat is that we need to get back to main `Settings` screen on camera dismissal, this is still due to navigation being handled in Rust. **If metadata is successfully scanned user will end up properly on Network Details with new metadata visible**

## Discussion
Due to navigation being in Rust, when camera is dismissed we need to end up on `Settings` screen. I don't want to add conditional fake navigation back to `Manage Networks` when dismissing camera as this would add unnecessary complications and this can be improved when we rely less on Rust navigation.

## Screenshots

| Light | Dark |
|-|-|
|||
|||

## Recording
| Camera scan | Metadata removal | Sign metadata |
|-|-|-|
||||


montekki (approved):
👨‍🍳 👌

tetekinandrey (approved):

PR #1513, opened 2023-01-04
feat: target android NDK api we are actually using

Dmitry-Borodin:
no need to target old one


Dmitry-Borodin:
FYI tested on device SDK api 23. It should be NDK api as in documentation

montekki (approved):

PR #1522, opened 2023-01-07
feat: derive key in network

montekki:
## Purpose
Derive new keys with redesigned flow and `Allow use on any network` capabilities

## Scope
Rust:
- `get_all_networks` method to fetch all networks
- `try_create_address` to create derived key

iOS:
- new UI flow for Create Derived Key
- dedicated screen for Derivation Path Name
- 3 info modals to give information on derivation methods, paths, derived keys themselves
- network selection UI with `Allow use on any network` option
- Rust API integration
- update to Key Set Details to use `get_all_networks` call

## TODOs
- [ ] Resolve issues with `Allow use on any network`

### Create Derived Key
| UI | UI |
|-|-|
|||
|||
|| | 

### Create Derivation Path
| UI | UI |
|-|-|
|||
|||
|||
|||

### Confirmation
| UI | UI |
|-|-|
|||
|||


rust/signer/src/lib.rs
```diff
@@ -280,6 +280,21 @@ fn encode_to_qr(payload: &[u8], is_danger: bool) -> anyhow::Result<Vec<u8>, Stri
     qrcode_static::png_qr(payload, sensitivity).map_err(|e| format!("{}", e))
 }
 
+/// Get all networks registered within this device
+fn get_all_networks(dbname: &str) -> anyhow::Result<Vec<MSCNetworkInfo>, String> {
```
Dmitry-Borodin, line 284: We have very similar logic in rust/db_handling/src/interface_signer.rs:396
Shouldn't we reuse it?

Dmitry-Borodin (approved):

krodak:
I'll be replacing `Network` within Key Set Details with data got from new endpoint, so we could also remove `Network` as data structure as its no longer needed @Dmitry-Borodin @montekki

PR #1544, opened 2023-01-16
feat: key address correctly

montekki:
## Purpose

Address keys have to have correct keys that reflect toe logic of single key-in-network usage.
Currently a network prefix is added to the keying material, however, for collision resistance
full network info such as genesis has to be added.


montekki:
@Dmitry-Borodin android build fix for this branch would be great

krodak (approved):

Dmitry-Borodin (approved):

PR #1562, opened 2023-01-20
feat: use a single db handle

montekki:
## Purpose

Fix racing on sleds:

> sled does not support multiple open instances for the time being. Please keep sled open for the duration of your process's lifespan. It's totally safe and often quite convenient to use a global lazy_static sled instance, modulo the normal global variable trade-offs. Every operation is threadsafe, and most are implemented under the hood with lock-free algorithms that avoid blocking in hot paths.


prybalko:
I wonder how this is going to work for tests. If there could be only 1 Db instance, does this mean tests will run ~sequentially?

montekki:
If you have only one db instance tests would be broken anyway since you would have races like
test_1 reads A
test_1 writes A
test_2 reads A
test_2 writes A

and the states they could see would be all permutations of the db states.

each state has to have a unique db for the run in tempdir at least if this test writes to db. since even the same test can be launched twice or more times in parallel.

montekki:
But you are also right, currently tests are also racy.

montekki:
@prybalko my explanation was not clear I think, what the doc states is obviously that it is not safe to have multiple open instances _of the same_ db:

```rust
sled::open("foo").unwrap();
sled::open("foo").unwrap();
```

It should be fine to have single instances to multiple different locations.

montekki:
@krodak @Dmitry-Borodin there are simple api changes in this patch; it would be nice to have the UI fixes. One non-trivial thing is that `init_navigation` with a path to DB has to be changed _before_ any calls to backend because they may relay on internal global db context that is initialized in `init_navigation` currently.

krodak (approved):

PR #1579, opened 2023-01-25
feat: allow empty derivations to expose root key on some network

montekki:
Usage of a root key on some network has to be allowed via creating an ampty
derivation in a certain network.


krodak (approved):

krodak:
@Dmitry-Borodin might be relevant to your work on #1569

PR #1582, opened 2023-01-25
feat: expose root key by default

montekki:
## Purpose

It would make sense to expose the root key on each known network instead
of creating derivations that may not even be used.

### DIFF
--- a/rust/db_handling/src/identities.rs
+++ b/rust/db_handling/src/identities.rs
@@ -891,7 +891,7 @@ fn populate_addresses(
         if let Ok(prep_data) = create_address(
             database,
             &address_prep,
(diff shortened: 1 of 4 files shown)


rust/db_handling/src/identities.rs
```diff
@@ -891,7 +891,7 @@ fn populate_addresses(
         if let Ok(prep_data) = create_address(
             database,
             &address_prep,
-            &network_specs.specs.path_id,
+            "",
```
prybalko, line 894: Please consider adding a test for this change

montekki: believe me I am changing the whole test base atm.

prybalko (approved):

PR #1596, opened 2023-01-30
fix: reduce frame rate of apng qr codes

andresilva:
I have been using an iPod as a signer device and I consistently have issues with importing new metadata since the QR code reader constantly loses track. I have to reposition the device so that it refocuses and starts tracking again, it usually takes me more than 3 minutes to get the metadata to update. I suspect this might be due to the low performance of this device and hence people using more powerful devices might not have this problem. Reducing the frame rate of the generated QR APNG completely solved this issue for me, the device does not lose track of progress and I'm able to update the metadata in just a couple of seconds.


andresilva:
It might be possible to not reduce the frame rate as much, I have only tested this value and it worked perfectly. I can test higher values if you feel that's appropriate.

andresilva:
I tested 20 and 25 FPS as well, and at least on the iPod 15 FPS is still the fastest to parse.

montekki:
@andresilva hi, thanks a lot for the patch and the testing! We are going to test the new builds on hardware such as iPod shortly with this patch

montekki:
@prybalko if we merge this how fast can the changes land on the portals?

prybalko:
@montekki a couple of days, but it'll affect only new metadata QRs

montekki (approved):

PR #1614, opened 2023-02-07
feat: Android empty screens state

Dmitry-Borodin:
## Purpose
- removed create derivation button and added header icon
- added empty state text for 2 screens
- updated viewmodel parameters to avoid activity leak and usage of old activity if it's recreated but old viewmodel is alive.

## Screenshots
| Left title, i.e. "Before" | Right title, i.e. "After" |
|-|-|
| black_details | white_details |
| keyset_black | keyset_white |

PR #1618, opened 2023-02-07
feat: don't show create bottom sheet for key sets if nobody requested it

Dmitry-Borodin:
We now have empty state there on a platform level
@krodak you have similar state to this one right? #1614

So we don't need to show bottom sheet by default anymore as per design.


krodak:
@Dmitry-Borodin for me there is no popup or anything - if you delete any derived key, you should display Key Set List - this also works for last key
When you go to Key Set Details with no keys - no popup is present on iOS.

Not sure we really need to change this just before release...

Dmitry-Borodin:
@krodak this only removes functionality of adding bottom sheet when key set list is requested, and no keys are there.
No extra features or something.

krodak (approved):

prybalko (approved):
👍

PR #1627, opened 2023-02-10
feat: do not ship metadata in release

montekki:
## Purpose

There is no point in shipping default metadata blobs with the Signer release.
Even if the latest blobs are published at the release time they will be outdated
before the next release which will force all users to go through the metadata
update process as a first action after installing Signer anyway.


Dmitry-Borodin:
This may affect testing builds as we can't update metadata on emulators via qr codes.

montekki:
Well you cannot scan txs via qr codes either then

Dmitry-Borodin:
> Well you cannot scan txs via qr codes either then

You can export and import from device with camera, this is how I test derivations now for example.
I agree we don't need it for release, but would be nice to still have default metadata for debug builds.

montekki:
You do not need metadata to create keys, you only need the network specs; specs remain.
You can suggest a flag to be passed at build time to `make-cold-release` step of the build and we can distinguish between which data to include. For iOS we have two bulid configs now: normal and `-Dev` one that among other things disables a strict airgap requirement or something

krodak (approved):

PR #1661, opened 2023-02-22
fix: do not gobble up errors in bulks

montekki:
## Purpose

If one TX in the bulk is faulty or is not a Tranasction Signing action,
return just it.


prybalko (commented):
I'm not following. If there are 100 transactions in a bulk and some of them are faulty, should we only return the first faulty one?

montekki:
You are either good to sign ALL or app bails out. It should not happen so that the app silently errors on 1 tx out of 100, and gives you 99 signatures for those correct ones. That behavior could be really dangerous.

prybalko:
Would make sense to show _all_ faulty transactions, not only the first one?

montekki:
That is not that easy to do interface-wise and I would prefer to do a hotfix now and polish it after the release

prybalko (approved):

Issue #1762, opened 2023-03-31
android: QR parsing is slow


prybalko:
QR parsing on Android is extremely slow. 

Here is a video comparing Pixel 4a with iPod touch:

https://user-images.githubusercontent.com/7165377/229068965-4bb118db-c1e7-4a46-9f27-05b3feed5d0f.MOV

Tbaut:
I have installed one of last week's master version on a Pixel 3a and was surprised how slow it was compared to before (version 5). I meant to test on a hot device to see if there was a noticeable change but this issue confirm this. I certainly remember needing less than a minute for 300+ frames. Now it's much more.

Zerophase:
Also happens with PIxel 4a (5g). Don't believe it scans anything. The screen turns off before the scan finishes.

Dmitry-Borodin:
Thank you for contribution. Please follow up in existing issue for this topic #1436
Framing changed in last/this week, please let us know if it still much slower than was before. You can check on master on recently released version.

Zerophase:
Last release I see was from almost a year ago. Are we close to having a new release?

Tbaut:
@Dmitry-Borodin I've tried with the latest build to date from maste, the frame position isn't the problem. I'm comparing v5 and v6, the difference is like day and night. On a pixel 4a, it's at least 2 or 3 times faster on v5. I changed my signer device because on v5 my old phone was too slow. Now I feel like I'm back to square one unfortunately.

The v6 camera is zoomed where the v5 is not, maybe that's something to take a look at, where zooming is loosing in resolution?

Dmitry-Borodin:
@Zerophase it was released end of last week.

@Tbaut Yes, you are right, it is zoomed because with new UI we need to show camera data for full screen vertically, it is done by cropping a lot from original camera image (mostly from sides).
Barcode scanner component scanning original images from camera and doesn't support only selected image at the moment.
I'm thinking of implementing it with manual cropping images before processing if other solution won't be found. But it's not the case at the moment.
Could you check if zooming in QR code on a screen helps? I've reproduced decrease of QR recognition performance (still usable through) when I open portal in small screen of my laptop and under weak light conditions. And it is solved by increasing browser scale to make QR code on a screen bigger, so I can keep device more far away from computer screen.

So 2 simple ways to check if you mind:
 - zooming in qr code on computer by browser scale.
[...]

PR #1764, opened 2023-04-04
fix: Android scan improvement

Dmitry-Borodin:
## Purpose
Improve scanning performance on some devices where it's very poor

## Scope
- Required resolution is changed - default is 640*480 and it's not enough
- analyzing in background thread as big images taking a lot of CPU to analyze
- paddings for frame increased so users won't keep phone too close to the screen. 40.dp as it was before

This targeting to improve #1762
Feedback on scanning on different devices is very welcome
Build with those changes to download - https://drive.google.com/file/d/17YWhZIZgY8JM6HzsYAFxe2tUzKlCGWFj/view?usp=share_link


android/src/main/java/io/parity/signer/screens/scan/camera/CameraView.kt
```diff
@@ -27,6 +32,10 @@ internal fun CameraViewInternal(viewModel: CameraViewModel) {
 	val cameraProviderFuture =
 		remember { ProcessCameraProvider.getInstance(context) }
 	val coroutineScope = rememberCoroutineScope()
+	val configuration = LocalConfiguration.current
+val backgroundExecutor = remember {
```
krodak, line 36:
```suggestion
    val backgroundExecutor = remember {
```
But seems like something that linter / formatter should catch

android/src/main/java/io/parity/signer/screens/scan/camera/CameraView.kt
```diff
@@ -57,19 +68,32 @@ internal fun CameraViewInternal(viewModel: CameraViewModel) {
...
+					.setOutputImageFormat(ImageAnalysis.OUTPUT_IMAGE_FORMAT_YUV_420_888)
+					.setTargetResolution(Size(1920,1080))
 					.build()
 					.apply {
 						setAnalyzer(executor) { imageProxy ->
 							viewModel.processFrame(barcodeScanner, imageProxy)
 						}
 					}
 
+				Log.e("rational is ", "${configuration.screenWidthDp} ${configuration.screenHeightDp}")
```
krodak, line 80: We probably shouldn't leave that log here

krodak (approved):

android/src/main/java/io/parity/signer/screens/scan/camera/CameraView.kt
Dmitry-Borodin: fixed

android/src/main/java/io/parity/signer/screens/scan/camera/CameraView.kt
Dmitry-Borodin: we don't have automated formatter, formatted.

android/src/main/java/io/parity/signer/screens/scan/camera/CameraViewModel.kt
```diff
@@ -44,6 +46,7 @@ class CameraViewModel() : ViewModel() {
 		barcodeScanner: BarcodeScanner,
 		imageProxy: ImageProxy
 	) {
+
```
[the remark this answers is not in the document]
Dmitry-Borodin: Just semantic separation of different sections. I think this is ok.
This part was changed few times during changes and will probably be changed in future optimisations

Thanks for review.

Tbaut:
Thanks, this fixes the issue mentioned in #1762 
I've had the best results with the phone actually pretty far from the screen, something like this

sign

PR #1768, opened 2023-04-06
feat: updated version for release

Dmitry-Borodin:
## Purpose
update version for hotfix release with scanning improved

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -12,8 +12,8 @@ android {
 		applicationId "io.parity.signer"
 		minSdk 23
 		targetSdk 33
-		versionCode 60000
-		versionName "6.0.0"
+		versionCode 60100
+		versionName "6.1.0"
 		ndk {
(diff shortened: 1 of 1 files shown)


Dmitry-Borodin:
> Q: If this includes more of the bugfixes, then new functionality, shouldn't we go with 6.0.1? At least for iOS I've assumed: -> single screen design updates -> increase patch version -> bug fixes for existing features -> increase patch version -> internal improvements of existing features -> increase patch version
> 
> -> rollout of new feature -> increase minor version -> major extension of existing feature -> increase minor version
> 
> -> ??? probably another paradigm change or redesign -> increase major version

I assume that 6.0.1 is only bugfixes backported to old release branch so you are safe to update et.c.
This is new master with behavioral changes in both camera (different scanning focus because of different serolution) and other features - some redesign
As I cannot guarantee it's fully backwards compatible I increase semantic version second number.

krodak:
I guess as long as we haven’t updated Rust API integration that would change models and thus affect database it is compatible but ok

krodak (approved):

PR #1779, opened 2023-04-13
feat: Android release process

Dmitry-Borodin:
## Purpose
Make release flow in github action instead of manual process in clean room.

## Scope
- build and sign APK in github
- upload release apk to internal track in play store
- readme updated of how we can release now.

Feedback is welcome.


montekki (approved):

.github/workflows/android-app.yml
```diff
@@ -7,8 +7,8 @@ on:
       - master
       - stable
 jobs:
-  build:
-    name: Build
+  tests:
+    name: Build android unit tests
```
krodak, line 11: You need to revert to `Build` for name as this is required check for PR merge

krodak (changes requested):
I think workflow trigger might be wrong and would trigger release on every push to main

krodak:
My suggestion would be to align with iOS release pipeline and switch to manual trigger via CI:
```
on:
 workflow_dispatch:
```

.github/workflows/android-release.yml
```diff
@@ -0,0 +1,171 @@
+name: Release android app
+
+on:
+  push:
+    branches:
+      - master
```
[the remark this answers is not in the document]
krodak: Here is additional extended command that suggest that indeed this is "OR" operation between branch name and tag. It also goes through suggestions on how to apply "AND" between those rules, using `if` triggers: https://stackoverflow.com/a/66463380

.github/workflows/android-release.yml
```diff
@@ -0,0 +1,171 @@
...
+            ${{ runner.os }}-gradle-
+
+      # Build and publish
+
+      - name:           Assemble release unsigned apk
+        run:            |
+          ./gradlew clean assembleRelease --stacktrace
+
+      - name: Sign app APK
+        uses: r0adkll/sign-android-release@v1
```
prybalko, line 145: This repository seems abandoned. Could we use another one?
https://github.com/r0adkll/sign-android-release

.github/workflows/android-release.yml
```diff
@@ -0,0 +1,171 @@
...
+          access_token: ${{ github.token }}
+
+      - name:           Checkout Sources
+        uses:           actions/checkout@v3.5.0
+        with:
+          fetch-depth:  50
+          submodules:   'recursive'
+
+      - name:           Install System Dependencies (Linux)
+        if: runner.os == 'Linux'
```
prybalko, line 92: What else could it be? Line 77 says `runs-on: ubuntu-22.04`

.github/workflows/android-release.yml
```diff
@@ -0,0 +1,171 @@
+name: Release android app
+
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - 'v*'
+jobs:
+  tests:
```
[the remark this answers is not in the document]
Dmitry-Borodin: I would keep it just to see just as a precaution that all tests passed for release task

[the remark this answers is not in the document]
krodak: Its pointless as with iOS release, as any code that is merged is going through that other workflow.
So you’d be calling same action twice on same code, thus increasing CI usage for no good reason

.github/workflows/android-release.yml
prybalko: It uses deprecated Node12, so one day it'll stop working this summer
https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/
https://github.com/r0adkll/sign-android-release/pull/70

.github/workflows/android-release.yml
Dmitry-Borodin: Good point, this part if copied from existing android flow to run tests. Removed.

Dmitry-Borodin:
@krodak fixed issues you requested to change. Thx for review.

krodak (approved):
LGTM

.github/workflows/android-app.yml
krodak: Judging by PR status I think it’s still outstanding @Dmitry-Borodin

PR #1800, opened 2023-04-23
feat: ios: native navigation key set

krodak:
## Purpose
This PR aims to deliver native navigation for Key Set tab and all subscreens

## Scope
- create services for Key Set and subsequent screens
- add `NavigationView` / `NavigationLink` to support native navigation
- replace all navigation calls by services call with fake navigation
- retest all functionalities

## TODOs
- [x] create services for Key Set and subsequent screens
- [x] add `NavigationView` / `NavigationLink` to support native navigation
- [x] replace all navigation calls by services call with fake navigation
- [x] retest all functionalities


prybalko (approved):
nice 👍

PR #1830, opened 2023-05-14
chore: do not create root network keys by default

prybalko:
## Purpose

The idea is not to create any keys automatically. When the user creates a new KeySet, it will be empty. The root keys can be added manually afterwards.

## Discussion

Any thoughts on this?


krodak (approved):

Dmitry-Borodin:
So I think we'll need new UI part to add those in platform level, right?
DO we have final designes for those? Just want to make sure we'll keep track of it and end up with clear user experience.

krodak:
@Dmitry-Borodin I think once we have designs from @tetekinandrey we'll do those changes on this branch and only merge then

Dmitry-Borodin:
Sounds good, converted to draft to avoid accidental merge

prybalko:
Won't do.

prybalko:
> @prybalko this will not create networks for restore keyset as well, right?

Correct

prybalko:
I added `create_key_set` and `print_new_seed` APIs. They might be useful for KeySet creation

Dmitry-Borodin:
@krodak this breaks master as no keys are created now, we just discussed that we will merge this together with mobile changes...

krodak:
@Dmitry-Borodin nothing is broken as user can create keys individually still. 
I think we've discussed that this is ok to not merge whole functionality in one PR as you wanted smaller PRs and we don't make any build from that - app is working and you can create key set, you just need to create keys manually

krodak:
And now each platform can make separate PR with changes.

PR #1836, opened 2023-05-17
feat: ios: generate xcframework instead of fat .a library

krodak:
## Purpose
This PR attempts to update Rust dependencies flow for iOS to use XCFramework instead of FAT .a library


krodak:
Closing as it's still unsupported by uniffi

PR #1838, opened 2023-05-19
feat: ios: Prepare SPM plugin to use uniffi and encapsulate Rust code in SPM

krodak:
## Purpose
This PR tries to prepare Signer-specific (at first) SPM plugin that would use `uniffi-bindgen` to generate Swift code in dedicated SPM package to keep Rust-generated code separated from main application module, 

## Scope
- SPM `buildTools` plugin to generate code using uniffi
- SPM `commandLine` plugin to generate code using uniffi
- SPM package `Signer` that uses `buildTools` plugin to generate Swift code from Rust on every build
- [XCFramework generation instead of FAT library as dependency](https://github.com/paritytech/parity-signer/pull/1836) - will revisit whether .a can work as well 

## TODOs
- [x] SPM `buildTools` plugin to generate code using uniffi
- [x] SPM `commandLine` plugin to generate code using uniffi
- [x] SPM package `Signer` that uses `buildTools` plugin to generate Swift code from Rust on every build
- [ ] Resolve issue where `signerFFI.h` header is not visible properly for `signer.swift` within Signer SPM package
- [ ] When issue with headers is resolved, refactor SPM buildTools plugin to make it universal
- [ ] Post PR to [uniffi-rs](https://github.com/mozilla/uniffi-rs) with SPM build tools plugin as viable alternative to custom build rules to integrate generated code in iOS project

## Discussion
Main issue right now is to properly expose headers generated by `uniffi` to be visible within .swift codebase. There are potentially 2 areas to look for issue: 
- make better work of integrating headers when generating xcframework - and maybe even include `signer.swift` within xcframework itself (ideal solution)
- improve `Package.swift` to properly locate headers


krodak:
Closing as #1836 is still not possible

Issue #1846, opened 2023-05-23
Trying to set up Polkadot Vault with a custom Viewer Certificate can lead to irrecoverable crash


alecdwm:
Hey! I'm going to list a few scenarios for a user, one of them leads to a crash which requires the user to delete and re-install the app from scratch.

I'm hoping this can be fixed!

Scenario 1 (no crash):
1. Set up fresh vault
2. Remove viewer certificate
3. Scan chainspec with new viewer certificate
4. Add key set and carry on

Scenario 2 (crash):
1. Set up fresh vault
2. Remove viewer certificate
3. Add key set
4. Scan chainspec with new viewer certificate
5. Close QR scanner -> app crashes
6. Open app, enter passcode -> app crashes

Scenario 3 (no crash):
1. Set up fresh vault
2. Remove viewer certificate
3. Add key set
4. Remove the three default `derived keys` (with blank derivation paths) from the key set (Polkadot, Kusama, Westend)
5. Scan chainspec with new viewer certicate
6. Close QR scanner and carry on

alecdwm:
I'm guessing that the three root accounts created by default are somehow set up with the default viewer certificate.
Perhaps the fix is to not create them if it has been removed?

krodak:
@alecdwm thanks for detailed description on the issue 🙇🏻‍♂️ 
Pushed changes that will fix error presentation and avoid crash on iOS app 🙏🏻 
@prybalko currently to fetch list of key sets, we are making navigation call to `.start` and then `.navbarKeys` (as currently with native navigation, each `Service` that interacts with backend through navigation state, assumes start from `.start` state), we are getting following error on every call to `.start` or consequent calls that tries to transition to different state:

So now @alecdwm if you run this from Xcode you should see following:

![IMG_6688](https://github.com/paritytech/parity-signer/assets/1955364/cd9f5afd-cda9-4c2c-81b5-2170ea0f839b)

``` 
▿ Optional<ActionResult>
  ▿ some : ActionResult
    - screenLabel : "Select seed"
    - back : false
    - footer : true
    ▿ footerButton : Optional<FooterButton>
      - some : PolkadotVault.FooterButton.keys
    ▿ rightButton : Optional<RightButton>
      - some : PolkadotVault.RightButton.newSeed
    - screenNameType : PolkadotVault.ScreenNameType.h1
    ▿ screenData : ScreenData
      ▿ settings : 1 element
        ▿ f : MSettings
          - publicKey : nil
[...]

PR #1847, opened 2023-05-24
fix: #1846 - Handle removed verifier certificate when there are existing key sets state properly

krodak:
## Purpose
This PR aims to address issue: #1846

## Scope
- avoid crash when not receiving `MSeeds`
- fix navigation error state rendering for iOS

- remove default networks when general verifier is removed


prybalko:
@krodak I've pushed the fix to the backend. It removes all default networks and metadata when the general verifier is removed.

prybalko (approved):

prybalko:
Is this PR still relevant?

krodak:
@prybalko I'll close it down and realign when new design for this screen is ready if there is no verifier cert

PR #1879, opened 2023-06-08
feat: Android dont create root keys for all networks adjustements

Dmitry-Borodin:
## Purpose
create proper path keys only for networks use is selected when keyset is created or restored

## Screenshots
| new white | new black |
|-|-|
| image | image |
| image | image |
| image | image |
| image | image |
|  |  |
| old black | old white |
|-|-|
| dark | light |
| Screenshot_20230608_173628 |  |
|  |  |


android/src/main/java/io/parity/signer/domain/Models.kt
```diff
@@ -263,6 +263,9 @@ data class NetworkModel(
 	}
 }
 
+fun NetworkModel.createDefaultPath(): String = "//${this.title}"
```
Dmitry-Borodin, line 266: @krodak are you using the same logic on iOS?

Dmitry-Borodin:
I will reopen this when will use new rust functions instead of navigation or just new logic for default network path

android/src/main/java/io/parity/signer/domain/Models.kt
krodak:
```swift
 let pathAndNetworks: [(path: String, network: MmNetwork)] = networks
            .map { (path: "//\($0.title)", network: $0) }
```

prybalko: `MNetworkDetails` model as a `path_id` and that's exactly the derivation path prefix.

tetekinandrey (approved):

PR #1903, opened 2023-06-23
feat: ios: default derived key name

krodak:
## Purpose
This PR adds default derived key name strategy and removes "Allowed on any network" option when creating key

## Screenshots

| Primary derived key | "Numeric" derived key |
|-|-|
|||


prybalko (approved):
Nice work 👍

PR #1924, opened 2023-07-03
feat: android support for empty path only password derivation path

### LINKED ISSUE #1914: Password derivation not working consistently between iOS and Android
### On iOS
A password derivation of ///password is accepted

iOS Polkadot Vault Version: 6.1.0

### On Android
A password derivation path of ///password is not accepted, the "Create Derived Key" remains grayed out.

A password derivation path of //polkador///password is accepted on Android

Android Polkadot Vault Version: 6.1.2

Dmitry-Borodin:
## Purpose
Fixes #1914 

## Scope
android added support for empty path only password derivation path

## Screenshots
<!--
| Left title, i.e. "Before" | Right title, i.e. "After" |
|-|-|
| image |  |

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -29,7 +29,6 @@ jobs:
 
       - name:                 Install dependencies
         run:                  |
-          brew update
           brew install swiftgen
           brew install swiftformat
           brew install xcbeautify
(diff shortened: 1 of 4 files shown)


krodak:
@Dmitry-Borodin I've pushed commit that should fix CI / CD pipeline for iOS 🙏🏻

Issue #1945, opened 2023-07-12
Render identicons on mobile


prybalko:
### Background
Currently Vault supports 2 types of identity icons:
- `dot` style for substrate based network accounts
- `blockies` for Etherium accounts

There is a plan to also use [jdenticon](https://jdenticon.com/) to represent root public key.

## Proposal
Move icon randering from Rust to Mobile. Rust would return only data and icon style.

### Benefits
- separate data and representation
- use existing libraries to render `jdenticon`:
   - Swift: https://github.com/alejandro-isaza/jdenticon-swift
   - Kotlin: https://github.com/WycliffeAssociates/jdenticon-kotlin
   - Rust: no library
- use existing libraries to render `blockies`:
   - Swift: https://github.com/Boilertalk/BlockiesSwift
   - Kotlin: https://github.com/5afe/svalinn-kotlin
   - Rust: https://github.com/snoopy3476/eth-blockies-rs

### Challenges
There are no existing Swift\Kotlin libraries to generate `dot` style icons.

Existing implementations:
JS: https://github.com/polkadot-js/ui/blob/master/packages/react-identicon
Rust: https://github.com/paritytech/polkadot-identicon-rust

krodak:
Just FYI, we'd need to use our own code for `blockies` and `jdenticon` representation on iOS, as currently linked libraries are 5-7 years old, does not support SPM for distribution which we are using, and won't be compatible with current SwiftUI tech stack.

`BlockiesSwift` looks real simple, so should be fine to replicate similar behaviour 👍🏻
 Same goes for `jdenticon-swift`
 I'm not a JS dev, but based on linked file, I'm not sure how much effort `dot` styles would require, might be easier to reference it from Rust code maybe? All I see in linked files is imported files for jdenticon and ethereum icons and some logic to switch between those; any other pseudocode would be fine as reference for me

prybalko:
Here is the Rust lib for `dot`
https://github.com/paritytech/polkadot-identicon-rust

prybalko:
Resolved in https://github.com/paritytech/parity-signer/pull/1947

PR #1947, opened 2023-07-13
chore: #1945: rust: move identicon generation from rust to mobile

prybalko:
## Purpose

Move identicon generation from rust to mobile.
For more details: https://github.com/paritytech/parity-signer/issues/1945

## ToDo:
- [x] migrate iOS
- [x] migrate Android

## iOS Screenshots
Identicons on both screenshots should be identical
| Current solution | Old solution |
|-|-|
|| |


Dmitry-Borodin:
adapted changes on android
image
From android side it's ready to merge.

Dmitry-Borodin (approved):

android/src/main/java/io/parity/signer/components/networkicon/IdentIconImage.kt
```diff
@@ -0,0 +1,91 @@
...
+			)
+		}
+
+		is Identicon.Dots -> {
+			DotIcon(seed = identIcon.identity,
+				size = size,
+				modifier = modifier,
+				)
+		}
+//		is ImageContent.Svg -> {//will be used for another type of icons
```
krodak, line 41: Q: Shouldn't we just delete it, if we need to bring it back, we can use git history?

krodak (approved):
Good to merge from my side, iOS updated

PR #1948, opened 2023-07-13
feat: android - require secure environment 

Dmitry-Borodin:
Require secure environment which means don't run inside container apps on android device

https://support.google.com/googleplay/android-developer/answer/9888379 - this will be required to respect by container apps, at least those available on google play


krodak (approved):

Dmitry-Borodin:
@prybalko could you take a look at clippy?
Interesting how at appeared in master if checks don't pass...

PR #2018, opened 2023-08-21
fix: don't stop auth flow if only one auth attempt is failed

### LINKED ISSUE #1988: Parity Signer drops seed phrase on authentification failure
When you _Recover seed_, you enter the account name and then the seed phrase. After you're done with that, you click _Next_, and you'll be asked to authenticate yourself biometrically. When this fails, Parity Signer will throw you out of the dialog and drop the seed phrase you just entered. You have to do it all over again and hope that authentification doesn't fail this time.

This is endlessly infuriating, especially with 24 word phrases. Why don't we just get another chance to authenticate?

Dmitry-Borodin:
## Purpose
fixes #1988

I can't reproduce actual problem with api 30 or my devices, but this should never stop auth process if only 1 sttempt was failed.


Dmitry-Borodin:
Despite I cannot reproduce original issue, I think it make sense to merge this one as we don't need to stop auth flow with filed attempt.

prybalko (approved):

PR #2047, opened 2023-08-29
feat: iOS: Recovery key set API update

krodak:
## Purpose
Update Recovery Key Set flow to use non-navigation based API

## Screenshots

| Updated flow |
|-|
||


prybalko (approved):
👍

PR #2057, opened 2023-09-05
feat: ios: key details redesign part 1

krodak:
## Purpose

Initial changes on Key Details:
- jdenticon in header
- header redesign
- updated Show Root Key modal along with QR code for it
- updated icon for Export Keys (not visible on gifs)

## Screenshots

| Light | Dark |
|-|-|
|||


prybalko:
A couple of comments:
1. What does this sign mean? It looks actionable, but it isn't.
Screenshot 2023-09-05 at 11 23 14

2. I would have expected the 'swipe down' gesture to close this modal, but it doesn't work this way. Is this something that can be fixed?

Screenshot 2023-09-05 at 11 30 42

krodak:
@prybalko 
1. In next PR, tap on seed name and arrow will open this modal:
 
which I'm currently working on
2. This is not something that is built in for full screen modal in SwiftUI (it is full screen underneath as we are presenting "dark dim" part above actual modal context), but we can think about building this as functionality that would work app-wide, let me add this to my to-do list 🙏🏻 But for sure this would require separate PR and testing this with all full screen modals we have

PR #2063, opened 2023-09-11
feat: ios: remove old modal from Key Set List and add new modal for new Key Sets Manage view

krodak:
## Purpose
Further steps into new redesigned flow, moving keys management to Key Set Details screen

## Screenshots

| Light | Dark |
|-|-|
|||


prybalko (approved):
It might be more intuitive to make jdenticon clickable as well. It's larger and easer to tap on.

krodak:
> It might be more intuitive to make jdenticon clickable as well. It's larger and easer to tap on.

@tetekinandrey what do you think about that? I think that would be a good idea, I'd add this in next PR if you are fine with that 🙏🏻

PR #2064, opened 2023-09-11
feat: ios: support space tap on seed phrase recovery

krodak:
## Purpose
As @Dmitry-Borodin noticed, in old design we were supporting following use case:
- user types in whole word that matches guess word
- user taps "space"
- word is added automatically to recovered seed phrase without a need to tap guess word

This PR brings this functionality back in new redesigned interface

## Screenshots
| Example|
|-|
||


Dmitry-Borodin (approved):

prybalko (approved):
👍

PR #2066, opened 2023-09-11
feat: #2065 ios: Secret recovery field is not tappable

### LINKED ISSUE #2065: Secret recovery field is not tappable
As a user
I'd like to tap on a center of a secret recovery text field and start typing
But the large area of the field is unresponsive
IMG_0070

krodak:
## Purpose
Resolves #2065

## Screenshots

| Example |
|-|
||


prybalko (approved):
Thanks!

krodak:
No worries @prybalko , if you see anything like this that could be improved from UX standpoint, just add a task, these should be usually minor fixes in native UI and working on app on daily basis it's hard sometimes to notice those as oddities 🙏🏻

PR #2077, opened 2023-09-19
feat: add sufficient crypto api

prybalko:
## Purpose
Add apis:
- `get_keys_for_signing`
- `sign_metadata_with_key` api
- `sign_network_spec_with_key`

Add `WrongPassword` error


rust/signer/src/lib.rs
```diff
@@ -86,6 +87,8 @@ pub enum ErrorDisplayed {
     NoMetadata {
         name: String,
     },
+    /// Provided password is incorrect
+    WrongPassword,
```
Dmitry-Borodin, line 91: I think this is not displayed error - this is one of states we should ask for password rather than just show text information about failure.
I think we should find a way to mark only those actions with it, that may actually require password. 
Either mark only those endpoint with exception or return enum of result and this marker in those 3 actions - export key qr code, transaction signing, specs signing.

prybalko: Password request screen happens before `sign_metadata_with_key` is called. So `WrongPassword` is not a state asking for a password, but rather an indication that the provided password is incorrect.

krodak (approved):

rust/signer/src/lib.rs
krodak: In general I agree with what you are saying @Dmitry-Borodin but I’m fine with current solution for now
What I do is to handle screen-specific screen in given flow(I.e. transaction special errors when scanning QR code) and if error is different enum value just fallback to default error modal with appropriate message for that error.
So in this case you can support special enum value within given flow while still not dealing with every other enum value in each place

prybalko: > if error is different enum value just fallback to default error modal with appropriate message for that error.

That was my idea, yes. 

@Dmitry-Borodin Do you mean the error enums should look like this?

```rust
enum GenericApiError {
    DbNotInitialized(String)
}

enum TransactionApiError {
    DbNotInitialized(String),
    WrongPassword,
    NoMetadata(number),
    NoNetowrk(String)
}

enum PasswordedApiError{
    DbNotInitialized(String),
    WrongPassword,
}
```

Dmitry-Borodin: @prybalko I think in this API wrong password is not even an error, it's enum of valid result. if we pass it an en error - I would handle it separately.
We can group errors to handle them together, then it will be something like 

> /// Unreachable irrecoverable error - only restart of the app is possible
> enum GenericApiError { 
    DbNotInitialized(String)
    MutexPoisoned()
}

> /// something failed with current call - show text about what happened is continue working
> enum DisplayError {
    NoMetadata(number),
    NoNetowrk(String),
    LoadMetaUnknownNetwork(String)
    SpecsKnown(String, String)
    Str(String)
}

And things like Wrong password should be marked for only those 3 api calls can may need a password. Generic for All calls. Display errors only for calls that may return them. 
I don't think we need to wrap it in it's own enum. I still have problem in Kotlin that exceptions are not checked, but at least in specs I will see what should I handle where, instead of trying to remember it for each screen.
In my code I"m converting it to Result tuple object, and it would be nice to see which exact cases I should handle where.


So APi will look like

>    [Throws=ErrorDisplayed, ErrorGeneric, WrongPassword]
    void remove_key_set([ByRef] string address_key);

prybalko: > [Throws=ErrorDisplayed, ErrorGeneric, WrongPassword]

I don't think that's possible. 
Rust doesn't have exceptions. Normally functions return `Result<T, E>`, which is a enum of 2 possible values: ok or error value. Uniffi converts `E` into an expection for Kotlin\Swift.

https://mozilla.github.io/uniffi-rs/udl/errors.html?highlight=Throws#throwing-errors

Dmitry-Borodin: Then we'll need to unite those 2-3 errors in single enum types.
The point is to follow interface segregation principle and force called to handle those errors, that it need to handle, not all of possible cases overall.
I like idea to unite all things "show message and do nothing about it", but something like "try different password" shouldn't be one of them.

Btw it's sad they using java approach for kotlin with exceptions, but looks like it's not actively developed now, so I"ll convert on my own.

PR #2082, opened 2023-09-22
refactor: replace `ENV[]` with `ENV.fetch`

philipjonsen:
`ENV[]` silently fails and returns `nil` when the environment variable is unset, which may cause unexpected behaviors when the developer forgets to set it. On the other hand, `ENV.fetch` raises `KeyError` or returns the explicitly specified default value.

### DIFF
--- a/ios/fastlane/lanes/lane_build.rb
+++ b/ios/fastlane/lanes/lane_build.rb
@@ -20,7 +20,7 @@ lane :build_release do |options|
   clean_build_artifacts
 
   increment_build_number(
-    build_number: ENV["BUILD_NUMBER"], # based on commit, defined in GA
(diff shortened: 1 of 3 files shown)


ios/fastlane/lanes/lane_build.rb
```diff
@@ -20,7 +20,7 @@
   clean_build_artifacts
 
   increment_build_number(
-    build_number: ENV["BUILD_NUMBER"], # based on commit, defined in GA
+    build_number: ENV.fetch("BUILD_NUMBER", nil), # based on commit, defined in GA
```
prybalko, line 23: Thank you for the contribution!
Although I would prefer to use `ENV.fetch` without a default value, otherwise, it's identical to what we carrently have.

prybalko (changes requested):

PR #2098, opened 2023-09-28
chore: store and check db schema version

prybalko:
## Purpose
Detect db schema change, prevent app from crashing

## Scope
- store db version
- expose `check_db_version` to check if app has been upgraded without the re-install.
- added `DbSchemaMismatch` error to catch

Scheme version should be bumped manually every time there is a breaking change on the backend.

### DIFF
--- a/android/src/main/java/io/parity/signer/screens/scan/errors/TransactionErrorsModels.kt
+++ b/android/src/main/java/io/parity/signer/screens/scan/errors/TransactionErrorsModels.kt
@@ -39,7 +39,7 @@ tailrec fun Throwable.findErrorDisplayed(): ErrorDisplayed? {
 
 fun ErrorDisplayed.toTransactionError(): TransactionError {
 	return when (this) {
-		is ErrorDisplayed.DbNotInitialized -> TransactionError.Generic("")
+		is ErrorDisplayed.DbNotInitialized -> TransactionError.Generic("Db Not Initialized. Unreachable state here.")
 		is ErrorDisplayed.LoadMetaUnknownNetwork -> TransactionError.MetadataForUnknownNetwork(
 			name
 		)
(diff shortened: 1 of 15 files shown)


Dmitry-Borodin:
Will this rely on you keeping manual track of db schema and updating constant?
I've thought we will just catch db errors from unmarchalling data and return this error in case we face it. Then we will not rely on manual keeping track on schema changing.

krodak (approved):

prybalko:
> Will this rely on you keeping manual track of db schema and updating constant? I've thought we will just catch db errors from unmarchalling data and return this error in case we face it. I see we have now. Not sure that we need manual check and versioning on top of that.

Yeah, I was thinking about that. It's tempting to have automatic checks, but it feels hacky to me. It's like we're trying to _guess_ the reason why unmarchalling is failing. Maybe it's a problem writing data to the db? Or is db somehow got corrupted? We'll never know. It'll just show "db schema version error", which is not true. 
In addition, if db _key_ changes, there won't be an error, just an empty state.

@krodak what's your input on that?

krodak:
> > Will this rely on you keeping manual track of db schema and updating constant? I've thought we will just catch db errors from unmarchalling data and return this error in case we face it. I see we have now. Not sure that we need manual check and versioning on top of that.
> 
> Yeah, I was thinking about that. It's tempting to have automatic checks, but it feels hacky to me. It's like we're trying to _guess_ the reason why unmarchalling is failing. Is it a problem writing data to the db? Or is db somehow got corrupted? We'll never know. It'll just show "db schema version error", which is not true. In addition, if db _key_ changes, there won't be an error, just an empty state.
> 
> @krodak what's your input on that?

Although not ideal situation in case of maintenance, I think db scheme version check would be good solution for that now, it's as explicit as it gets and we are small enough of the team to be able to pay attention to when there are breaking changes in model (i.e. all three of us need to push to same PR to fix builds 😉 ).

prybalko:
I'll merge it as is, not to delay mobile development. There is `DbSchemaMismatch` that will stay anyways, the logic behind could be changed later

PR #2110, opened 2023-10-05
fix: android onboarding remove corousel

Dmitry-Borodin:
Actually decided to remove onboarding coousel
and
 fix: android onboarding last step second box was not shown in real pager


prybalko (approved):

android/src/main/java/io/parity/signer/ui/rootnavigation/RootNavigationGraph.kt
```diff
@@ -81,23 +81,16 @@ fun NavGraphBuilder.firstTimeOnlyOnboarding(
...
-		}
 		composable(route = FirstTimeOnboarding.termsConsentRoute) {
 			TermsConsentScreenFull(navigateNextScreen = onNextStepsNavigate)
 		}
 	}
 }
 
 private object FirstTimeOnboarding {
-	const val onboardingExplanationRoute = "navigation_onboarding_explanation"
+//	const val onboardingExplanationRoute = "navigation_onboarding_explanation"
```
krodak, line 93: Shouldn't this be removed, given we have git history to bring it back?

android/src/main/java/io/parity/signer/screens/initial/explanation/OnboardingScreen4.kt
```diff
@@ -26,24 +26,26 @@ import io.parity.signer.ui.theme.*
...
-		contentAlignment = Alignment.BottomCenter,
-	) {
-		PrimaryButtonWide(
-			label = stringResource(R.string.onboarding_cta_continue),
-			activeBackground = Color.White,
-			activeText = MaterialTheme.colors.pink500,
-			modifier = Modifier.padding(24.dp),
-			onClicked = onContinue,
+	Box() {
+		OnboardingScreenGeneric(
```
krodak, line 30: Shouldn't all onboarding screens be removed in this PR? We can always bring them back from git history

krodak (commented):
If we don't plan on bringing back this onboarding very soon, I think deleting all screens and related resources could be a good move, I followed that within #2104

android/src/main/java/io/parity/signer/ui/rootnavigation/RootNavigationGraph.kt
Dmitry-Borodin: will remove on subsequent redesign pr

android/src/main/java/io/parity/signer/screens/initial/explanation/OnboardingScreen4.kt
Dmitry-Borodin: I have a feeling we will need it soon -) If not will remove soon, but it's much easier to search in existing files than in a history for unknown filenames

Issue #2092, opened 2023-09-26
Do not crash after app upgrade


prybalko:
Vault doesn't have database migrations, so when the user updates the app, it crashes. 
Instead of a crash, there should be a message saying, `App upgrade is not supported; please re-install the app.`

Tasks:
- [x] fix iOS
- [x] fix Android

Dmitry-Borodin:
@prybalko What error rust will tell us in case of not migrated db?

prybalko:
Added `DbSchemaMismatch` in https://github.com/paritytech/parity-signer/pull/2098

Dmitry-Borodin:
#2164 implemented new screens as per design and manual version check. So this issue is done.

PR #2114, opened 2023-10-06
feat: iso: database scheme error baseline scenario handling

krodak:
## Purpose
This handles #2092 in baseline scenario:
- user opens the app
- app was updated and db scheme was changed
- user is not able to proceed, but able to retrieve seed phrase

## Discussion
We should probably discuss:
- what if `checkDbVersion` will return different error (as we use enum, technically we need to handle this scenario) - for now I've assumed same behaviour as baseline scenario
- should we react to `DbSchemaMismatch` anywhere else and act the same - database scheme should not be changed  during runtime or without reinstall

## Screenshots
| Example | |
|-|-|
|| |


prybalko (approved):
> what if checkDbVersion will return different error (as we use enum, technically we need to handle this scenario) - for now I've assumed same behaviour as baseline scenario

Makes sense

> should we react to DbSchemaMismatch anywhere else and act the same - database scheme should not be changed during runtime or without reinstall

`DbSchemaMismatch` is never returned in other API calls, so it can be handled as a simple string error

PR #2139, opened 2023-10-20
feat: ios: Migrate from SwiftGen to new native Xcode 15 assets accessors

krodak:
## Purpose
To reduce dependency on 3rd party tooling, use new Xcode 15 setting to autogenerate assets / colors accessors


prybalko (approved):
👍

PR #2142, opened 2023-10-23
feat: force auth for each transaction signature even if old not timed out

Dmitry-Borodin:
Currently storage unlocked on android for 5 min. I am forcing auth on UI for some cases like removing certificate, login auth or exporting secrets.
For signing transaction old version was requring pin only if auth times out. 
Because user will need to scan qr code and pass result qr to camera back - it cannot happen accidentally I think. 
I not sure if forcing auth each time will improve user experience but those are changes as discussed in security channel.

maybe @tetekinandrey have some thoughts.


Dmitry-Borodin:
btw it's ready to merge if we want it.

PR #2143, opened 2023-10-24
feat: ios: introduce way to create default derived key

krodak:
## Purpose
Introduce way to create derived key quickly without specifying name

## Screenshots

| Light | Dark |
|-|-|
|||


ios/PolkadotVault.xcodeproj/project.pbxproj
```diff
@@ -2477,7 +2477,7 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				IPHONEOS_DEPLOYMENT_TARGET = 15.6;
+				IPHONEOS_DEPLOYMENT_TARGET = 15.5;
```
prybalko, line 2480: downgrade?

krodak: Yes, as this would exclude some iPod Touch models, without any upside

PR #2185, opened 2023-11-15
feat: removing tell rust seed names from mobile clients

Dmitry-Borodin:
Removing tell rust seed names as we don't use rust state machine and they don't need to know seed names anymore


prybalko (approved):

krodak:
@Dmitry-Borodin pushed the same update on my side, hope you don't mind 😅 🙏🏻

krodak (approved):

PR #2194, opened 2023-11-20
feat: expose validate_key_password

prybalko:
## Purpose
Expose `validate_key_password`, because it might be useful to remove navstate


rust/signer/src/lib.rs
```diff
@@ -598,6 +598,21 @@ fn get_keys_for_signing() -> Result<MSignSufficientCrypto, ErrorDisplayed> {
     Ok(MSignSufficientCrypto { identities })
 }
 
+fn assert_account_password(
```
Dmitry-Borodin, line 601: validate_key_password ?

Dmitry-Borodin: I think account was used for keyset

prybalko: Good point, fixed

krodak (approved):

PR #2302, opened 2024-01-22
feat: Android add seed handle and show proper errors

Dmitry-Borodin:
## Purpose
previous implementation was showing if operation successful or not. This will pass exact error and show error screen with description if creation of seed is impossible for some reason.


Dmitry-Borodin:
checking what are the errors..

krodak (approved):

PR #2366, opened 2024-03-11
feat: ios: banana split recovery update

krodak:
## Purpose
Add option to trigger qr code scanner when in recovery key set mode

## Scope
- add qr code button in recovery
- add option to dismiss all modals when BS recovery is done through new flow
- pass seed name to QR code scanner flow for BS


prybalko (approved):
🚀

PR #2429, opened 2025-02-03
fix: xcode16 build

ERussel:
## Purpose

The PR fixes iOS specific build errors

## Scope

- fixed renamed method in the UIApplication
- add support for opencv 4.11.0 in the backend

## TODOs

Currently, CI relies on whatever version of the opencv it can find via `brew install`. It will be safer if it will be using the fixed version of the opencv.


stepanLav (approved):

FlorianFranzen:
@ERussel @stepanLav Really appreciate your work, however this requires even the parity-signer rust library to be build against rather modern version of opencv. 

Could you please consider my change in #2450 to make us much more backwards compatible?

PR #2448, opened 2025-03-09
feat: support signing via v15 metadata proof

ERussel:
## Purpose

This PR adds an option to sign extrinsic without requiring latest metadata for a chain. The chain still needs to be added to the App via chain spec scaning.

## Scope

- introduce new message type (6) that requires scale encoded metadata proof in the parameters: `<network_type>|<crypto_type>|<message_type>|<account_id>|<metadata_proof>|<call>|<extensions>|<genesis_hash>`

- introduce StateMachine implementation to decode call and extensions based on the proof types into a list of cards to display in the interface


rust/parser/src/number_state.rs
```diff
@@ -0,0 +1,231 @@
...
+        Ok(output)
+    }
+
+    fn process_i8(&self, input: i8, indent: u32) -> Result<StateOutput, StateError> {
+        let output = self.process_number(input.to_string(), indent);
+
+        Ok(output)
+    }
+
+    fn process_i16<'scale, 'resolver>(
```
valentunn, line 92: Do we need lifetimes here? I don't see them in other number methods, e.g. `i32` or `i8`

rust/parser/src/state_machine.rs
```diff
@@ -0,0 +1,647 @@
...
+/// to the current state the decision-making regarding the generated output cards and
+/// transitions to the next state.
+///
+/// For more information on how state transitions and output generation are handled,
+/// refer to the [`State`] trait documentation.
+pub struct TypeRegistry(BTreeMap<u32, Vec<Type>>);
+
+impl TypeRegistry {
+    pub fn new<'a>(types: impl Iterator<Item = &'a Type>) -> Self {
+        let tree: BTreeMap<u32, Vec<Type>> = types.fold(Default::default(), |mut map, ty| {
```
valentunn, line 29: Is there a specific case when there may be multiple types corresponding to a single `type_id`? Is that how we store enum variants?

valentunn (approved):

rust/parser/src/account_state.rs
```diff
@@ -0,0 +1,161 @@
...
+        let card = match current_len {
+            32 => {
+                let account_id: [u8; 32] = new_partial
+                    .try_into()
+                    .map_err(|_err| StateError::BadInput("32 bytes expected".to_string()))?;
+
+                OutputCard {
+                    card: ParserCard::Id {
+                        id: AccountId32::new(account_id),
+                        base58prefix: self.extra_info.base58_prefix,
```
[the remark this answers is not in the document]
ERussel: Makes sense. But I would do it separately.

rust/parser/src/number_state.rs
ERussel: I don't think so we need those parameters. Removed.

rust/parser/src/state_machine.rs
ERussel: Yes, a list of enum variants is stored per enum type id

PR #2450, opened 2025-04-02
fix: opencv backwards compatibility

FlorianFranzen:
## Purpose
We were updating the metadata-portal to be compatible with newer version of opencv and had to realize that the way opencv is used in the rust library of the signer is not backward compatible (see #2429 for details).

As the signer does not make use of the additional flags available in newer opencv, there is actually no reason to not be backwards compatible, so this PR fixes that. 

This has been tested with the latest version of the metadata-portal and the changes proposed in paritytech/metadata-portal#845. The result is that the metadata-portal can be build with older and newer version of opencv without any issues.

## Scope
Use different opencv binding that assumes default 

@ERussel @stepanLav

### DIFF
--- a/rust/qr_reader_pc/src/lib.rs
+++ b/rust/qr_reader_pc/src/lib.rs
@@ -13,9 +13,8 @@ use indicatif::ProgressBar;
 use qr_reader_phone::process_payload::{process_decoded_payload, InProgress, Ready};
 
 use opencv::{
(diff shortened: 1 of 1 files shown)


ERussel (approved):
Thank you, Florian. We had some issues with iOS compatibility previously but looks like tests are passing now.

PR #2466, opened 2025-04-16
fix: migration app store

ERussel:
## Purpose

The PR introduces a fix to App Store migration. Since the is migrated to Novasama Technologies App Store the previous keychain/seeds will not be available to a user. However, the old database file still available to the application that introduces inconsistencies with the new keychain data. The solution is to change database filename and thus triggering database recreation after the migrated update is installed above old app version.

### DIFF
--- a/ios/PolkadotVault/Core/Database/DatabaseMediator.swift
+++ b/ios/PolkadotVault/Core/Database/DatabaseMediator.swift
@@ -27,7 +27,8 @@ protocol DatabaseMediating: AnyObject {
 /// Class that should act as main accessor for database
 final class DatabaseMediator: DatabaseMediating {
     private enum Constants {
-        static let resource = "DatabaseV7_0_0"
+        static let bundleResource = "Database"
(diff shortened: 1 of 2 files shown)


ios/PolkadotVault/Core/Database/DatabaseMediator.swift
```diff
@@ -27,7 +27,8 @@ protocol DatabaseMediating: AnyObject {
 /// Class that should act as main accessor for database
 final class DatabaseMediator: DatabaseMediating {
     private enum Constants {
-        static let resource = "Database"
+        static let bundleResource = "Database"
```
sergkononenko, line 30: Wouldn't it be more informative to name it smth like `oldBundleResource` or even add a deprecation message here?

ERussel: The idea here is to separate naming for the bundle resource and file in the filesystem. When the app starts and there is no database created in the filesystem it is copied from the bundle

sergkononenko (approved):

PR #2536, opened 2025-12-10
fix: Increase target sdk version to 35

antonijzelinskij:
## Purpose
- Increase targetSdk version to follow Google Play Console requirements

## Scope
- Increased targetSdk version
- Enabled edge to edge screen mode
- Fixed bottom sheet bugs

## Discussion
During making this task I understood we need to refactor insets in this project since bottom sheets must have their own insets.
For now insets are set for full navigation graph that is wrong. By this reason we had that state:
imageimage

I fixed it by removing elevation for all bottom sheets but it's not the right way. For now it's looking like that. Not Perfect but nice as a little fix:
imageimageimageimage

### DIFF
--- a/ios/.swiftformat
+++ b/ios/.swiftformat
@@ -8,6 +8,8 @@
 
 # file options
 --exclude PolkadotVault/Generated/
+--exclude PolkadotVault/Stubs/Stubs.swift
+--exclude PolkadotVault/Previews/Stubs.swift
 
 # rules
 --enable blankLineAfterImports


android/src/main/java/io/parity/signer/MainActivity.kt
```diff
@@ -20,7 +20,7 @@ class MainActivity : AppCompatActivity() {
 		super.onCreate(savedInstanceState)
 		ServiceLocator.initActivityDependencies(this)
 
-		WindowCompat.enableEdgeToEdge(window)
+		enableEdgeToEdge()
```
antonijzelinskij: This is more fresh method that setups system bars colors

PR #2543, opened 2026-02-01
fix: malicious use cases during extrinsic parsing

ERussel:
## Purpose

This PR fixes several DoS vulnerabilities in the metadata parser that could cause the application to crash or hang when processing malicious payloads:

1. **Out of memory during array parsing** - Arrays with zero-sized element types (e.g., `[(); N]`) where N is extremely large would cause the decoder to iterate billions of times
2. **Out of memory during sequence parsing** - Sequences claiming more items than available bytes (e.g., a `Vec<u8>` header claiming 1 million items with only 4 bytes of data)
3. **Stack overflow from deeply nested types** - Malicious payloads with deeply nested structures (e.g., 1000 levels of `utility.batch` calls) would overflow the native call stack. See #2542

### Early Bailout via `unchecked_decode_as_type`

The straightforward approach of adding checks in visitor methods (`visit_sequence`, `visit_array`, etc.) doesn't work because `scale_decode` still calls `skip_decoding` to advance past unprocessed bytes. This `skip_decoding` internally uses `IgnoreVisitor` which recursively traverses the entire nested structure, causing stack overflow before our checks can fire.

The solution is to implement `unchecked_decode_as_type` on our `StateMachineParser`. This method is called **before** type resolution and decoding begins. By returning `DecodeAsTypeResult::Decoded(Err(...))` here, we bypass the normal decoding path entirely, including the `skip_decoding` call.

We check for:
- **Recursion depth** - Prevents stack overflow from deeply nested types
- **Array length** - Rejects arrays claiming more items than available bytes
- **Sequence length** - Peeks at compact-encoded length and rejects impossible claims

### Direct `decode_item` Instead of Iterators

The PR also changes how visitor methods iterate over fields/items. The problem with using iterators on `scale_decode` types (`Composite`, `Sequence`, `Tuple`, `Array`) is that their `Iterator::next()` implementation calls `decode_item(IgnoreVisitor)` to pre-decode each element and determine byte boundaries. This recursive pre-decoding bypasses our depth checks.

**Before:**
```rust
for (index, field_result) in value.fields().enumerate() {
    let field = field_result?;
    visitor = field.decode_with_visitor(visitor)?;
}
```

**After:**
```rust
// Extract metadata upfront from the type definition
let field_infos: Vec<_> = value.fields().fields().iter()
    .map(|f| (f.name, f.id.clone()))
    .collect();

for (index, (field_name, field_type_id)) in field_infos.into_iter().enumerate() {
    // Use decode_item directly - decodes with OUR visitor, not IgnoreVisitor
    visitor = value.fields().decode_item(visitor)
        .ok_or_else(|| StateError::BadInput("Unexpected end of fields".into()))??;
}
```

This ensures all recursive decoding goes through our `unchecked_decode_as_type` check.

## Configuration

`MAX_RECURSION_DEPTH = 128` allows:
- 32 levels of nested batch calls (128 / 4 entries per level)
- Safe operation on iOS (512KB thread stack) and Android (~1MB)
- Empirically tested: 128 works safely, 256 overflows in debug builds
- Far exceeds legitimate use cases (complex governance rarely exceeds 6 levels)

## Testing

- `parse_extrinsic_with_malicious_array_type` - Verifies large array rejection
- `parse_extrinsic_with_malicious_sequence_length` - Verifies impossible sequence length rejection
- `parse_extrinsic_with_deeply_nested_calls` - Verifies 1000-level nesting is rejected without stack overflow

## Changes

- `state_machine.rs`: Implement `unchecked_decode_as_type` with depth and length checks
- `state_machine.rs`: Refactor all visitor methods to use `decode_item` directly
- `tests.rs`: Add test cases for malicious payloads
- `for_tests/deeply_nested_extrinsic`: Binary test data with 1000 nested batch calls


rust/parser/src/state_machine.rs
```diff
@@ -479,19 +573,14 @@ impl Visitor for StateMachineParser<'_> {
             .process_tuple(&input, visitor.indent)?;
         visitor.apply(output);
 
-        for (index, field_result) in value.enumerate() {
+        // Use decode_item directly instead of iterator to avoid IgnoreVisitor pre-decoding
+        for index in 0..items_count {
             visitor.push_indent();
 
-            let field = field_result?;
-            let item_path = visitor
```
svojsu, line 486: Why we decide to cut the path off here? For `visit_array` and `visit_sequence`, you extract `item_type_id` upfront from the `TypeDef`. Could we do the same for tuples by extracting the field types from `TypeDef::Tuple`?

ERussel: It is usually empty for Tuple. So not big miss but thank you for pointing that. Added additional logic

svojsu (approved):

rust/parser/src/state_machine.rs
```diff
@@ -138,6 +151,72 @@ impl Visitor for StateMachineParser<'_> {
...
+                        return DecodeAsTypeResult::Decoded(Err(StateError::BadInput(format!(
+                            "Array claims {} items but only {} bytes remain",
+                            arr.len,
+                            input.len()
+                        ))));
+                    }
+                }
+                TypeDef::Sequence(_) => {
+                    // Peek at the compact-encoded length without consuming bytes
+                    let mut peek_input = *input;
```
valentunn, line 199: How does it ensure we are not consuming `input` bytes?

rust/parser/src/state_machine.rs
```diff
@@ -327,10 +406,14 @@ impl Visitor for StateMachineParser<'_> {
...
         let mut visitor = self;
 
-        let path = visitor
-            .type_registry
-            .get_first_type(&type_id)
-            .map(|v| v.path);
+        let seq_type = visitor.type_registry.get_first_type(&type_id);
+        let path = seq_type.as_ref().map(|v| v.path.clone());
+
+        // Extract item type from sequence's TypeDef to avoid using iterator which pre-decodes with IgnoreVisitor
```
valentunn, line 412: I'd leave a comment that links to this PR to keep a note why pre-decoding with `IgnoreVisitor` is not desired

valentunn (approved):

rust/parser/src/state_machine.rs
ERussel: Dereferencing of input copies entire byte array into peek_input

